PR #16, opened 2021-12-17
Dvorak layout

AlexandraAlter:
Prospective Dvorak layout!

I've had to move numbers to the middle row, since it's the only row in Dvorak with the required ten keys.
Since there's less room on the bottom row, the backspace and shift keys have also ended up on the top row - sorry.
I'll be experimenting with a locally-built copy for a while, to see if it feels right, but in the meantime, any comments on the code are welcome.


Julow (approved):
Awesome, thanks !

The first row seems a little weird at first glance but why not.
Happy to merge once you validate the design.

res/xml/dvorak.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+	<row>
+		<key key0="backspace" key1="esc" key2="`" key3="~" key4="delete" width="1.5" />
+		<key key0="shift" width="1.5" />
+		<key key0="p" key1="accent_ring" />
```
Julow, line 6: My finger are less precise on the backspace key. Also it's useful to have an empty corner to be able to modulate the key repeat.
What do you think of moving these to the `p` ?

```suggestion
		<key key0="backspace" key1="esc" key4="delete" width="1.5" />
		<key key0="shift" width="1.5" />
		<key key0="p" key1="accent_ring" key2="`" key3="~" />
```

res/xml/dvorak.xml
```diff
@@ -0,0 +1,44 @@
...
+		<key key0="s" key2="0" />
+	</row>
+	<row>
+		<key key0="q" shift="0.5" />
+		<key key0="j" />
+		<key key0="k" />
+		<key key0="x" key1="accent_cedille" key2="." key3="&lt;" />
+		<key key0="b" key2="," key3="&gt;" />
+		<key key0="m" key2="/" key3="\?" />
+		<key key0="w" key1="accent_tilde" key2=";" key3=":" />
```
Julow, line 33: Shouldn't `;` and `:` be on the `q` ?
Similarly:
- `_` and `-` should be on `z` instead of `f`
- `|` and `\\` on `s` instead of `l`
- `+` and `=` on `l`

res/xml/dvorak.xml
Julow: Also, what about moving the backspace to the end of the top row ?

Julow (commented):
I would like to merge this soon, before the release. How is your experiment going ? Happy to merge once you mark the PR as ready.

There have been many changes on `master` so I pushed a merge commit.

res/xml/dvorak.xml
AlexandraAlter: I think having the accents on the P key is a great idea.

I'm less sure about moving the backspace, since it'd throw off the alignment of the other keys, but it might be fine? I've been using the layout contiuously since making the PR, and while I've settled into it, it does still feel a little odd. I haven't really struck upon a solution that appeals to me, but it's worth trying it with the backspace in a more normal position.

res/xml/dvorak.xml
AlexandraAlter: If you want them to be closest to their normal-keyboard equivalents, then yes, but with the far corners being harder to manipulate, I'm still unsure about the symbols I left on L, since I keep ending up with a pipe instead of an L. I think this might be to do with the screen having a high DPI though. Even with the lowest precision, it's all too easy to make mistakes.

If we wanted closest-equivalents, but without using the edge keys too much, it might work.
Putting `,`/`<` on P and `.`/`>` on Y would better mirror how the normal Dvorak keyboard looks.
There's not much room to put `'`/`"` nearby, but those could always be placed on J instead, with Q having `;`/`:`.
I'm not overly fond of using S, since it's on the very edge already. However, with Z free and a half-space from the edge, it's a good candidate for `_`/`-`.
With the backspace moved over to the right, L won't be on the edge, so `+`/`=` might be okay there?
`|` and `\` are a little trickier. I feel like there'd be room somewhere along the bottom row, though.

res/xml/dvorak.xml
AlexandraAlter: > I would like to merge this soon, before the release. How is your experiment going ? Happy to merge once you mark the PR as ready.
> 
> There have been many changes on `master` so I pushed a merge commit.

Thanks for merging the changes.

I've been testing it since I made the PR, and have had a few issues, but I'm still mulling over what the solution might look like.

res/xml/dvorak.xml
Julow: Avoiding the edge keys in some cases is a good idea. It's OK to have some keys far from where they are on a full-size keyboard but I wonder if future Dvorak users will complain.

It's worth investigating the backspace on the right but if that doesn't work for you, let's keep it on the left.

Julow: `,`/`<` and `.`/`>` look like important landmark of the layout so they might currently be a little too far down. If something have to be moved to the third row, I'd say `+`/`=` is the less frequent and less important.
`_`/`-` should be close to its real position because it's frequent and needs to be typed fast in the middle of a word.

AlexandraAlter: Sounds like a good plan. I'll make those changes tomorrow and give it a test.

Julow:
I would like to merge this today, did you find successful key arrangements ?

AlexandraAlter:
Sorry, I've still been testing it out. It seems to work fine though, and I'm pretty happy with where the symbols are.

I'd appreciate if you could have another look over where I've put them, to see if they're consistent and friendly enough. A few things did have to move around to keep the middle row for numeric symbols only.

Otherwise, I'm happy to put this up as ready.

Julow:
The layout looks very nice :)

I'll merge before the next release to give you time to make some more change if needed.

Julow:
Thanks a lot ! Merging now :)

Issue #4, opened 2021-12-04
German layout


Moini:
Hi! 

I like the concept with the sliding to get a different character a lot, but it doesn't work for my language yet.

So I'd like to add German QWERTZ.

I'm able to edit the XML files (I believe this means I'd need to create a copy of one of this file: https://github.com/Julow/Unexpected-Keyboard/blob/master/res/xml/qwerty.xml, not sure about [the accents](https://github.com/Julow/Unexpected-Keyboard/blob/master/res/xml/method.xml), German only has umlauts by default) , but not to compile and test the result.

Julow:
Hi! I'd be happy to accept contributions !

You are right about the xml file. You also need to add some plumbing: [here](https://github.com/Julow/Unexpected-Keyboard/blob/master/res/values/arrays.xml#L7) and [here](https://github.com/Julow/Unexpected-Keyboard/blob/master/res/values/strings.xml#L17) then [here](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/Config.java#L104).
The `accents` option specifies which accents are needed for each languages. Do you think you can use `trema` for your language ? It is implemented with [`KeyCharacterMap.getDeadChar`](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L103). If that's not enough, it is fine to add new cases or a new accent type.

The development environment is the hardest part. I suggest you to use the `shell.nix` present in the repo (warning, the closure is big).
Otherwise, there's not many requirements: `openjdk8`, the Android SDK.
In the SDK, you need: build tools `30.0.3`, platform `29`.

Moini:
Hi @Julow - thanks for the welcome!

German Umlauts are not accented letters, but letters of their own (in Unicode terms, 'precomposed' vs. 'decomposed'), so they would need their own little icon, commonly (as keyboard size is restricted / or even fixed?) on the letter that looks similar, so it would be letter `a` in the center, and `ä` and `´` and ` in three of those corners, same for `o/ö`, `u/ü` and `s/ß`.

I'd like to try just editing the xml files without setting up an Android development environment - will check the xml for validity, but I'll probably need a screenshot of the result to know whether it looks correct. Would you help me with that, when I think it's ready?

Julow:
[...]

PR #20, opened 2021-12-29
WIP: German keyboard added

Moini:
* Translated metadata to German
* Fixed a couple English strings and two small syntax issues I stumbled upon
* Tried adding a German keyboard, currently stuck at trying to make it work without having to use two keystrokes to type a single German letter ä, ö, ü or ß.
* Also added the ß as an 'accent' type thing, for other languages 

What I'd like help with:
* A screenshot to check my layout.
* A check whether I found all the places where I need to change / add code (I don't speak Java, nor have I ever done any Android app programming)
* Trying to find a way to make it work with a single key swipe in German to get ä,ö,ü,ß, it would be so awkward otherwise.

Please have a look, @Julow ! Merci!

If it doesn't work out with the one-keyswipe thing, please pick what you need for the other solution.


srcs/juloo.keyboard2/KeyValue.java
```diff
@@ -190,6 +192,11 @@ private static void addEventKey(String name, String symbol, int event, int flags
     addCharKey('#', KeyEvent.KEYCODE_POUND);
     addCharKey('(', KeyEvent.KEYCODE_NUMPAD_LEFT_PAREN);
     addCharKey(')', KeyEvent.KEYCODE_NUMPAD_RIGHT_PAREN);
+    // Can't combine here, right?
+    // Is there another way to have these reachable directly?
+    // addCharKey('ä', KeyEvent.VK_DEAD_DIAERESIS + KeyEvent.VK_A)
+    // addCharKey('ö', KeyEvent.VK_DEAD_DIAERESIS + KeyEvent.VK_O)
+    // addCharKey('ü', KeyEvent.VK_DEAD_DIAERESIS + KeyEvent.VK_U)
```
Julow, line 199: We don't need keycodes for these characters, so no need to define them here.

srcs/juloo.keyboard2/KeyValue.java
```diff
@@ -42,10 +42,11 @@
   public static final int FLAG_ACCENT_SUPERSCRIPT = (1 << 22);
   public static final int FLAG_ACCENT_SUBSCRIPT = (1 << 23);
   public static final int FLAG_ACCENT_RING = (1 << 24);
+  public static final int FLAG_ACCENT_SZLIG = (1 << 25);
```
Julow, line 45: As you said, having an accent for this is annoying: need two gestures to type the `ß` character.

Adding that character on all the layouts would be fine, but we need a way to remove it for non-German users. (I'll work on that part)

res/xml/method.xml
```diff
@@ -3,6 +3,7 @@
   <subtype android:label="@string/subtype_label" android:languageTag="en" android:imeSubtypeLocale="en_US" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty"/>
   <subtype android:label="@string/subtype_label" android:languageTag="fr" android:imeSubtypeLocale="fr_FR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=azerty,accents=grave|aigu|circonflexe|cedille|trema"/>
   <subtype android:label="@string/subtype_label" android:languageTag="es" android:imeSubtypeLocale="es_ES" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,accents=aigu|tilde|trema"/>
+  <subtype android:label="@string/subtype_label" android:languageTag="de" android:imeSubtypeLocale="de_DE" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz,accents=grave|aigu|circonflexe"/>
```
Julow, line 6: These are `è`, `é` and `ê`. Do you mean this instead ?
```suggestion
  <subtype android:label="@string/subtype_label" android:languageTag="de" android:imeSubtypeLocale="de_DE" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz,accents=trema"/>
```

metadata/android/de-DE/full_description.txt
```diff
@@ -0,0 +1,6 @@
+Alle benötigten Buchstaben und Zeichen sind von einer einzigen Ansicht aus erreichbar. Strg, Alt, Escape und viele weitere stehen ebenfalls zur Verfügung.
```
Julow, line 1: This is nice, thanks! However, I won't be able to maintain these (I'll improve the `full_description` soon and there's new changelogs to write all the time).
Would it make sense to not translate the changelogs and the full description (German readers will see them in english) to be sure they are uptodate ?

res/xml/qwertz.xml
```diff
@@ -0,0 +1,44 @@
...
+    <key key0="r" key2="4" key3="$"/>
+    <key key0="t" key2="5" key3="%"/>
+    <key key0="z" key2="6" key3="&amp;" key4="{"/>
+    <key key0="u" key2="7" key3="ü" key4="}"/>
+    <key key0="i" key1="(" key2="8" key4="["/>
+    <key key0="o" key1=")" key2="9" key3="ö" key4="]"/>
+    <key key0="p" key1="=" key2="0" key3="~" key2="\?"/>
+  </row>
+  <row>
+    <key shift="0.5" key0="a" key1="tab" key2="`" key3="ä"/>
```
Julow, line 16: Are `ä` and `ü` useful if they can already be typed with the `trema` accent ?
`é` and `è` are very common in French but I didn't feel the need to add a dedicated key for them, accents are fast enough for me.

res/xml/qwertz.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?> 
+<keyboard>
+  <row>
+    <key key0="q" key1="esc" key2="1" key3="\@" key4="!"/>
+    <key key0="w" key1="^" key2="2" key3="&quot;" key4="°"/>
+    <key key0="e" key1="§" key2="3" key4="€"/>
```
Julow, line 6: `€` is already `Fn+$`. But perhaps we can keep it here anyway.

res/xml/qwertz.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?> 
+<keyboard>
+  <row>
+    <key key0="q" key1="esc" key2="1" key3="\@" key4="!"/>
+    <key key0="w" key1="^" key2="2" key3="&quot;" key4="°"/>
```
Julow, line 5: `°` is already `(accent)^+*`, which is only available for french. What do you think of adding `Fn+*` as a combination for `°` ?

res/xml/qwertz.xml
```diff
@@ -0,0 +1,44 @@
...
+    <key key0="e" key1="§" key2="3" key4="€"/>
+    <key key0="r" key2="4" key3="$"/>
+    <key key0="t" key2="5" key3="%"/>
+    <key key0="z" key2="6" key3="&amp;" key4="{"/>
+    <key key0="u" key2="7" key3="ü" key4="}"/>
+    <key key0="i" key1="(" key2="8" key4="["/>
+    <key key0="o" key1=")" key2="9" key3="ö" key4="]"/>
+    <key key0="p" key1="=" key2="0" key3="~" key2="\?"/>
```
Julow, line 13: `~` is also on the second row.

Julow (commented):
Amazing, thanks !

res/xml/qwertz.xml
Julow: Also the `?` can't be `key2`:
```suggestion
    <key key0="p" key1="=" key2="0" key4="\?"/>
```

Julow:
I've made some changes upstream so I pushed a merge commit to your branch to fix the conflicts.

That's how the new layout looks currently:
image

res/xml/qwertz.xml
Moini: Yes, please! It's needed frequently.

res/xml/qwertz.xml
Moini: If we have proper äöü and ÄÖÜ, we do not need trema - I've never needed it, except for writing Joyeux Noel in French... Please, we do not see the umlauts as accents, they're independent letters of the alphabet.

metadata/android/de-DE/full_description.txt
Moini: I was unsure about the changelog, but if it's not too much, I thought I could keep things up-to-date.

srcs/juloo.keyboard2/KeyValue.java
Moini: Thank you!

res/xml/qwertz.xml
Moini: Maybe? I don't mind too much. Just thought it was practical, as it's easily reachable on a German keyboard by just pressing Shift and then the top left key. It's drawn on the key.

res/xml/qwertz.xml
Moini: ad 1.) Yes, that's one of the duplicates that could be removed – I didn't know about ~ + space key. On a German keyboard, ~ is AltGr +  plus key. Quick one hand typing!

ad 2.) Correct, will fix!

srcs/juloo.keyboard2/KeyValue.java
Moini: I thought it was needed for the Capital letters - to be able to combine with Shift to get ÄÖÜ - is that incorrect?

res/xml/method.xml
```diff
@@ -3,6 +3,7 @@
   <subtype android:label="%s" android:languageTag="en" android:imeSubtypeLocale="en_US" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty"/>
   <subtype android:label="%s" android:languageTag="fr" android:imeSubtypeLocale="fr_FR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=azerty,accents=grave|aigu|circonflexe|cedille|trema"/>
   <subtype android:label="%s" android:languageTag="es" android:imeSubtypeLocale="es_ES" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,accents=aigu|tilde|trema"/>
+  <subtype android:label="%s" android:languageTag="de" android:imeSubtypeLocale="de_DE" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz,accents=grave|aigu|circonflexe"/>
   <subtype android:label="%s" android:languageTag="it" android:imeSubtypeLocale="it_IT" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,accents=grave|aigu"/>
```
Moini, line 7: Oh, I remember now why I wanted the circonflexe... Because the sign (if not combined with a vowel) is also used in maths to designate 'power', as in 2^3 = 8 .

srcs/juloo.keyboard2/KeyValue.java
Julow: It's not needed. Character modifications are first handled by Java's functions and when that's not enough, the rules are in `KeyModifier.java`.

res/xml/qwertz.xml
Julow: I understand that they are independent letters, it's the same in french. But it's not how they are presented in the keyboard, on purpose.
This concept is called a dead key: https://en.wikipedia.org/wiki/Dead_key

res/xml/method.xml
Julow: The `^` should already be on the keyboard because it's an ASCII character. The "circonflexe" key is something else, it is shown as `◌̂`

res/xml/qwertz.xml
Julow: I've changed the combination to `Fn+*`. No need to dedicate a key now.

Julow: > It's drawn on the key.

That's a limitation of this app. There can be up to 5 characters drawn on a key already, it's not possible to draw all the combinations: shift, fn and then all the accents.

res/xml/qwertz.xml
Julow: The `~` on the second row really is the ASCII character `~`. It's not `accent_tilde`, which will be hidden for you most probably.

Julow:
I've added the support for the German language and for the ß in 93e84ba and I've also picked your commit about strings 4b43645. I've pushed an other merge commit to your branch.

Some tweaks are still needed before we can merge the new layout :)

About the translations, could you open a separate PR for them ? I plan to improve the english version first and I don't want to delay this PR more.

Julow:
I've added a few tweaks to the keyboard:
- Removed `°` and fixed `?`
- Removed the umlauts characters, which are accessible with the `◌̈` dead key. This is consistent with how other languages are implemented.

Here's how it looks now:

image

I'm merging now to prepare for the release. Thanks a lot for your work !
Some parts of your initial PR were merged separately: 4b43645c4b6632d0cd2388b90e7bc2bbde98087e 93e84ba7e9d1edc3bd57272f58f28c4a308bcaf2

Moini:
@Julow Thanks! When I understood that this keyboard project will not do what I was looking for, I lost interested. Sorry for not replying sooner.

KloolK:
After reading through this and #4 I'm not sure why the german layout was implemented this way. As @Moini said, the Umlauts are very common in german and requiring two swipes each is very cumbersome.

The screenshot shown here looks exactly like what I think we would need:
https://github.com/Julow/Unexpected-Keyboard/pull/20#issuecomment-1003317825

Why was this chosen instead?
https://github.com/Julow/Unexpected-Keyboard/pull/20#issuecomment-1013744702

Julow:
You are right, I wish to revert that. Are you interested in contributing the Umlauts back ? (and at the same time validate the qwertz layout)

The reason was to make the keyboard multi-language and focused on programming. But this is not reasonable on every layouts, I'll now make the distinction between the localized and the programming layouts.
The dead-key mechanism will be enforced only on programming layouts (only qwerty at the moment) and I'll accept new or modified layouts for every languages.

PR #35, opened 2022-01-22
Use symbols for configuration and emoji keys

eandersons:
`srcs/juloo.keyboard2/KeyValue.java` was updated to replace `Conf` with
`⛭` (`\u2699`) and `:)` with `☻` (`\u263B`).

image


Julow:
This looks a lot better ! Thanks.

The gear symbol is different on my device but that's a known issue that I plan to fix later by shipping a font with the app (it's done for some symbols but it needs to be improved a bit)

PR #47, opened 2022-01-30
Move corner extrakeys out of phone edge

sdrapha:
Hi Jules

First thing, this is not an actual pull request, it's more like a discussion.
I did some changes into my fork, build the apk and I'm starting to use the modified version now as a test.
The changes has more to do with personal preference (MAYBE). So that is why don't feel any pressure to accept these changes to your app. But I wanted to share my viewpoint.
The goal was to move frequently accessed commands away from phone edge, so it's easier to use with a bulky phone case.
Please check [my branch](https://github.com/sdrapha/Unexpected-Keyboard/tree/move-corner-extrakeys-out-of-phone-edge) as a comparison, I built the apk, but I don't know if it can be attached, or you can just build it too. It was my first android apk build ever. Pretty easy using your shell.nix. thanks!


res/xml/qwerty.xml
```diff
@@ -10,18 +10,18 @@
     <key key0="u" key2="7" key3="&amp;"/>
     <key key0="i" key2="8" key3="*"/>
     <key key0="o" key1="accent_macron" key2="9" key3="(" key4=")"/>
-    <key key0="p" key2="0"/>
+    <key key0="p" key1="0"/>
   </row>
   <row>
-    <key shift="0.5" key0="a" key1="tab" key2="`"/>
+    <key shift="0.5" key0="a" key4="tab" key2="`"/>
```
Julow, line 16: `tab` and `l` don't seem to be problematic for qwerty as they are not close to the edge of the screen.

res/xml/qwerty.xml
```diff
@@ -32,13 +32,13 @@
     <key key0="b" key2="\?" key3="/"/>
     <key key0="n" key1="accent_tilde" key2=":" key3=";"/>
     <key key0="m" key2="&quot;" key3="'"/>
-    <key width="1.5" key0="backspace" key2="delete"/>
+    <key width="1.5" key0="backspace" key1="delete"/>
   </row>
   <row height="0.95">
-    <key width="1.8" key0="ctrl" key3="switch_numeric"/>
+    <key width="1.8" key0="ctrl" key2="switch_numeric"/>
```
Julow, line 38: After a bit of testing, I'd say `key4` is easier to hit for this key.

Julow (commented):
This is interesting ! I'll try that for some time too.

What about `esc` and `~` ? These seems to be the most problematic positions

res/xml/qwerty.xml
sdrapha: It's true, that row of keys has some extra margin. but I still felt like an improvement to not hit my finger on the sharp edge of my phone case in an extended and natural finger swipe.
I guess the new tab position I'll have to create some muscle memory first, because I was so used to the old. That is why it's good to test first. 
For the pipe symbol, the adoption of the new position was natural and seems right for me.

res/xml/qwerty.xml
sdrapha: I guess it works too. But I liked the idea of swiping towards the center of the keyboard to change it's layout to numbers, almost like bringing the numbers, if there was an animation for the transition. I'll have to test on position four.

sdrapha:
> This is interesting ! I'll try that for some time too.
> 
> What about `esc` and `~` ? These seems to be the most problematic positions

That is a hard one, I didn't dare to touch these because the key was already FULL in all corners. I guess the ESC could go to a totally different key, but that would be more radical change and may look weird. Maybe top right of 's' (Key2) .... at least the sound has some familiarity,  I don't know....

res/xml/qwerty.xml
Julow: Nice justification :)

sdrapha:
Instead of modifing the main qwerty layout, I changed my aproach to a separate alternative qwerty layout, this way this can be seamless integrated with the current one without radically changing what people are already used to.

I decided to call it qwerty2 since it is an alternative full qwerty (programming layout grade, all chars), not a localization layout.

Also, now there is some logic that enables an alternative nuneric layout to be paired to an existing text layout.

The goal of this layout is still to move the corners extrakeys away from the phone edge so it's easier to use with a bulky phone case.

Last thing, this layout increases the proportions of the navigation key, since it's a wonderful and incredibly useful key, it deserves more screen area.

sdrapha:
Screenshots:
Screenshot_20220317-023607

Screenshot_20220317-023651

Julow:
That's nice thanks! I'm not a fan of adding a new layout however. It's an other layout to maintain (plus the bottom row and the numeric layout) and it won't scale well to new problems/preferences or to localized layouts.
I'd prefer if the main layout was improved or if this was solved in a different way. My policy would be to accept new options if they are simple to explain but to reject invasive changes.

On your screenshots you use a very thin margin on the sides. What do you think of increasing that option ? Combined with a few tweaks to the layout, this might be enough to fix your issue.

sdrapha:
> 
> On your screenshots you use a very thin margin on the sides. What do you think of increasing that option ? Combined with a few tweaks to the layout, this might be enough to fix your issue.

Regular layout without modifications, adding horizontal margins:
</img>

To make the regular qwerty layout less worse for me (regarding the edge issue only, ((I still love this keyboard))), by using margins, I need to bump the margins to `20dp`, and change the swipe distance from `Far` to `Normal`.
By doing this, I loose a lot of screen area, the keys become skinner and the precision also drops with such tiny keys. AND the problematic corners keys still a problem for me after all that.

To make a summary of the biggest annoyances from the original layout for ME:
- Half the times I try to forward delete something, it ends up backspacing delete
- Numbers often get type like `2p22` where it should be `2022`
- The switch to numeric pad, is way easier to reach on the new position, old position the phone almost falls off hand if not using both hands together.
- I found out how useful `Esc` is since it moved to an easier location, I didn't know I would use this key so frequently until that change.
- Tab also got easier to swipe, and it is a really frequent used key
- Bigger navigation keys, is not a breaking change, I guess this change in particular could be done on the original layout without conflicts.

I know it may be personal, and I agree with the point of invasive changes policy. 
That is why since the beginning I was cautious and I don't want to push anything. 
The change for a separate layout was an attempt to not be invasive on a current already established layout.

I'll keep using this qwerty2 layout myself, and maintaining it on my personal fork for my use.
If  this pull request has no future, feel free to close it. No problem 🙂

sdrapha:
Hi, out of the scope of this PR, I just wanted to share an extra experimental layout I've been playing with. If you have a chance and time to check that out.
It's in a separate branch 
https://github.com/sdrapha/Unexpected-Keyboard/tree/test-layouts-variations

There are a bunch of layouts on that build.
The one in particular I like the most so far is QWERTY6, see if there is anything you like.

Last CI run for that branch with the built debug apk:
https://github.com/sdrapha/Unexpected-Keyboard/actions/runs/2046110485

Note that on that branch there will be two apks, the fork-debug apk can be installed in parallel to an existing debug apk

Julow:
Thanks for all your ideas! I've modified all the layouts with inspiration from your `test-layouts-variations` branch.
See d62e764
I've done some changes to the bottom row too, 695f5b4, b85b67c

I would like to release soon but I'd appreciate your feedback.

sdrapha:
oh, cool! 
- I really liked your idea to swap Alt with Fn since I use the Home and End keys so often!
- Also appreciate the easier access for switching between regular layout and numeric layout. But I also miss the easy access to Ctrl that I also use all the time; it is still acceptable in the new location though. 
(For me the ideal location for the numeric switch key is swipping up on the "space" key, or "0" key to come back, but I acknowledge that I changed the numeric keyboard to better accomplish that, otherwise the "0" key would not be centered enough in the screen. With that approach, the Ctrl key could remain in place as before)

But overall, those last commits changes are already an improvement. In my humble oppinion. 
The ideal layout for me is still a bit different but it's hard to judge what is an improvement for everybody vs personal preference.

Julow:
Thanks for your feedback. I couldn't stand the new location of the Ctrl key, reverted 2b348b7.

sdrapha:
> Thanks for your feedback. I couldn't stand the new location of the Ctrl key, reverted [2b348b7](https://github.com/Julow/Unexpected-Keyboard/commit/2b348b7a31d1b0d3a2f95280a43d3670d087ee4f).

Agreed, it is better now, after 2b348b7

Julow:
There's not many keys near the edges of the screen anymore and I'm keeping this in mind every time a new layout is proposed.

I close this PR but I'm still interested on whether the latest version is fine for you. I'll consider this problem solved for now.

sdrapha:
Yeah thanks, I will still keep updating and using my personal layout as the other changes I made still a must for me, especially the position of the `switch_to_numeric` and back.
But the idea of moving important keys away from the edge got trough, apreciate that.(ps. except delete,and action_key that still on the edge)
Also, after #138, it got better. The swipping at hard to reach areas are more tolerable.
If there is anything else from my alternative layout that is desirable please say so then I can do a more focused pr, but only if you want, other than that I'll refrain myself from requesting changes like that.

Julow:
Since https://github.com/Julow/Unexpected-Keyboard/pull/138, I don't think it's worth moving delete and the action key anymore. I have nothing specific in mind but can you look at the layouts and tell me what prevents you from switching back to the defaults ?

PR #48, opened 2022-01-30
Sticky ctrl alt

sdrapha:
Hey, same as my other pull request, more like an discussion.
Don't feel you need to like this idea.

I made my Ctrl and Alt able to Stick ON, just like the Shift already does.

So I can keep them pressed while using the arrow keys to move code around

https://user-images.githubusercontent.com/25059996/151714318-a145748b-d3aa-4a21-96f3-117bdf7d81b4.mp4

Also this branch is based on my other branch, if needed I can rebase. So the extra commits go away.

### DIFF
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -1,0 +1,36 @@
+name: Make Apk CI
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  Build-Apk:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install nix
+      uses: cachix/install-nix-action@v15
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - uses: cachix/cachix-action@v10
+      with:
+        name: julow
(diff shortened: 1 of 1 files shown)


Julow:
I'm not a fan of sticky modifiers, I'm even considering removing it for Shift. Perhaps that can be an option.

Note that you can keep pressing Ctrl or Alt and use the arrows with your other hand.

sdrapha:
Yeah I know I can use the other hand, when it's available, doing that for a while , as a plus on the scenario of holding alt continuosly with the other hand, it already happened to me to my finger move a bit on top of alt, bacaming an FN press, while I was using the arrows, so it messed it up good what I was doing, that happened many times, even more often before the precision on this keyboard was improved just recently.

**BUT, an option in the settings would be GREAT, so it caters to all.**

It would also be hard to hold two keys together with one hand and use the other for the arrows. Without the sticking capability

https://user-images.githubusercontent.com/25059996/151716855-a1f5c012-dbb2-4303-96c3-65510ecf3556.mp4

Julow:
Sublime text-like selection is a good argument !

Issue #56, opened 2022-02-02
Unintuitive arrow key


MaxGyver83:
Thanks for this awesome app! I really like the concept of swiping to the key corners for accessing additional keys!

Except for the arrow key:

![Arrow key](https://user-images.githubusercontent.com/13970628/152232543-2a912032-09e5-4c28-bb35-1e9475cf9d56.png)

It's not intuitive at all swiping diagonally if you mean left, right, up or down. It would make more sense this way:

![Arrow key edges](https://user-images.githubusercontent.com/13970628/152232629-5d0d63c5-7013-4677-bcee-9baf224af361.png)

I have already implemented this. Just wait some minutes for my pull request.

avently:
disagree here. i figured out how to use the keyboard even without reading the docs and the first thing i did is swiped on the arrow keys. so it's intuitive, cool and consistent with other keys

sdrapha:
The old arrows worked fine for me also, @avently. BUT the HOME, END, PG UP and PG DOWN, were always confusing. I merged @MaxGyver83 changes to my fork and I'm currently using it. The Home/END/PGUP/DOWN works better now.  For the regular arrow keys, I keep going diagonally because of muscle memory, but I want to like this new layout, make sense to me even if it's not consistent with the other keys. 
Well, I'm not concluding anything, I just need to test it for a little while longer.

Julow:
Merged :)

PR #57, opened 2022-02-02
Allow edge keys instead of corner keys (swipe vertically/horizontally)

MaxGyver83:
Fix/implement #56.

Add a new boolean parameter "edgekeys" for defining keys that have the additional (swipe) keys on the edges (top, right, left, bottom) instead of at the corners (top left, top right, bottom left, bottom right).

Example:

```xml
<key width="1.2" key1="up" key2="right" key3="left" key4="down" edgekeys="true"/>
```

If the `edgekeys` parameter is omitted, it defaults to `false` (swipe diagonally for additional keys).

This pull request also changes the arrow keys in all layouts to the alternative edge style:

Before:
Arrow key

After:
Arrow key edges

### DIFF
--- a/res/xml/azerty.xml
+++ b/res/xml/azerty.xml
@@ -38,7 +38,7 @@
     <key width="1.8" key0="ctrl" key3="switch_numeric"/>
     <key width="1.2" key0="alt" key1="fn" key2="change_method" key3="switch_emoji" key4="config"/>
     <key width="4.0" key0="space"/>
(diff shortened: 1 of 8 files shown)


Julow:
I've used this since yesterday and it's working great :) I do prefer the arrows now, though I don't use them much.
Thanks a lot !

@avently mentioned that the arrow keys were already fine but I think vertical/horizontal movements are easier than diagonals at the bottom of the screen in addition to being more intuitive. I'm also investigating improving cursor movements and this will help a lot.

PR #59, opened 2022-02-02
Replace unusual return symbol

### LINKED ISSUE #53: Return/enter symbol upside down
The standard return symbol is ⏎, but UK's goes upwards instead.

MaxGyver83:
Fix #53.

Before:
return key upside down

After:
Screenshot_2022-02-02-22-15-37-09_c7d7c78e17792d54d6e4296d9f6a5da4

### DIFF
--- a/srcs/juloo.keyboard2/KeyValue.java
+++ b/srcs/juloo.keyboard2/KeyValue.java
@@ -224,7 +224,7 @@ class KeyValue
     addSpecialKey("action", "Action", EVENT_ACTION); // Will always be replaced
 
     addEventKey("esc", "Esc", KeyEvent.KEYCODE_ESCAPE);
-    addEventKey("enter", "\uE800", KeyEvent.KEYCODE_ENTER, FLAG_KEY_FONT);
+    addEventKey("enter", "\u23CE", KeyEvent.KEYCODE_ENTER, FLAG_KEY_FONT);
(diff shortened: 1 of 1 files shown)


sdrapha:
Instead of replacing the char for a different unicode char. Could the change be done on the current icon inside the .ttf instead? This way it could be a nicer looking symbol, not just the outline. I'm also afraid of plain unicode chars looking different across devices, with different system fonts. There was an old version of this keyboard that the PG_UP and PG_DOWN rendered as weird blue squares icons that was part of the system font on my phone.
I'm not sure if that is really an issue, but wanted to share my viewpoint.

MaxGyver83:
> I'm also afraid of plain unicode chars looking different across devices, with different system fonts.

Good point!

> Could the change be done on the current icon inside the .ttf instead?

I think this is possible. I just don't know how to modifiy a .ttf file. But I'm sure @Julow knows how to do it.

Julow:
@sdrapha is right, the included ttf needs to be changed too. And I'd prefer a full and lighter symbol too.
But the char needs to be changed too, so the font is not too wrong.
I'll merge this and work on the font later, which needs more work than just modifying a glyph. I'd like to keep the glyph in text format (perhaps svg) and build the font deterministically because font editing softwares are a pain to use.

Thanks !

PR #60, opened 2022-02-03
Create github action to compile debug_APK

sdrapha:
Make it easy to compile a debug apk for testing with the latest available code by using a Github Action.
It can be manually run from the actions tab, and the result APK is saved as an artifact of the run.

Having this action on the master branch is also good so every forked branch will also have it which makes testing really easy, and avoids any difficulties of setting up the correct environment


.github/workflows/make-apk.yml
```diff
@@ -0,0 +1,30 @@
+name: Make Apk CI
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
Julow, line 5: This is great for power usage. To make things easier, do you know if there's a way to trigger the workflow directly from a PR page ?

Julow (commented):
That's a really great idea !

There's an inconvenient however, the debug APK is signed with a temporary key that is generated by the Makefile automatically. With this setup, the key will always change and Android will refuse to install the application unless the previous version is uninstalled first.

This can be fixed by caching the file `_build/debug.keystore` between runs. Do you think that's possible?

.github/workflows/make-apk.yml
sdrapha: this will trigger the action manually and also automatically with a PR
```yaml
on:  
  workflow_dispatch:
  pull_request:
```

you can also do this
```yaml
on:  
  workflow_dispatch:
  push:
  pull_request:
```
so any pushed commit will trigger the action and generate a new debug_apk automatically

sdrapha:
>debug APK is signed with a temporary key that is generated by the Makefile automatically. With this setup, the key will always change 

That is true, not ideal, that issue fixed would make this action much better.

Probably there is more than one way to accomplish that.
If I understood correctly, the debug key is not really secret right?  specially for an open source app, so including `_build/debug.keystore` into the repo, that would always be there when the action git clones the repo.
I know there is also a way of using github Secrets to inject non public keys into an action, but I never used.
I'm learning as I go. All new stuff to me.

Julow:
I've played a bit with it and caching the debug certificate was straightforward.
I also added cachix to cache the few things that have to be built but it didn't help much. The build time is mostly the same but it's now pulling 450Mb from the caches instead of 300Mb on every runs. I'll try to lower the size of the final closure later or, if I don't manage to, I'll disable the build on every pushes.

Julow:
Thanks a lot :) Merged now.

sdrapha:
Awesome!
Ps. ( food for thought, I'm curious now how that is gonna play out on forked repos, if each forked repo needs to create its ows secret for `CACHIX_SIGNING_KEY`, I would guess so, I'm going to do some tests if I get some available time)
Thank you again!

Issue #46, opened 2022-01-28
[feature] swipe delete


Quantenzitrone:
https://user-images.githubusercontent.com/74491719/151580765-c79ab179-e3c0-4618-9c0c-6414d38b7ed4.mp4

like in this video of OpenBoard.

Julow:
This would be an awesome feature. (similar suggestion: https://github.com/Julow/Unexpected-Keyboard/issues/7)

However, I don't want to copy this idea as it is because it wouldn't be the best fit for this keyboard. There's already a "precise repeat" feature that is not working well and I planned to improve it at some point.

Julow:
The "precise repeat" feature will be improved during the next release but maybe not enough to close this issue yet. (c85e9a9)

Quantenzitrone:
i think this issue is resolved enough with "precise repeat" and Ctrl+Backspace (delete word)

Julow:
You might be interested in https://github.com/Julow/Unexpected-Keyboard/issues/7#issuecomment-1399625673

moya2162:
I would like to add my opinion that this feature would be nice to have as well.

Swipe to move cursor which was recently added is a huge welcome. Shift + swipe to highlight a word is nice as well when trying to specifically delete a word somewhere in a sentence. 

I would like to add that for speed and ergonomics, it would be nice to have swipe on delete to quickly delete a mistake. Its faster than having to hit shift, highlight, hit delete which is 3 gestures vs 1. My humble 2 cents.

Julow:
This seems nice indeed.

mexicancartel:
Came here to request this feature. I keeps switching back to simple keyboard for regular use because of this.

Sx1Fl0NdqlkM6N4s2pswXKUI0k9FKTaOckLbNXu:
I don't know if keys like "select all" are somehow fundamentally different than "key = key combination", but having access to those would mostly solve the issue:

ctrl+backspace, ctrl+delete = delete word
ctrl+left, ctrl+right = move by word
ctrl+shift+left, ctrl+shift+right = select word
shift+left, shift+right = select character

unless we can already send key combinations?

ChristianKurz:
I'm also missing this feature from simple keyboard. I don't  think key combinations are a good replacement - swipe deletion with selecfion is fast and very precise.

evantj:
setting w="selection_cursor_left" on the backspace key and then just typing over the selection works pretty well for me

[...]

PR #61, opened 2022-02-03
Add delete-word-key

MaxGyver83:
Implement a delete-word key (=`Ctrl-backspace`) and add it as top-left corner key to the backspace key (in all layouts). This does not exactly implement #46 but it makes deleting a whole word much easier.

Screenshot_2022-02-03-19-09-04-22_c7d7c78e17792d54d6e4296d9f6a5da4

Maybe you have an idea for a better symbol than `⌫⌫` for deleting a whole word.

### DIFF
--- a/res/xml/azerty.xml
+++ b/res/xml/azerty.xml
@@ -32,7 +32,7 @@
     <key key0="v" key3=";" key4="."/>
     <key key0="b" key3=":" key4="/"/>
     <key key0="n" key1="accent_tilde" key2="§" key4="!"/>
-    <key width="2.0" key0="backspace" key2="delete"/>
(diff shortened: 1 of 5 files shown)


Julow:
This doesn't feel like it's helping much. What's the problem with using the Ctrl key ?

I'd rather have a better answer to #46 but at the moment I don't have a satisfying plan, though.

MaxGyver83:
> This doesn't feel like it's helping much. What's the problem with using the Ctrl key ?

In my opinion, swiping the backspace key is easier than pressing first Ctrl, then backspace, especially when typing one-handed.

But I would also love to see the solution shown in #46.

Julow:
I think it is not an ideal solution. Before we have #46, the solution is to hold Ctrl.

PR #66, opened 2022-02-06
Add Russian layout

carrot-cookie:
Greetings. 
I made a Russian layout. It looks like this:
image

In my opinion, the right side is empty, and there we can either move additional keys from the center, or add something new


res/xml/qwerty_ru.xml
```diff
@@ -0,0 +1,50 @@
...
+  <row>
+    <key key0="й" key1="esc" key2="1" key3="~" key4="!"/>
+    <key key0="ц" key2="2" key3="\@"/>
+    <key key0="у" key2="3" key3="\#"/>
+    <key key0="к" key2="4" key3="$"/>
+    <key key0="е" key1="ё" key2="5" key3="%"/>
+    <key key0="н" key2="6" key3="^"/>
+    <key key0="г" key2="7" key3="&amp;"/>
+    <key key0="ш" key2="8" key3="*"/>
+    <key key0="щ" key1="accent_macron" key2="9" key3="(" key4=")"/>
```
Julow, line 12: You don't need to include any accents, none of them will have any effects on Russian letters. The `accent_...` keys are only meant for the "programming" layouts, currently only `qwerty`.

res/values/arrays.xml
```diff
@@ -5,6 +5,7 @@
     <item>azerty</item>
     <item>qwerty</item>
     <item>qwerty_lv</item>
+    <item>qwerty_ru</item>
```
Julow, line 8: I think this layout isn't named qwerty. Wikipedia calls it [JCUKEN](https://en.wikipedia.org/wiki/JCUKEN).

There's an other layout that would be attributed the name "russian qwerty" ([phonetic layout](https://en.wikipedia.org/wiki/Phonetic_keyboard_layout)) so keep the name free.

res/xml/qwerty_ru.xml
```diff
@@ -0,0 +1,50 @@
...
+    <key key0="з" key2="0"/>
+    <key key0="х"/>
+    <key key0="ъ"/>
+  </row>
+  <row>
+    <key shift="0.5" key0="ф" key1="tab" key2="`"/>
+    <key key0="ы" key1="accent_ring" key3="ß"/>
```
Julow, line 19: No need to include `ß` either.

Julow (commented):
Thanks a lot!

The symbols keys look stacked on the left, I think it would be better to spread them a little, especially the top row. At the same time, avoiding keys at the left and right edges of the keyboard, which are harder to type.

carrot-cookie:
Fixed the problems above. Here's what happened
image

Julow:
There have been some changes on the master branch so I fixed the conflicts and I renamed the layout.

Thanks a lot for the work :) Merging !

PR #67, opened 2022-02-07
Invert return key in TTF

Roy-Orbison:
Was a "level-up" symbol, now looks like a return key in identical style.

I just flipped the glyph in FontForge. I hope it's okay, but if not I can change the font generation settings.


Roy-Orbison:
Pursuant to #53 & #59.

PR #72, opened 2022-02-14
Clicking twice on CTRL or ALT will lock them in 'ON' state

sdrapha:
Previous pull request #48 got closed accidentally/automatically while I was trying to rebase the commits to the latest code.
Reopenig it again....


sdrapha:
Hi @Julow , I made the shift,ctrl,alt locking behaviour optional and configurable in the settings.
It is working great on my tests.
I don't know how to achieve the same with less lines of code (talking about the switch,case block), I hope my best effort is good enough. It is working for me.

Food for thought: Since I had the preferences settings for the each LOCK-able key, I could get rid, of the FLAG_LOCK entirely from the class definition, BUT I really don't want to take this capability away, since it may be still usefull to use it for different keys (example: Fn, Superscript, Underscript, Meta) which doesn't have the dedicated preferences setting counterpart.

For now, I'm done doing changes in this branch,pull request, I'm happy with the current status.

srcs/juloo.keyboard2/Config.java
```diff
@@ -60,6 +63,9 @@ private Config(Context context, IKeyEventHandler h)
     keyVerticalInterval = res.getDimension(R.dimen.key_vertical_interval);
     keyHorizontalInterval = res.getDimension(R.dimen.key_horizontal_interval);
     preciseRepeat = true;
+    lockShift = true;
+    lockCtrl = true;
+    lockAlt = true;
```
Julow, line 68: This is the default value when the option is unset, should be `false` (except for shift).

res/xml/settings.xml
```diff
@@ -9,6 +9,9 @@
     <juloo.common.IntSlideBarPreference android:key="longpress_timeout" android:title="@string/pref_long_timeout_title" android:summary="%sms" android:defaultValue="600" min="50" max="2000"/>
     <juloo.common.IntSlideBarPreference android:key="longpress_interval" android:title="@string/pref_long_interval_title" android:summary="%sms" android:defaultValue="25" min="5" max="100"/>
     <CheckBoxPreference android:key="precise_repeat" android:title="@string/pref_precise_repeat_title" android:summary="@string/pref_precise_repeat_summary" android:defaultValue="true"/>
+    <CheckBoxPreference android:key="lockable_shift" android:title="@string/pref_lockable_shift_title" android:summary="@string/pref_lockable_shift_summary" android:defaultValue="true"/>
+    <CheckBoxPreference android:key="lockable_ctrl" android:title="@string/pref_lockable_ctrl_title" android:summary="@string/pref_lockable_ctrl_summary" android:defaultValue="true"/>
+    <CheckBoxPreference android:key="lockable_alt" android:title="@string/pref_lockable_alt_title" android:summary="@string/pref_lockable_alt_summary" android:defaultValue="true"/>
```
Julow, line 14: Default value should be the most usual one: true for shift, false for ctrl and alt.

Julow (commented):
Thanks for the exploration. That's not an ideal implementation because it adds a lot of code and specific cases to a class already doing complicated things.
I'd suggest this alternative implementation:
The keys declared in `KeyValue` are unchanged and the code in `Pointer` is unchanged too.
In `Keyboard2View.setKeyboard`, there's code that modify the layout depending on the config. (you can move that code to a new function, `Config.modify_layout`)
You can "replace" the keys depending on the new option here, with or without the `FLAG_LOCK` flag.

Otherwise, I like the idea ! I'll happily implement the modification I suggested if you don't have the time.

sdrapha:
Suddenly there is too much in my life happening now that I won't have time to explore the alternative any time soon. Feel free to take over if you want. Thanks a lot!

Julow:
No problem! Thanks for starting the work.

Julow:
I've improved the way the keyboard is dynamically modified (e92504ae) and adapted the PR to use that. I've turned the settings into a separate page (not a popup for compatibility) because that was taking too much space on the page and added too many strings to translate.

sdrapha:
Awesome! Thanks so much!
I finally got a chance to have a look at the github messages (on the phone only, I'm still away from computers)
I did install the debug apk from the last actions run. I liked the settings nested page.
Unfortunally it seems that there is something weird going on with the lockable keys. They don't consistently turns blue and then green. most of time they don't change colors when pressed at all, and then if you press ALL of them, then the keyboard renders the colors. It's a bit erratic on the displaying side, the behaviour works fine though.

Julow:
I fixed the inconsistent colors (2ea256e).

sdrapha:
I tested and can confirm it's working fine now, awesome! Thanks!!! 😃

sdrapha:
Ops. There was a catch. I found a bug now with the latest commit. Anytime I try to use the arrow keys together with a modifier key (ctrl, fn, or alt) the keyboard crashes.

Julow:
Thanks, fixed in f4978f5

rbreaves:
Some reason I am not seeing this as an option "Modifier that can be locked by typing them twice". Was it taken back out? I would like to have this feature as I am a heavy sublime text user and I even add its keymaps to vscode.

Julow:
This feature has indeed been replaced by holding to lock. Glad this works for you.

PR #73, opened 2022-02-17
add colon symbol to numeric keyboard

sdrapha:
The colon symbol : is used together with numbers really all the time, it should be available in the numeric keypad


Julow (approved):
Thanks!

PR #74, opened 2022-02-17
add-ordinal-numbers-symbol-system

sdrapha:
Add a way to input ordinal markers, using the accent system inside the numeric keypad.
So we can type 1ª (primeira) , 2º (segundo) , 1ˢᵗ , 2ⁿᵈ , 3ʳᵈ, 20ᵗʰ, 50ʳᵉ, 60ᵈᵉʳ
(note that especially the ª º are NOT superscript 'a o'. they are specific symbols for ordinals)
The other letters can be combined for a broad range of different languages ᵉ ᵉʳᵈ ᵈᵉʳ ʳᵉ ˢᵗ ʳᵉˢ ᵈᵉʳˢ ᵉˢ ᵈˢ ⁿᵈ ʳᵈ ᵗʰ

Screenshot_20220217-063733_Brave_1

Screenshot_20220217-063749_Brave_1

### DIFF
--- a/res/xml/numeric.xml
+++ b/res/xml/numeric.xml
@@ -20,7 +20,7 @@
   </row>
   <row>
     <key shift="0.35" width="1.15" key0="shift" key2="fn" key4="alt"/>
-    <key key0="1" key1="superscript" key3="subscript"/>
+    <key key0="1" key1="superscript" key2="ordinal" key3="subscript"/>
     <key key0="2"/>
     <key key0="3"/>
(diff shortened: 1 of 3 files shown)


Julow (approved):
Thanks a lot !

It's unfortunate that `ªº` are not always rendered with the underline.

It's missing the french "2ème" but I'm not sure anyone write that in superscript. "2ᵉ" and "2°" are more common. The later is `Fn+*` currently, so I'll add `Ord+*` as a synonym.

PR #77, opened 2022-02-25
Add a missing full stop to Latvian QWERTY layout

eandersons:
`res/xml/qwerty_lv.xml` was updated to add the missing full stop character (`.`) as it got lost when the bottom row was moved to a separate XML `res/xml/bottom_row.xml` as it was placed on the arrows key.
Additionally `?` was relocated, giving its place to the full stop.


Julow (approved):
Thanks! Sorry for the regression.

Issue #79, opened 2022-02-25
Inconsistent symbols across different devices


sdrapha:
Different devices have different system fonts. So they don't always render the symbols as expected.
For example, in my Galaxy Tab S6:
![image](https://user-images.githubusercontent.com/25059996/155744868-53bd771a-a0fd-443d-948f-ea6dcfd947b2.png)

So what is supposed to be a Cog symbol for the settings, renders as a brightness? icon, the emoji key also doesnt't render as expected

I already made changes to that and I will follow up with a pull request

Julow:
#80 is merged so I close this. The issue remains for some other symbols https://github.com/Julow/Unexpected-Keyboard/issues/82

Issue #78, opened 2022-02-25
Lable size of secondary symbols


ChiefMikeK:
### Maybe increase the contrast and use standard or web safe colors ` ¯\_(ツ)_/¯ `   
     res/values/themes.xml

```
<style name="Dark">
30 ~ <item name="colorLabel">#FFFFFF</item>
33 ~ <item name="colorSubLabel">#A0A0A0</item>  
    #9e9e9e  

<style name="Light">
46 ~ <item name="colorLabel">#000000</item>
49 ~ <item name="colorSubLabel">#5e5e5e</item>  
    #616161  

<style name="Black">
62 ~ <item name="colorLabel">#eeeeee</item>
65 ~ <item name="colorSubLabel">#bbbbbb</item>  
    #eeeeee

```

ChiefMikeK:
The lable size of the secondary symbols are too small, especially (){} ., and should be adjusted via a seperate settings
![Screenshot1](https://user-images.githubusercontent.com/20070053/155745556-35ab887e-9a86-4133-b448-e51f8a542b5f.png)
![Screenshot_2](https://user-images.githubusercontent.com/20070053/155745591-5537d52c-0d50-48c0-9d38-e3b7f174f9ca.png)

Julow:
I definitely agree ! The label sizes need to be reworked a bit and I'll improve the color too. Thanks for the suggestions.

Julow:
I've made improvements to the way the label size was computed and increased the contrast slightly (23685dd, 1451857, f6d4554). Here's how the keyboard looks now:

![screen](https://user-images.githubusercontent.com/2310568/155886556-81c9b442-67b5-413a-bace-47f4ef10b827.png)

PR #80, opened 2022-02-25
Update symbols from unicode to font-Icon

sdrapha:
This PR fixes issue #79 

Currently some of the symbols don't render consistently across different devices because they are unicode symbols, which render unpredictably with different phone's system fonts.

I'm updating them to use font-icons that I added in the included embed font from this keyboard

So the Emoji and Settings Cog will always render consistently,
also once I was already adding icons, I updated the Home, End, PageUp and PageDown too, to better represent them
image


ChiefMikeK:
could you increase the contrast and post a screen shot here 🤔 as i've suggested here?

https://github.com/Julow/Unexpected-Keyboard/issues/78#issue-1150599968

sdrapha:
> could you increase the contrast and post a screen shot here 🤔 as i've suggested here?
> 
> [#78 (comment)](https://github.com/Julow/Unexpected-Keyboard/issues/78#issue-1150599968)

that is something else. You are talking about color themes. This PR doesn't touch any theme definition at all, it's separate from that.

ChiefMikeK:
And sorry to bother you but, several dead keys render poorly  
https://github.com/Julow/Unexpected-Keyboard/blob/7b04048c3e951027bf186c56790c8d3cddc55742/srcs/juloo.keyboard2/KeyValue.java#L162
macron

ChiefMikeK:
> > could you increase the contrast and post a screen shot here 🤔 as i've suggested here?
> > [#78 (comment)](https://github.com/Julow/Unexpected-Keyboard/issues/78#issue-1150599968)
> 
> that is something else. You are talking about color themes. This PR doesn't touch any theme definition at all, it's separate from that.

ok I would fix myself but me internet is very poor here

sdrapha:
That I agree. Same thing I'm doing in this PR could be done for more icons. 
That could be done later, following the same scheme. Of course if @Julow agreeing to all these changes.
I won't have time to change this myself right now. 
Maybe I could try helping with that later. 
I basically used FontForge software to edit the .ttf
You can also check the current font-icons in the .ttf using VSCode with the fonts extension installed. 
For the Home and End symbols, I created them entirely in FontForge. Also, there are some websites that can provide some free licensed font-icons like Fontenello and Icomoon. I used some from them, copy and pasted the specific icon from their .ttf into this keyboard .ttf using FontForge, and fixed the size proportions and position to match the existing ones.

Julow:
Thanks a lot ! The glyphs you chose look very nice. It's generally good practice to credit the original author of such things (and sometimes mandated by the license), where do they come exactly ?

Many other glyphs need to be improved and I planned to do that eventually. Editing the font file manually is not ideal and I'm looking at other solutions, that's why it's taking so much time. On the long term, I'd like the font file to be reproducibly built from text descriptions like svg, or not using a font at all.

Julow:
Thanks. I've added that to a README.

Issue #83, opened 2022-02-25
Max allowed char label size still too small for big screens


sdrapha:
In my tablets this keyboard doesn't scale up well for higher resolutions, this screenshot is taken already using the max allowed char label size of 1.2. It's still small.

<img src=https://user-images.githubusercontent.com/25059996/155810309-f482b2c9-dfe0-4129-9660-121440e763f7.png height=800px>

I already done some changes and will follow up with PR.

I noticed issue #78 that may ride along with this one, but the request there is not exactly the same from this issue here.

Julow:
The label size is now relative to the height of the keyboard (23685dd) and should look right by default (when the label size setting is set to `1`).

PR #84, opened 2022-02-25
Increase max size of char label, for big screens

sdrapha:
This fix issue #83

Also issue #78 may or may not benefit from this.

Increasing the max allowed char label size. The current max is too small for tablet resolutions.

I also added a description, so if user goes overboard with the label size setting in a small screen and it ends up looking weird, there is the reason why.

image


Julow:
Thanks! This is a possible workaround but I'd like to fix the issue at its root. The sizes should be right by default, I'll have a try during the weekend.

Julow:
This should now be solved without tweaking the options. Thanks for the investigation!

PR #91, opened 2022-03-02
Brazilian portuguese layout

igorSilCar:
Providing a comprehensible Brazilian Portuguese keyboard layout


res/xml/method.xml
```diff
@@ -8,7 +8,7 @@
...
   <subtype android:label="%s" android:languageTag="it" android:imeSubtypeLocale="it_IT" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,extra_keys=grave|aigu|euro"/>
   <subtype android:label="%s" android:languageTag="lv" android:imeSubtypeLocale="lv_LV" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_lv,extra_keys=caron|cedille|macron|euro"/>
-  <subtype android:label="%s" android:languageTag="pt" android:imeSubtypeLocale="pt_BR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,extra_keys=aigu|cedille|circonflexe|grave|tilde|euro"/>
+  <subtype android:label="%s" android:languageTag="pt" android:imeSubtypeLocale="pt_BR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,extra_keys=aigu|circonflexe|grave|tilde|euro"/>
```
Julow, line 11: Also need to specify the layout corresponding to the locale:
```suggestion
  <subtype android:label="%s" android:languageTag="pt" android:imeSubtypeLocale="pt_BR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_pt,extra_keys=aigu|circonflexe|grave|tilde|euro"/>
```
Thanks for the fix.

res/xml/qwerty_pt.xml
```diff
@@ -0,0 +1,38 @@
...
+    <key key0="t" key2="5" key3="%"/>
+    <key key0="y" key2="6" key3="^" key4="accent_trema"/>
+    <key key0="u" key2="7" key3="&amp;"/>
+    <key key0="i" key2="8" key3="*"/>
+    <key key0="o" key2="9" key3="(" key4=")"/>
+    <key key0="p" key2="0" key3="\\" key4="|"/>
+  </row>
+  <row>
+    <key key0="a" key1="tab" key2="'" key3="`" key4="&quot;"/>
+    <key key0="s" key1="accent_ring" key2="accent_caron" key3="ß" key4="accent_cedille"/>
```
Julow, line 17: I think you can remove the accents and use the room to add the accented letters directly. The accents are meant for the "international qwerty" layout.

Julow (commented):
Thanks :) I have a few comments before merging.

res/xml/qwerty_pt.xml
igorSilCar: I though those were to indicate to the keyboard where to put the accents for "show accents for all languages" and I personally prefer using dead keys (lines up with how I use my PC keyboard). In any case, I'll revise it and see how it plays out. Drafting for now.

igorSilCar: @Julow, upon testing, I've confirmed the accents declared on the layout are being used by the "show accents for all installed languages" and "show all accents" options to decide where to show accents, besides the "show accents for the selected language" option, which will show only the declared at res/xml/method.xml.
From [bb5f237](https://github.com/Julow/Unexpected-Keyboard/pull/91/commits/bb5f237cebdd1c3bf092c8ba17e40feddd4e390a) I can provide all the accents but I feel it will get the keyboard visually cluttered (might not be an issue, since if the user wants to get all the accents they need to be shown) or revert to the dead keys, but I will decided on it after further feedback from you and some close friends whom I enlisted to test the layout.

Julow: Yes the "show all accents" option works with the `accent_*` keys. There's tension between having dead-keys or dedicated keys. Many people prefer the later so I've accepted localized layouts without dead keys on them but this is not a requirement.

If you use dead keys on your PC keyboard, I think it's better to have dead keys here too! Are they common in Brazil ?
There's no strong rules on how layouts should be yet, sorry for the wrong suggestion, I didn't anticipate how the keyboard would look.

Julow: I've reverted that commit.

Julow:
Thanks! Merged.

res/xml/qwerty_pt.xml
igorSilCar: Sorry for being mia for this release cycle. I'm totally on board for the revert! About dead keys, they're quite common for developers in Brazil, so I think it's fitting for the target audience. Thank you for accepting the PR, I might come back to it latter this week.

PR #100, opened 2022-03-08
Brazilian portuguese translations

igorSilCar:
Providing translation for the new setting under the app settings.


Julow (approved):
Thanks!

PR #106, opened 2022-03-19
Translate into Chinese-Simplified

9-2-1:
Translated into Chinese-Simplified (res/value-zh-rCN).


Julow (approved):
Very appreciated, thanks!

PR #107, opened 2022-03-20
Miscellaneous tweaks

ArenaL5:
First of all, thank you for creating this keyboard and thank F-Droid for listing it. I wouldn't ever have known this was even a thing otherwise.

This is a bunch of tweaks I find useful, and a fix for a minor display glitch. None of these tweaks looked major enough to warrant their own pull request, but I think all of them will be useful.

Please let me know if you want me to edit or drop some commits.


Julow (approved):
Awesome, thanks!

PR #111, opened 2022-03-25
Upgrade CI workflow

sdrapha:
I got in this rabbit hole, and spent my night without sleeping, but I ended with a working CI flow, which utilizes caches and runs in 40sec after the first cache is done.
The steps are simplified, and there is no dirty OS upgrade anymore in order to get a newer version of FontForge. The new FontForge is self-contained and has no dependency.

Only doubt I have is if its a good thing to cache the whole folder /usr/local/bin, but it's working well so far.

### DIFF
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   Build-Apk:
-    runs-on: ubuntu-20.04 # That is the latest OS version available today
+    runs-on: ubuntu-latest
     steps:
     - name: Cache fontforge and extra dependencies
       id: cache_localbin
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -7,17 +7,23 @@ on:
 
 jobs:
   Build-Apk:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-20.04 # That is the latest OS version available today
     steps:
-    - name: Tweaks to get more recent version of fontforge
-      run: |
-        # Dirty: Get a more recent version of fontforge
-        sudo sed -i 's/focal/impish/g' /etc/apt/sources.list
-        sudo apt-get update -y
+    # Dirty OS upgrade to get newer version of FontForge
+    # - name: Tweaks to get more recent version of fontforge
+    #   run: |
+    #     # Dirty: Get a more recent version of fontforge
+    #     sudo sed -i 's/focal/impish/g' /etc/apt/sources.list
+    #     sudo apt-get update -y
+    # - name: Install external dependencies and cache them
+    #   uses: awalsh128/cache-apt-pkgs-action@v1
+    #   with:
+    #     packages: fontforge fontforge-common libfontforge4 libpython3.9 libgtk-3-0
+    #     version: 1.0
     - name: Install external dependencies and cache them
       uses: awalsh128/cache-apt-pkgs-action@v1
       with:
-        packages: fontforge fontforge-common libfontforge4 libpython3.9 libgtk-3-0
+        packages: fontforge fontforge-common libfontforge3 libgdraw6 libc6 libpython3.8 libuninameslist1 libgtk-3-0
         version: 1.0
     - name: Try to install fontforge again to get any missing non cached dependency
       run: sudo apt-get install -y fontforge
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -10,17 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Cache fontforge and extra dependencies
-      id: cache_localbin
       uses: actions/cache@v2
       with:
         path: /usr/local/bin
         key: usr-local-bin
     - name: Install latest FontForge version (using AppImage)
-      if: steps.cache_localbin.outputs.cache-hit != 'true'
       run: |
         # Get most recent version of FontForge
         # Using AppImage there is no dependecy problem, it is the latest version and it's easier to cache
-        # -------
         cd /usr/local/bin
         sudo wget -c -N https://github.com/fontforge/fontforge/releases/download/20220308/FontForge-2022-03-08-582bd41-x86_64.AppImage
         sudo chmod +x ./FontForge-2022-03-08-582bd41-x86_64.AppImage
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -10,17 +10,19 @@ jobs:
     runs-on: ubuntu-20.04 # That is the latest OS version available today
     steps:
     - name: Cache fontforge and extra dependencies
+      id: cache_localbin
       uses: actions/cache@v2
       with:
         path: /usr/local/bin
         key: usr-local-bin
     - name: Install latest FontForge version (using AppImage)
+      if: steps.cache_localbin.outputs.cache-hit != 'true'
       run: |
         # Get most recent version of FontForge
         # Using AppImage there is no dependecy problem, it is the latest version and it's easier to cache
         # -------
         cd /usr/local/bin
-        sudo wget https://github.com/fontforge/fontforge/releases/download/20220308/FontForge-2022-03-08-582bd41-x86_64.AppImage
+        sudo wget -c -N https://github.com/fontforge/fontforge/releases/download/20220308/FontForge-2022-03-08-582bd41-x86_64.AppImage
         sudo chmod +x ./FontForge-2022-03-08-582bd41-x86_64.AppImage
         sudo ln --symbolic --force /usr/local/bin/FontForge-2022-03-08-582bd41-x86_64.AppImage /usr/local/bin/fontforge
     - uses: actions/setup-java@v2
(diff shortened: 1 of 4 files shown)


sdrapha:
Next step, I want t be able to generate the debug keystore from a github secret, so if configured, the keystore from the CI can be the same as the local one in my computer, and both with same signature.
I saw somewhere a way of stringify the local binary keystore into base64 so it can be transfered by a secret. But thats for another time.

sdrapha:
> Next step, I want t be able to generate the debug keystore from a github secret, so if configured, the keystore from the CI can be the same as the local one in my computer, and both with same signature. I saw somewhere a way of stringify the local binary keystore into base64 so it can be transfered by a secret. But thats for another time.

on the works... getting close...

sdrapha:
> > Next step, I want t be able to generate the debug keystore from a github secret, so if configured, the keystore from the CI can be the same as the local one in my computer, and both with same signature. I saw somewhere a way of stringify the local binary keystore into base64 so it can be transfered by a secret. But thats for another time.
> 
> on the works... getting close...

Done!!!

Now, if you run `Make` locally in the computer, 
there will be a new file in the local `_build/` folder, called `debug.keystore.asc`

The contents of this file can be copied and pasted into a Github Secret with the name `DEBUG_KEYSTORE`

Similar to the image bellow, just the name is different!

So, now we can share the debug signature across local builds and CI runs, even across different branches, YAY!! 
Easier to update the APK without unnistalling it everytime.

image
img source, and great guide that helped me a lot: [link](https://stefma.medium.com/how-to-store-a-android-keystore-safely-on-github-actions-f0cef9413784)

Makefile
```diff
@@ -49,14 +49,18 @@ RES_FILES = $(shell find $(RES_DIR) -type f)
 # Debug signing
 
 DEBUG_KEYSTORE = _build/debug.keystore
+DEBUG_KEYSTORE_ASC = _build/debug.keystore.asc
 DEBUG_PASSWD = debug0
 
+$(DEBUG_KEYSTORE_ASC):
+	gpg -c --armor --output "$@" --pinentry-mode loopback --passphrase debug0 --yes $(DEBUG_KEYSTORE)
```
Julow, line 56: This rule isn't doing something more than just encode the keystore in a different format. I think this shouldn't be in the Makefile, the best place would be in documentation explaining how the github secret is used (perhaps in Contributing.md)

.github/workflows/make-apk.yml
```diff
@@ -7,14 +7,24 @@ on:
 
 jobs:
   Build-Apk:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-20.04 # That is the latest OS version available today
```
Julow, line 10: Is here any step that will stop working on future versions ? Otherwise, this should be fine:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/make-apk.yml
```diff
@@ -7,14 +7,24 @@ on:
...
-        sudo sed -i 's/focal/impish/g' /etc/apt/sources.list
-        sudo apt-get update -y
-        sudo apt-get install -y fontforge
+        # Get most recent version of FontForge
+        # Using AppImage there is no dependecy problem, it is the latest version and it's easier to cache
+        # -------
+        cd /usr/local/bin
+        sudo wget -c -N https://github.com/fontforge/fontforge/releases/download/20220308/FontForge-2022-03-08-582bd41-x86_64.AppImage
+        sudo chmod +x ./FontForge-2022-03-08-582bd41-x86_64.AppImage
+        sudo ln --symbolic --force /usr/local/bin/FontForge-2022-03-08-582bd41-x86_64.AppImage /usr/local/bin/fontforge
```
Julow, line 27: This can be a `mv`.

Makefile
```diff
@@ -97,12 +101,13 @@ $(R_FILE): $(RES_FILES) $(MANIFEST_FILE)
...
-SPECIAL_FONT_GLYPHS = $(wildcard srcs/special_font/*.svg)
-SPECIAL_FONT_SCRIPT = srcs/special_font/build.pe
+CURRENT_DIRECTORY := $(shell pwd)
+SPECIAL_FONT_GLYPHS = $(wildcard $(CURRENT_DIRECTORY)/srcs/special_font/*.svg)
+SPECIAL_FONT_SCRIPT = $(CURRENT_DIRECTORY)/srcs/special_font/build.pe
 
 _build/assets/special_font.ttf: $(SPECIAL_FONT_SCRIPT) $(SPECIAL_FONT_GLYPHS)
 	mkdir -p $(@D)
-	fontforge -lang=ff -script $(SPECIAL_FONT_SCRIPT) $@ $(SPECIAL_FONT_GLYPHS)
+	fontforge -lang=ff -script $(SPECIAL_FONT_SCRIPT) $(CURRENT_DIRECTORY)/$@ $(SPECIAL_FONT_GLYPHS)
```
Julow, line 110: Why this change ?

Julow (commented):
Thanks for the work! The secret is a really good idea.

.github/workflows/make-apk.yml
sdrapha: That line was there before me, I only added the comment because I did tried to run ubuntu 21.10 to get a newer version of fonforge, and then I found out that v20.04 was the latest available for actions. As of today, pining the version doesn't change anything as it is also the latest. If in the future there is some breaking change, that should be fixed anyway.

.github/workflows/make-apk.yml
sdrapha: Only thing though, with the `mv` method, by renaming the wget target, wget will redownload the target again in the next run. and that would also make the caching process more complex, we would end up with a bunch of _if_ statements to avoid redownload fontforge again.
Leaving the dowloaded file the original name, the cache will populate the file before wget gets executed, and then wget doesn't need to download anything when we have cache.

Makefile
sdrapha: I think because fontforge is an AppImage now, files passed to it as arguments are ?mounted? somehow in a weird way, and relative paths doesn't work well.

See my test runs, the logs when the build fails
https://github.com/sdrapha/Unexpected-Keyboard/runs/5688302619?check_suite_focus=true
image

So using absolute paths fixed the issue

Makefile
sdrapha: I thought that was very convenient ( a good thing ), can we at least have that command inside makefile, but instead on a separate phony rule, so it is not a requirement for the debug.apk ?

If you really say that the way to go is to remove that, and put that on documentation, so the dev working on the app needs to type the whole command manually, that is okay too. I can remove that. But honestly I was really happy with the way it is right now, the most convenient.

sdrapha: I've added documentation for the way it is implemeted currently:  3912d8b6d90ece5279e13b5a42c0389e19951241
https://github.com/sdrapha/Unexpected-Keyboard/blob/all-branches-merged-in-one/CONTRIBUTING.md#contributing=

If you decide you want the other way, that documentation also needs adjustments.

What do you think?

.github/workflows/make-apk.yml
Julow: I added it because of the patching of the sources.list, so can be removed.

.github/workflows/make-apk.yml
Julow: This if statement is already here: `if: steps.cache_localbin.outputs.cache-hit != 'true'`. If this line doesn't do anything interesting, it should be removed.

Makefile
Julow: Yes, that's what I meant, remove the rule from the Makefile and expect people to copy it from the documentation. I don't like that it's a requirements to the apk and that's it's present in the Makefile at all. A PHONY rule wouldn't be useful, it's not much more complicated to type the full command than to type the make command.

sdrapha: ok, I'll remove from the makefile then, and adjust the documentation accordingly

.github/workflows/make-apk.yml
sdrapha: I'm not fully confident that `if: steps.cache_localbin.outputs.cache-hit != 'true'` is doing all the work alone, so I rather eliminate `if: steps.cache_localbin.outputs.cache-hit != 'true'` and rely only on the `wget` behaviour of not downloading a file again if there is no need.

sdrapha: removed the redundant lines

sdrapha:
Done all changes

Makefile
Julow: I think this is a bug in the packaging of Fontforge. I don't know if AppImage is intended to package CLI programs but paths must be relative. Are there an other prebuilt package we can use ? (perhaps from Ubuntu's store?)
I've tried using Nix and caching `/nix`. It doesn't seem to be working as I expect and I fear there can be complications. It builds in less than a minute, so still a lot more than your solution.

sdrapha: I'm out of ideas, the appimage is the official distribution method that fontforge recommends. I don't know about alternative packages.
The paths are still relative, they are just expanded to absolute paths at execution. Is there a problem on that? Really asking out of curiosity, I really don't know what could go wrong, It seems ok so far though.

Julow: It seems that I will have to accept the workaround then. You can use `$(CURDIR)` instead, which is the current directory and is automatically set by make.

I'll merge this weekend.

Julow (approved):
Thanks!

PR #114, opened 2022-03-27
`Arrows` and `Box` system

sdrapha:
I liked the streamlining of the symbols with FN keys from PR #112,
but I disliked the arrows taking room on all keys on the numeric keyboard. Some of that conflicting with some of my layouts.
So I cherry picked @ArenaL5 commits from his PR, and worked on top of it. Instead of allocating the arrows directly to the keyboard on all keys, I created a Single key for Arrows, using the accent system, and once I was doing that, I also created one for Box frames. Both Arrows and Box can have variations if paired with Shift or Alt in the Box case. The Box key can also be configurable to be Lockable.

arrows and box

### DIFF
--- a/srcs/juloo.keyboard2/KeyModifier.java
+++ b/srcs/juloo.keyboard2/KeyModifier.java
@@ -204,6 +204,7 @@ class KeyModifier
       case "!": name = "¡"; break;
       case "?": name = "¿"; break;
       case "tab": name = "\\t"; break;
+      case "space": name = "nbsp"; break;
       case "€": case "£": return removed_key; // Avoid showing these twice
       default: return k;
     }
--- a/srcs/juloo.keyboard2/KeyModifier.java
+++ b/srcs/juloo.keyboard2/KeyModifier.java
@@ -145,6 +145,74 @@ class KeyModifier
           case 'o': return 'ₒ';
           default: return c;
         }
+      case KeyValue.FLAG_ACCENT_ARROWS:
+        if ((flags & KeyValue.FLAG_SHIFT) == 0)
+        {
+          switch (c)
+          {
+            case '1': return '↙';
+            case '2': return '↓';
+            case '3': return '↘';
+            case '4': return '←';
+            case '6': return '→';
+            case '7': return '↖';
+            case '8': return '↑';
+            case '9': return '↗';
+            default: return c;
+          }
+        }
+        else
+        {
+          switch (c)
+          {
+            case '1': return '⇙';
+            case '2': return '⇓';
+            case '3': return '⇘';
+            case '4': return '⇐';
+            case '6': return '⇒';
+            case '7': return '⇖';
+            case '8': return '⇑';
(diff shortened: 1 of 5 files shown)


srcs/juloo.keyboard2/KeyModifier.java
```diff
@@ -180,6 +145,88 @@ private static char handleAccentChar(char c, int flags)
           case 'o': return 'ₒ';
           default: return c;
         }
+      case KeyValue.FLAG_ACCENT_ARROWS:
+        switch ((flags & KeyValue.FLAG_SHIFT))
+        {
+        case KeyValue.FLAG_SHIFT:
```
Julow, line 151: This is `if ((flags & KeyValue.FLAG_SHIFT) != 1)`.

Julow (commented):
Not a bad idea but I have some thoughts.

Are these symbols useful ? I would like to be able to type any character on the keyboard, but not necessarily easily. If these symbols are not useful enough, maybe they can wait the new features.

Are here more such moderately-useful symbols ? We can consider adding a new pane to fit all of these.

ChiefMikeK:
## 💡
@Julow. With this and other changes it's probably a good time to open-up the **Discussions** and create an issue template for it.
- https://docs.github.com/en/discussions/managing-discussions-for-your-community/managing-categories-for-discussions-in-your-repository
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

srcs/juloo.keyboard2/KeyModifier.java
sdrapha: Removed the switch-case, also removed the extra Alt variation, now it's just shift variation using a simpler if logic

sdrapha:
> Not a bad idea but I have some thoughts.
> 
> Are these symbols useful ? I would like to be able to type any character on the keyboard, but not necessarily easily. If these symbols are not useful enough, maybe they can wait the new features.
> 
> Are here more such moderately-useful symbols ? We can consider adding a new pane to fit all of these.

I agree with the point of adding too much to the keyboard, and those symbols were not in my priority list. What triggered me to do that was the other PR taking keys space in the numeric layout, which I didn't wanted. At the end, after this pr addition, I liked the result and I think it's simple and just a small improvement that doesn't hurt. It's not a code refactoring.
Adding a new pane would not be a bad idea, but that would be a lot of change, and things to plan, and code to refactor. I like the idea though, but I'm not sure if it's really necessary. And again, I'm just a happy user of the keyboard, it was you that came up with this wonderful keyboard, thanks. You know better what you have planned.😎

Julow (approved):
That's a good idea and I'll merge but I might remove them in the future if a new feature of the keyboard is more suited for these.

Julow:
@ChiefMikeK Thanks for suggestion, I'll keep it in mind but that's not needed now.

PR #121, opened 2022-04-16
Update strings.xml

marciozomb13:
Tiny fixes and improvements.


res/values-pt/strings.xml
```diff
@@ -1,49 +1,53 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <string name="settings_activity_label">Configurações Unexpected Keyboard</string>
```
Julow, line 3: The xml files are auto formatted with `xmllint`, that's why there are often very long lines. To avoid the indentation from changing again next time, I updated your PR.

Julow (approved):
Thanks!

PR #123, opened 2022-04-18
Update store description

sdrapha:
Include the fact that the app is clean and opensource, and a link, on the description.

### DIFF
--- a/metadata/android/en-US/full_description.txt
+++ b/metadata/android/en-US/full_description.txt
@@ -17,5 +17,4 @@ Highlight of some of the features:
 Like any other virtual keyboards, it must be enabled in the system settings. Open the System Settings and go to:
 System > Languages & input > On-screen keyboard > Manage on-screen keyboards.
 
-There are no ads, no tracking, no unnecessary permissions!
-Also, it is open source! Visit the repository at https://github.com/Julow/Unexpected-Keyboard
+The project is open source. Visit the repository at https://github.com/Julow/Unexpected-Keyboard
--- a/metadata/android/en-US/full_description.txt
+++ b/metadata/android/en-US/full_description.txt
@@ -17,5 +17,5 @@ Highlight of some of the features:
 Like any other virtual keyboards, it must be enabled in the system settings. Open the System Settings and go to:
 System > Languages & input > On-screen keyboard > Manage on-screen keyboards.
 
-There are no ads, no tracking, no unnecessary permisions!
+There are no ads, no tracking, no unnecessary permissions!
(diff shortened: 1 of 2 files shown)


metadata/android/en-US/full_description.txt
```diff
@@ -16,3 +16,6 @@ Highlight of some of the features:
 
 Like any other virtual keyboards, it must be enabled in the system settings. Open the System Settings and go to:
 System > Languages & input > On-screen keyboard > Manage on-screen keyboards.
+
+There are no ads, no tracking, no unnecessary permisions!
```
sdrapha, line 20:
```suggestion
There are no ads, no tracking, no unnecessary permissions!
```

metadata/android/en-US/full_description.txt
```diff
@@ -16,3 +16,6 @@ Highlight of some of the features:
 
 Like any other virtual keyboards, it must be enabled in the system settings. Open the System Settings and go to:
 System > Languages & input > On-screen keyboard > Manage on-screen keyboards.
+
+There are no ads, no tracking, no unnecessary permissions!
```
Julow, line 20: Ads and tracking are mentioned on line 13, you might want to remove the first occurrence.
There's a lot of '!' ! I would prefer if one of them is removed.

Julow:
F-Droid already shows a link to the repository and apps are already expected to be clean of ads and tracking. Users don't expect to see this mentioned in the description.

This only applies to the play store which is updated manually for now, I plan to try [fastlane](https://docs.fastlane.tools/actions/supply/) at some point.
In the meantime, I'd prefer if this new bit of description was in a separate file, `google_extra_description`, so it is not lost until I have a better workflow and so that F-Droid is unaffected.

metadata/android/en-US/full_description.txt
sdrapha: oh, my bad, didn't notice at the end of the line. 
I removed it.

sdrapha:
> F-Droid already shows a link to the repository and apps are already expected to be clean of ads and tracking. Users don't expect to see this mentioned in the description.
> 
> This only applies to the play store which is updated manually for now, I plan to try [fastlane](https://docs.fastlane.tools/actions/supply/) at some point. In the meantime, I'd prefer if this new bit of description was in a separate file, `google_extra_description`, so it is not lost until I have a better workflow and so that F-Droid is unaffected.

Ah ok. I got it. So the description.txt is for fdroid only, and not play store. So yeah, these changes should have a different aproach then 👍

Julow:
Sorry, I'm closing this PR. I'm still interested in improving the store description but it's quite low in my TODO list for this project.

PR #125, opened 2022-04-19
Added neo 2 layout

matthiakl:
Added the [Neo2 Layout](https://neo-layout.org/). I only included layers 1 to 3, otherwise the keyboard would be quite full. The layout contains all ASCII characters and dead keys, therefore I added it to the programming layouts. I left out most of the secondary keys on the number line, they are accessible on the Fn layer. The dead keys from the corners are shifted towards different keys, but still close to their original positions.
(The Contributing file still only lists qwerty as programming layout, should I update it?)

Screenshot_20220419-090841_1 
The full layout for reference: Bildschirmfoto vom 2022-04-19 19-22-36


Julow (commented):
Awesome!

It's missing some accents: `ű` and `ė` (related to https://github.com/Julow/Unexpected-Keyboard/pull/127 and https://github.com/Julow/Unexpected-Keyboard/issues/128) and some others I can't type. I would gladly add these diacritics and support for the corresponding languages with some help.

There's many characters close to the edges of the screen, which can be hard to type for some users. I don't mandate this but it could be nice to move some symbols, at the cost of deviating even more from the original layout.

> I only included layers 1 to 3

Layer 4 is more or less already supported by the numeric pane (I'd accept more symbols on it) and layers 5-6 would be better as an additional pane.

>  (The Contributing file still only lists qwerty as programming layout, should I update it?)

Yes, thanks.

matthiakl:
> It's missing some accents: ű and ė (related to https://github.com/Julow/Unexpected-Keyboard/pull/127 and https://github.com/Julow/Unexpected-Keyboard/issues/128) and some others I can't type.

Indeed, but this is a German layout and we only need dead key accents for non-German words, so it is not a problem for this one.

>There's many characters close to the edges of the screen, which can be hard to type for some users.

I moved most of the symbols from the edge. All dead accents are now in the second row (one place to give away :wink:):
Bildschirmfoto vom 2022-04-23 21-55-42


> Layer 4 is more or less already supported by the numeric pane (I'd accept more symbols on it) and layers 5-6 would be better as an additional pane.

I agree. Maybe I will create a symbolic layout for the numeric pane later.

res/values/arrays.xml
```diff
@@ -27,16 +28,19 @@
     <item>QWERTZ</item>
     <item>Bulgarian (Phonetic Traditional)</item>
     <item>Dvorak</item>
+    <item>Neo 2</item>
   </string-array>
   <string-array name="pref_programming_layout_values">
     <item>none</item>
     <item>qwerty</item>
     <item>dvorak</item>
+    <item>neo2</item>
```
Julow, line 37: Is it actually a programming layout ?

Programming layouts are made for people that need a different layout for programming and for typing in their usual language. (makes most sense for non-Latin languages users) Is it plausible that someone would want to use `neo2` when programming but an other layout the rest of the time ?

Julow (approved):
I have a second thought about the programming layout thing but other looks good :) Happy to merge once you think it's ready.

res/values/arrays.xml
matthiakl: Hm, not really. I don't think non German people will benefit from it. Should I remove it from the list then?

Julow: Yes, let's remove it from that list.

matthiakl: Done. I also added the new accents (accent_ogonek definitely is not present on the keyboard)

Julow (approved):
Looks very good! Thanks.

PR #126, opened 2022-04-19
German translation update

polyctena:
Translated the newly added strings (lockable modifiers, programming keyboard layout)


Julow (approved):
Thanks!

Issue #129, opened 2022-04-24
Add Colemak layout


dircsem:
I've added this issue to suggest to add this layout to the app.
I'm trying to do that.

dircsem:
I'm having this issue on building the app
not sure if this is related to the version of the 
Import: Wrong number of arguments
Called from...
/Unexpected-Keyboard/srcs/special_font/build.pe: line 10
make: *** [Makefile:105: _build/assets/special_font.ttf] Error 1

Is this related to some dependency?

Julow:
Thanks for contributing!

This is due to the Fontforge dependency, you need at least the version `20201107`, which isn't available in many distros.

If you are ok with using Nix, you should have everything right with `nix-shell ./shell.nix`. As a last recourse, you can let the CI build the app for you (push a branch to your fork and go to the "Actions" tab on Github).
This is not ideal! I plan on removing the dependency on Fontforge soon.

dircsem:
I  don't know how to use this nix-shell will give a try. 
It's good to have this instructions at the contributing doc, I'll add it there

dircsem:
![photo_2022-04-24_15-24-26](https://user-images.githubusercontent.com/2041058/164990848-401847a4-2629-48e1-a735-4ba1b29baf90.jpg)
Any tips on how to position the non character keys? Is there a problem to split esc and shift? Is is recommended to have symbol keys apart from the alphabet keys, or the idea is to have those always placed with alphabet keys?

Julow:
There's no constraint on the placement of shift and backspace, the dvorak layout have them on the top row too.

It's also entirely fine to have keys for non-alphabet characters like Esc, `>`, `<`, Tab, etc..., the other layouts are like this mainly to look more familiar. However, you might end up with very tiny keys: Shift and Backspace are on the third row for that reason.

You can also consider modifying the bottom row by adding `bottom_row="false"` to the `<keyboard>` markup and taking the bottom row from `bottom_row.xml`. The neo2 layout does that: e52e537

Btw, Fontforge is no longer a build dependency (39952f8).

dircsem:
![photo_2022-04-24_20-15-39](https://user-images.githubusercontent.com/2041058/165000745-3386b5cf-a5f4-474d-8f08-322642851f28.jpg)
I think that the keys are ok for now, I just need to adapt the symbols.

dircsem:
[...]

PR #135, opened 2022-05-02
add colemak layout #129

dircsem:


res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard botton_row="false">
```
Julow, line 2: There's a typo so this has no effect.
```suggestion
<keyboard>
```

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+  <row>
+    <key key0="shift" width="1.5" />
+    <key key0="z" />
+    <key key0="x" key1="&gt;" key2="&lt;"/>
+    <key key0="c" key1="{" key2="}" key3="accent_cedille"/>
+    <key key0="v" key1="[" key2="]"/>
+    <key key0="b" key1="(" key2=")"/>
+    <key key0="k" key1=";" key2=":"/>
+    <key key0="m" key1="|" key2="\?" />
+    <key key0="backspace" shift="0.25" key1="delete" width="1.25"/>
```
Julow, line 38: Why this empty space between `m` and backspace ?

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+    <key key0="a" key2="tab"/>
+    <key key0="r" key1="accent_grave" key2="accent_tilde"/>
+    <key key0="s" key1="`" key2="~" />
+    <key key0="t" key1="accent_double_aigu" key2="accent_ring"/>
+    <key key0="d" key1="&quot;" key2="'" />
+    <key key0="h"  />
+    <key key0="n" key1="accent_ogonek"/>
+    <key key0="e" key1="accent_circonflexe" key2="accent_aigu" />
+    <key key0="i" key1="\\" key2="/"/>
+    <key key0="o" />
```
Julow, line 26: The macron accent is missing, `o` is a good place for it:
```suggestion
    <key key0="o" key1="accent_macron" />
```

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+    <key key0="l" key2="7" key3="&amp;"/>
+    <key key0="u" key2="8" key3="*"/>
+    <key key0="y" key2="9" />
+    <key key0="," key1="0" key3="."/>
+
+  </row>
+  <row>
+    <key key0="a" key2="tab"/>
+    <key key0="r" key1="accent_grave" key2="accent_tilde"/>
+    <key key0="s" key1="`" key2="~" />
```
Julow, line 19: Missing key:
```suggestion
    <key key0="s" key1="`" key2="~" key3="ß" />
```
Don't worry, this key is only shown to German users.

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+    <key key0="t" key1="accent_double_aigu" key2="accent_ring"/>
+    <key key0="d" key1="&quot;" key2="'" />
+    <key key0="h"  />
+    <key key0="n" key1="accent_ogonek"/>
```
Julow, line 23: An other missing accent:
```suggestion
    <key key0="n" key1="accent_ogonek" key2="accent_dot_above"/>
```

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+  <row>
+    <key key0="q" key2="1" key4="esc"/>
+    <key key0="w" key1="~" key2="2" key3="\@" key4="!"/>
+    <key key0="f" key1="£" key2="3" key3="\#" key4="€"/>
+    <key key0="p" key2="4" key3="$" key4="f11_placeholder"/>
+    <key key0="g" key2="5" key3="%" key4="f12_placeholder"/>
+    <key key0="j" key1="accent_caron" key2="6" key3="^" key4="accent_trema"/>
+    <key key0="l" key2="7" key3="&amp;"/>
+    <key key0="u" key2="8" key3="*"/>
+    <key key0="y" key2="9" />
```
Julow, line 12: `-` and `_` are missing ! These are pretty important keys. I suggest to put them at the bottom of the `y` key:
```suggestion
    <key key0="y" key2="9" key3="-" key4="_" />
```

res/xml/colemak.xml
```diff
@@ -0,0 +1,43 @@
...
+  <row>
+    <key key0="a" key2="tab"/>
+    <key key0="r" key1="accent_grave" key2="accent_tilde"/>
+    <key key0="s" key1="`" key2="~" />
+    <key key0="t" key1="accent_double_aigu" key2="accent_ring"/>
+    <key key0="d" key1="&quot;" key2="'" />
+    <key key0="h"  />
+    <key key0="n" key1="accent_ogonek"/>
+    <key key0="e" key1="accent_circonflexe" key2="accent_aigu" />
+    <key key0="i" key1="\\" key2="/"/>
```
Julow, line 25: `=` and `+` are missing, I suggest to move `\\` and `/` to make room:
```suggestion
    <key key0="i" key1="=" key2="+" key3="\\" key4="/"/>
```

Julow (commented):
Thanks!

What's version of Colemak is it ? The `,` on the bottom right feels unusual.
There's a few missing keys, I suggested a place to insert them.

res/xml/colemak.xml
dircsem: This is because if they are together you can press delete instead of m

dircsem:
I've changed the layout on the last few days and I`ve removed on key from the top, also moved numbers to the second line. I`ll merge your requests and see how it'll be.
I've seen the missing =, the other keys I've need to use ;)
photo_2022-05-07_17-12-40
I'll post a picture after merging.

res/xml/colemak.xml
Julow: Makes sense but it's the only layout that have this. I wonder if this should be an option available to everyone.
Anyway, I don't mind this.

dircsem: Yes it could be an option, I haven't seen anything related to the app code I`ll try to find out where are the interfaces and try to place an option for that.

Julow: Let's not worry about the option yet as I'm not yet sure it is useful. Let's keep the gap in the layout and see how it is on the long run. Maybe other people will share their opinions.

dircsem: Ok I`ve take a look on how to put this option but it's another issue.
I've based this layout, the character keys, in the picture bellow:
https://lh5.ggpht.com/Y6MxBScWfyORVaMMkgMAO3Kl63ULQXTisNYIjCZdeU2c9hIy_j4bvXZyWoPms_I02_NA=h500
But I didn't followed the layout for symbols and had tried to put it when I thought that it was better.
I think that this ok, I'll keep using it and see if there are any keys that could be changed, please do a review to see if all necessary keys are here.
# this is another issue
Can the accents be merge in to a one key with their characters? Like merging "accent_tilde" with ~ character, if the user press it and press a compatible key like a it became ã, if the user press it and space it became ~.

Julow: > Can the accents be merge in to a one key with their characters?

No, this is not possible.

Julow (approved):
Merging when you think it's ready.

dircsem:
I think that you can merge it already, I'll test for another week and if I get any other changes I can push next weekend.

res/xml/colemak.xml
dircsem: This in fact is exact the behavior, but now we have to add like accent_tilde and ~ keys, but we could add it as one key, the "~" acting as accent_tilde instead of adding it twice.
In my opinion this would make easy to visualize at keyboard and would bring the same behavior as we have at computer keyboards.

Julow:
Merging :) Thanks !

Sorry for the delay, I planned on not releasing on the month of May.

PR #151, opened 2022-05-19
Turkish layout & translation

erqan:


res/xml/method.xml
```diff
@@ -14,4 +14,5 @@
   <subtype android:label="%s" android:languageTag="pt" android:imeSubtypeLocale="pt_BR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_pt,extra_keys=accent_aigu|accent_cedille|accent_circonflexe|accent_grave|accent_tilde|€"/>
   <subtype android:label="%s" android:languageTag="ru" android:imeSubtypeLocale="ru_RU" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=ru_jcuken"/>
   <subtype android:label="%s" android:languageTag="sv" android:imeSubtypeLocale="sv_SE" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_sv_se,extra_keys=accent_aigu|accent_trema|accent_ring|€"/>
+  <subtype android:label="%s" android:languageTag="tr" android:imeSubtypeLocale="tr_TR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_tr,extra_keys=ç|ğ|ı|ö|ş|ü|€"/>
```
Julow, line 17: Currently, none of the `loc ` keys you added will be shown in any other layouts. You can remove the `loc ` prefix in the new layout and change this line to:

```suggestion
  <subtype android:label="%s" android:languageTag="tr" android:imeSubtypeLocale="tr_TR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_tr,extra_keys=accent_cedille|accent_caron|accent_trema|ı|€"/>
```
This will add the required dead-keys on the other layouts. The `ı` won't be shown yet but I'm already working on improving that.

Julow (commented):
Thanks!

res/xml/method.xml
Julow: Since 2e81cb5cf, the `ı` will be added on other layouts too.

Julow (approved):
Awesome, thanks!

My suggestion above was a bit wrong, I added a commit to tweak the extra keys.

randomizednum:
There is no way to type the uppercase dotted i. I'm not sure if it's possible to switch letters depending on letter case with this layout system, but it's a pretty big annoyance when writing in Turkish.

Julow:
@randomizednum I added İ in fc8bb3a. For technical reason, i and ı will be swapped when shifting. You can find a debug version of the app here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/3180483459

PR #155, opened 2022-06-02
Bone keyboard layout

### LINKED ISSUE #116: Bone-keyboard-layout
Would you accept a Layout like this, if I would make it:
https://www.neo-layout.org/Layouts/bone/
?
There is unfortunately no English website, just scroll to the bottom and click through the tabs to see what it features.
I would probably put Layer 3 "Ebene 3" to 6 in the corners.

Quantenzitrone:
closes #116 
- [x] math additions to handleFn
- [x] normal bone layout pane
- [x] math pane
→ almost finished, I just have to decide if I want some greek letters multiple times (like in the desktop bone layout) or if I want to add some more letters and which letters.
Currently, I consider some of these: ϗϏ  Ϙϙ  Ϛϛ  Ϝϝ  Ϟϟ  Ϡϡ  Ϸϸ  Ͱͱ, but not all will fit.
- [x] slash diacritic for characters like this: øħ
- [x] switch from normal pane and greek&math
- [x] new special font characters for ◌⃗ and accent_slash (and other?)
- [x] finalize the layouts (especially the math layouts still need tweaking)


Quantenzitrone:
◌⃗ is by the way not a dead key but just a something that goes on any character:   ⃗ a⃗

srcs/juloo.keyboard2/KeyModifier.java
```diff
@@ -297,6 +300,32 @@ private static KeyValue handleFn(KeyValue k, int flags)
...
+      case "Θ": name = "ϴ"; break;
+      case "ε": name = "ϵ"; break;
+      case "β": name = "ϐ"; break;
+      case "ρ": name = "ϱ"; break;
+      case "σ": name = "ς"; break;
+      case "γ": name = "ɣ"; break;
+      case "φ": name = "ϕ"; break;
+      case "υ": name = "ϒ"; break;
+      case "κ": name = "ϰ"; break;
+      // alternating math characters
```
Julow, line 315: While typing math, the Shift key might be used more often, to type uppercase greek letters/symbols. I think these symbols should be under `Shift` instead of `Fn` to be more discoverable.
The `Fn` would be a lot less useful on the greek pane, which is a good idea.

What do you think ?

srcs/juloo.keyboard2/KeyModifier.java
```diff
@@ -297,6 +300,32 @@ private static KeyValue handleFn(KeyValue k, int flags)
       case "c": name = "¢"; break;
       case "p": name = "₱"; break;
       case "€": case "£": return removed_key; // Avoid showing these twice
+      // alternating greek letters
+      case "π": name = "ϖ"; break;
+      case "θ": name = "ϑ"; break;
+      case "Θ": name = "ϴ"; break;
```
Julow, line 306: This is `Shift+Fn+θ`, quite complicated, these symbols are used as often as the greek letters, I think.
What about adding `ϑ` to the keyboard ? `ϴ` would be `Shift+ϑ`.

Same for the others `ϐ`, `ϱ`, etc...

Julow (commented):
Thanks! Let me take care of the slash and the vector dead keys :) 

> ◌⃗ is by the way not a dead key but just Unicode combining character (U+20D7) that goes on any character like this: ⃗ a⃗

This is generally done using dead-keys on this keyboard. (dead keys in this app are a kind of interactive combining character)

>  Currently, I consider some of these: ϗϏ Ϙϙ Ϛϛ Ϝϝ Ϟϟ Ϡϡ Ϸϸ Ͱͱ, but not all will fit.

Half of these are the uppercase version of the other half. There's still a lot of room on the keyboard.

Julow:
About uppercase characters, most of them should be handled by Java's `toUpperCase` function. If some are missing, you can add a `switch case` in `handleShift` (`KeyModifier.java` line 45).

Quantenzitrone:
> This is generally done using dead-keys on this keyboard. (dead keys in this app are a kind of interactive combining character)

As far as I know, all the other dead keys are diacritics, which result in a different Unicode character when typed. 
This ◌⃗ is containing 2 characters, so I could combine any character with it. ,⃗ +⃗ 0⃗ ϰ⃗ with the arrow always at the same position. 
That's why some character look weird with it, for example M⃗ °⃗ —⃗

Quantenzitrone:
> Half of these are the uppercase version of the other half. There's still a lot of room on the keyboard.

Yeah, that's intentional, but these are 8 different letters and I only have 3 keys left on the keyboard which have no character yet, so max. 6 characters fit (with fn as a character modifier).
For example: I would use Ϛϛ (stigma) as a variation(fn) of Ϝϝ (diagramma) because acc. to [Wikipedia](https://en.wikipedia.org/wiki/Digamma) they are related.
The same with Ϟϟ (koppa) and Ϙϙ (archaic koppa).

srcs/juloo.keyboard2/KeyModifier.java
Quantenzitrone: They are not uppercase but variations of the lowercase character. For example \phi is rendered as ϕ in LaTeX and \varphi as φ 
maybe that should be the other way around, but on the original Bone-Layout φ is the normal lowercase phi on Shift+M3+f and ϕ is on Shift+M3+q.
the same with:
\pi as π and \varpi as ϖ
\rho as ρ and \varrho as ϱ
etc.

On the desktop Bone-Layout these characters are on different keys (see Ebene 5 [here](https://www.neo-layout.org/Layouts/bone/)) but since we have fn as a general character alternating key here, I thought this would be the most obvious way to do it.

srcs/juloo.keyboard2/KeyModifier.java
Quantenzitrone: Like in my comment in the above conversation, ϴ is a variation of Θ. 
LaTeX and the desktop Bone-Layout don't have this variation, but I thought it would make sense to include it anyway.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
+<?xml	version="1.0"	encoding="utf-8"?>
+<keyboard>
+	<row>
+		<key   width="1"   key0="j"							key2="°"						key4="esc"/>
```
Julow, line 4: All the `width="1"` can be removed.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1"   key0="d"							key2="§"						key4="_"/>
+		<key   width="1"   key0="u"							key2="ℓ"						key4="["/>
+		<key   width="1"   key0="a"							key2="»"						key4="]"/>
+		<key   width="1"   key0="x"							key2="«"						key4="^"/>
+		<key   width="1"   key0="p"			key1="$"						key3="!"		key4="7"/>
+		<key   width="1"   key0="h"			key1="€"		key2="⟨"		key3="&lt;"		key4="8"/>
+		<key   width="1"   key0="l"			key1="£"		key2="⟩"		key3="&rt;"		key4="9"/>
```
Julow, line 11:
```suggestion
		<key   width="1"   key0="l"			key1="£"		key2="⟩"		key3="&gt;"		key4="9"/>
```

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1"   key0="b"					key2="accent_macron"	key3="?"		key4="4"/>
+		<key   width="1"   key0="n"					key2="accent_tilde"		key3="("		key4="5"/>
+		<key   width="1"   key0="r"					key2="accent_trema"		key3=")"		key4="6"/>
+		<key   width="1"   key0="s"					key2="accent_slash"		key3="-"				/>
+		<key   width="1"   key0="g"			key1="@"						key3=":"				/>
+	</row>
+	<row>
+		<key   width="1.5" key0="shift"						key2="greekmath"				key4="#"/>
+		<key   width="1"   key0="f"															key4="|"/>
+		<key   width="1"   key0="v"											key3="⃗"			key4="~"/>
```
Julow, line 30:
```suggestion
		<key   width="1"   key0="v"											key3="accent_arrow_right"			key4="~"/>
```

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1.5" key0="backspace"					key2="delete"	key3=""					/>
+	</row>
+	<row height="0.95">
+		<key width="1.8"   key0="ctrl"						key2="meta" 					key4="switch_numeric"/>
+		<key width="1.2"   key0="fn"	key1="alt"			key2="change_method" key3="switch_emoji" key4="config"/>
+		<key width="4.0"   key0="space"	key1="superscript"	key2="0"		key3="subscript"		/>
+		<key width="1.2"   key1="up"						key2="right" 	key3="left" 	key4="down" edgekeys="true"/>
+		<key width="1.8"   key0="enter"										key3="action"/>
+	</row>
+</keybo6ard>
```
Julow, line 45:
```suggestion
</keyboard>
```

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1"   key0="c"							key2="tab"						key4="\"/>
+		<key   width="1"   key0="t"			key1="accent_circonflexe"	key2="accent_caron"	key4="/"/>
+		<key   width="1"   key0="i"			key1="accent_aigu"			key2="accent_grave"	key4="{"/>
+		<key   width="1"   key0="e"			key1="accent_cedille"	key2="accent_ogonek"	key4="}"/>
+		<key   width="1"   key0="o"			key1="accent_ring"		key2="accent_dot_above"	key4="*"/>
+		<key   width="1"   key0="b"					key2="accent_macron"	key3="?"		key4="4"/>
+		<key   width="1"   key0="n"					key2="accent_tilde"		key3="("		key4="5"/>
+		<key   width="1"   key0="r"					key2="accent_trema"		key3=")"		key4="6"/>
+		<key   width="1"   key0="s"					key2="accent_slash"		key3="-"				/>
+		<key   width="1"   key0="g"			key1="@"						key3=":"				/>
```
Julow, line 25: Some characters need to be escaped due to Android's XML having all kinds of new syntaxes (references to resources, colors). Same for `?` and `#`.
```suggestion
		<key   width="1"   key0="g"			key1="\@"						key3=":"				/>
```

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
+<?xml	version="1.0"	encoding="utf-8"?>
+<keyboard>
+	<row>
+		<key   width="1"   key0="j"							key2="°"						key4="esc"/>
+		<key   width="1"   key0="d"							key2="§"						key4="_"/>
+		<key   width="1"   key0="u"							key2="ℓ"						key4="["/>
+		<key   width="1"   key0="a"							key2="»"						key4="]"/>
+		<key   width="1"   key0="x"							key2="«"						key4="^"/>
```
Julow, line 8: `«` and `»` are `Fn+<` and `Fn+>`. Are there really needed on the main pane ?

res/xml/bone.xml
Julow, line 11: Are `⟨` and `⟩` common ? The `Fn` layer starts to fill up but I'd suggest `Fn+;` and `Fn+:`.

res/xml/bone_greekmath.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1.5" key0="backspace"					key2="delete"							/>
+	</row>
+	<row height="0.95">
+		<key width="1.8"   key0="ctrl"						key2="meta" 					key4="switch_numeric"/>
+		<key width="1.2"   key0="fn"	key1="alt"			key2="change_method" key3="switch_emoji" key4="config"/>
+		<key width="4.0"   key0="space"	key1="superscript"	key2="0"		key3="subscript"		/>
+		<key width="1.2"   key1="up"						key2="right" 	key3="left" 	key4="down" edgekeys="true"/>
+		<key width="1.8"   key0="enter"										key3="action"/>
+	</row>
+</keybo6ard>
```
Julow, line 45:
```suggestion
</keyboard>
```

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key   width="1"   key0="p"			key1="$"						key3="!"		key4="7"/>
+		<key   width="1"   key0="h"			key1="€"		key2="⟨"		key3="&lt;"		key4="8"/>
+		<key   width="1"   key0="l"			key1="£"		key2="⟩"		key3="&rt;"		key4="9"/>
+		<key   width="1"   key0="m"			key1="¥"						key3="="		key4="f11_placeholder"/>
+		<key   width="1"   key0="w"			key1="¢"						key3="&amp;"	key4="f12_placeholder"/>
+	</row>
+	<row>
+		<key   width="1"   key0="c"							key2="tab"						key4="\"/>
+		<key   width="1"   key0="t"			key1="accent_circonflexe"	key2="accent_caron"	key4="/"/>
```
Julow, line 17: The dead keys are visually heavy and more and more will be added to the keyboard.
They will be added automatically to the layout depending on the user's installed languages so you can remove them all.
If you wish to dedicate a place for some of them but to keep them hidden by default, you can add the `loc ` prefix.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
+<?xml	version="1.0"	encoding="utf-8"?>
+<keyboard>
```
Julow, line 2: Otherwise you'll have two bottom rows:
```suggestion
<keyboard bottom_row="false">
```

res/xml/bone.xml
Julow, line 4: `°` is already `Fn+*`.

Julow (commented):
Modifiers were represented as flags and we were running out of flags. Instead of increasing the size of the field, I decided to rewrite the way modifiers were represented internally. `KeyModifier` changed substantially, but I think you should be able to do `git merge master` without conflicts.

I've added the `accent_slash` (e4e5462) and `accent_arrow_right` (02124dd) modifiers.

> As far as I know, all the other dead keys are diacritics, which result in a different Unicode character when typed.

It's the first of a kind indeed but a new mechanism isn't necessary, modifiers work until new combining characters are added and new problems are found.

The layout is pretty full. Do you really need all these characters on the main pane ?

I've pushed a commit to your branch to add the layout to the Layout option. Once the XML file is fixed, you should be able to download a debug APK from the CI.

res/xml/bone.xml
Quantenzitrone: This is kinda intentional because I think it is part of the Bone Keyboard Philosophy to be able to type a lot of characters. But I guess it wouldn't hurt to make them hide according to the user's installed languages, because everyone can still enable to show all in the settings.

Quantenzitrone: ill do this later

res/xml/bone.xml
Quantenzitrone: ill have a look into this later

res/xml/bone.xml
Quantenzitrone: ill do this later

Julow:
Well played for the merge :)
Sorry, I won't have time to continue reviewing until next week.

Quantenzitrone:
It works 😁
tho I have spotted some oddities already:
- the "show all accents"-option is gone
- f12-placeholder is also f11 or at least it has the text f11, but I don't know how to test this

Quantenzitrone:
Wait what show every accent setting is something that changed in 1.16.0
why?
accents are wonderful

Quantenzitrone:
"This option cannot be implemented easily now that the set of "accents"
(localized keys) isn't defined." ah

Julow:
> Wait what show every accent setting is something that changed in 1.16.0
why?
accents are wonderful

This option was in the way of other improvements (see https://github.com/Julow/Unexpected-Keyboard/issues/164). Are you missing some dead-keys or characters ? If yes, what are your installed languages ?

Quantenzitrone:
I have English and German installed and as expected I only have the diacritic for Umlaut (`accent_trema`). So I don't want have to install 6 or more (idk) languages to be able to type all diacritic-characters.

Would bringing back the "show all accents" option be possible like this?:
```java
  private void refreshAccentsOption(InputMethodManager imm, InputMethodSubtype subtype)
  {
    HashSet<KeyValue> extra_keys = new HashSet<KeyValue>();
    List<InputMethodSubtype> enabled_subtypes = getEnabledSubtypes(imm);
    switch (_config.accents)
    {
      // '3' was "all accents", now unused
      case 1:
        extra_keys_of_subtype(extra_keys, subtype);
        for (InputMethodSubtype s : enabled_subtypes)
          extra_keys_of_subtype(extra_keys, s);
        break;
      case 2:
        extra_keys_of_subtype(extra_keys, subtype);
        break;
      case 3: 
        extra_keys.add(KeyValue.getKeyByName("accent_aigu"));
        extra_keys.add(KeyValue.getKeyByName("accent_caron"));
        extra_keys.add(KeyValue.getKeyByName("accent_cedille"));
        extra_keys.add(KeyValue.getKeyByName("accent_circonflexe"));
        extra_keys.add(KeyValue.getKeyByName("accent_grave"));
        extra_keys.add(KeyValue.getKeyByName("accent_macron"));
        extra_keys.add(KeyValue.getKeyByName("accent_ring"));
        extra_keys.add(KeyValue.getKeyByName("accent_tilde"));
        extra_keys.add(KeyValue.getKeyByName("accent_trema"));
        extra_keys.add(KeyValue.getKeyByName("accent_ogonek"));
        extra_keys.add(KeyValue.getKeyByName("accent_dot_above"));
        extra_keys.add(KeyValue.getKeyByName("accent_double_aigu"));
        extra_keys.add(KeyValue.getKeyByName("accent_slash"));
        extra_keys.add(KeyValue.getKeyByName("accent_arrow_right"));
      break;
      case 4: break;
      default: throw new IllegalArgumentException();
    }
    _config.extra_keys = extra_keys;
    if (enabled_subtypes.size() > 1)
      _config.shouldOfferSwitchingToNextInputMethod = true;
  }
```
I made this in my local repo, but I wasn't able to build it because the android-sdk packages for my distro are pretty out-of-date. :(
Should I push it to test it here?

Quantenzitrone:
oh it works, I managed to do it with nix

but the dead keys are not placed in the right places

Screenshot_KISS_launcher_20220620-103937

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key key0="j"				key2="°"				key4="esc"/>
+		<key key0="d"				key2="§"				key4="_"/>
+		<key key0="u"				key2="ℓ"				key4="["/>
+		<key key0="a"				key2="↻"				key4="]"/>
+		<key key0="x"				key2="↺"				key4="^"/>
+		<key key0="p"	key1="$"				key3="!"	key4="7"/>
+		<key key0="h"	key1="€"				key3="&lt;"	key4="8"/>
+		<key key0="l"	key1="£"				key3="&gt;"	key4="9"/>
+		<key key0="m"	key1="¥"				key3="="	key4="f11_placeholder"/>
+		<key key0="w"	key1="¢"				key3="&amp;"	key4="f12_placeholder"/>
```
Julow, line 13: The currency symbols are in the `Fn` layer already, no need to put them here.

Julow (commented):
Sorry for the slow reply.

Your idea for the "show all accents" option would work to some extent. I do not wish to maintain such a list, I'd rather have descriptions for more languages. An other problem is that the list might get bigger and bigger.

A better feel would be an option to add any keys to the keyboard (either predefined dead-keys or any custom characters). Even thought, I would prefer to improve the language descriptions first.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key key0="y"							key3="%"		key4="+"/>
+		<key key0="z"	key1="&quot;" key2="–"	key3=","		key4="1"/>
+		<key key0="q"	key1="&apos;" key2="•"	key3="."		key4="2"/>
+		<key key0="k"							key3=";"		key4="3"/>
+		<key width="1.5" key0="backspace"	key3="delete"/>
+	</row>
+	<row height="0.95">
+		<key width="1.8" key0="ctrl"						key2="meta" 								key4="switch_numeric"/>
+		<key width="1.2" key0="fn"		key1="alt"			key2="change_method"	key3="switch_emoji" key4="config"/>
+		<key width="4.0" key0="space"	key1="superscript"	key2="0"				key3="subscript"/>
```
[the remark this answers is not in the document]
Quantenzitrone: Yeah, but I don't like how ord is used in the num pane to make superscript letters.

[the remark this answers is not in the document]
Julow: Not every letters can be super/sub scripts (`in` for superscript, `eaxo` for subscript), so in the current state, that's pointless.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
...
+		<key key0="n"									key2="loc accent_tilde"		key3="("	key4="5"/>
+		<key key0="r"									key2="loc accent_trema"		key3=")"	key4="6"/>
+		<key key0="s"									key2="loc accent_slash"		key3="-"/>
+		<key key0="g"	key1="\@"													key3=":"/>
+	</row>
+	<row>
+		<key width="1.5" key0="shift"		key2="greekmath"	key4="\#"/>
+		<key key0="f"											key4="|"/>
+		<key key0="v"							key3="accent_arrow_right"	key4="~"/>
+		<key key0="ß"				key2="ſ"					key4="`"/>
```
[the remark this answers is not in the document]
Quantenzitrone: I have never used [ſ](https://en.wikipedia.org/wiki/Long_s) but it is on M3 + ß in Bone, that's why I added it there.

Quantenzitrone:
Would you be okay with 2 versions of Bone instead? One, similarly to the current implementation of Neo2, with a number row and one heavily customized to optimally fit into the theme of unexpected keyboard? Because I kinda realized that trying to do both at once doesn't work, but I kinda still want the actual Bone Layout to be implemented.
I kinda feel like the customized version wouldn't really attract many users at first, because it wouldn't be actual Bone anymore
But I also still think the customized version should exist because, if other Neo enthusiasts have ideas to improve it, it can be changed, and it could be a cool community project to bring Neo Layouts to mobile in an unjanky way.

res/xml/bone.xml
```diff
@@ -0,0 +1,45 @@
+<?xml	version="1.0"	encoding="utf-8"?>
+<keyboard bottom_row="false">
+	<row>
+		<key key0="j"				key2="°"				key4="esc"/>
+		<key key0="d"				key2="§"				key4="_"/>
+		<key key0="u"				key2="ℓ"				key4="["/>
+		<key key0="a"				key2="↻"				key4="]"/>
+		<key key0="x"				key2="↺"				key4="^"/>
```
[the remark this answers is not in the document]
Quantenzitrone: removed

res/xml/bone.xml
Quantenzitrone: removed

res/xml/bone.xml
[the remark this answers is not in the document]
Quantenzitrone: i just removed it, if someone needs it back it can they can make a new issue

Quantenzitrone:
Screenshot_Scarlet_Notes_FD_20220708-030445

Julow:
I've merged the greekmath layout, the ẞ and the greek/math Fn combinations. Thanks a lot!

I've added the key for switching to the greekmath pane and placed it on the numeric pane and on `Fn+switch_numeric` for the other layouts. I did small changes to it:
- Bigger buttons to move back to the other panes
- Removed modifiers (Alt, Ctrl, Mod) that have no effect and some other keys

I've pushed a merge commit to your branch to remove things that are already merged. Of course, feel free to make more modification to the greekmath layout.

Julow:
> Would you be okay with 2 versions of Bone instead?

Please add the layout that you would use but also one that works well with the app and the touch screen. You can't replicate exactly the same layout due to the constraints (no layers, swipe movements, hard-coded combinations).
I'm not against another version in the future but only if it has users.

Quantenzitrone:
So you made greekmath as a general layout? Other Neo Variants like Neo2 or NeoQuertz would need different math panes to reflect the placements of the keys in the main layout. α at the same position as a for example.
I'd rather have it implemented similarly to the keyboard layout for programming, where you can have different variants.
Right now, every layout has the math layout for bone…

Julow:
> So you made greekmath as a general layout?

Yes. I think it's better that way, so there's only one layout to maintain and everyone can enjoy it. If it is really needed, different variants can be added later but I doubt it will be justified.
The new greekmath will certainly evolve a lot as other users give feedback and as the app evolves, this is not possible if there's many variants.

Quantenzitrone:
i created a new branch where i reimplemented the math pane like the programming layout and created a querty oriented math pane version

Quantenzitrone:
please check it out when you have time and give feedback if this implementation would be ok for you too?

Julow:
I had a look at your branch (`extra-math-pane`) and it's very interesting, thanks, but I wouldn't accept a second math pane now.

The current math pane has 0 user and it isn't satisfying yet: too crowded, some symbols are missing. As I said above, I'm not against adding an option in general but only if it is justified. If it is important to you that the math pane is exactly the one from Bone, let's add the option if it ever change too much.

Feel free to tweak the greekmath pane in this PR.

Julow:
Ping me when you think the Bone layout is ready to be merged.

Quantenzitrone:
i wanted to finish it in the train today but `nix-shell shell.nix` needed to download something and internet in german trains is apparantly too slow to download 200mb before the connection times out💀

Julow:
Hi! I would like to release soon :) What's missing to the Bone layout ?
If you have some time, I'd like to have your opinion on the greek and math symbols too.

Quantenzitrone:
Yeah, I kinda forgot about it...

I just pushed some minor fixes for the bone layout I already had locally.

Since i have been using this layout for a while now and I'm pretty happy with it, it is ready to be merged.

Quantenzitrone:
About the math layout, it is alright. I haven't been using it as much as I thought I would, i also think that `fn+123+` is a little to uneasily accessible. But it contains about every unicode character someone could want to write unicode-heavy formulas, so it does what it needs to do...

Julow:
I've added an option to add the switching key onto the main layout 2fc4051. I consider that `Fn+123+` or the dedicated key on the numeric layout are enough so I might remove it in the future depending on feedback.

Julow (approved):
I'll merge now :) Thanks for your work! I recognize I asked for a lot of changes and made incompatible changes on master at the same time.

PR #160, opened 2022-06-14
Move 'ъ' where it belongs and increase the width of the keys

Thunder-Squirrel:
The 'ъ' key is rarely used, and therefore we may move it to 'ь', as usual, and then increase the width of the keys.
ru


Julow (approved):
Thanks.

PR #180, opened 2022-07-18
Rename resulting CI artifact

sdrapha:
Add details to the name of the artifact, to distiguish downloads of it between multiple branches while testing

for example:
the resulting zip file will look like this:

Screenshot_20220718-010502_Brave

the unique name for each branch helps a lot while testing/debuging diferent versions


sdrapha:
It was working fine on my commited tests. but this pr failed, because of a slash. so it needs more work to get it right. Converting pr to a draft..

Julow (approved):
Looks nice, thanks!

PR #187, opened 2022-08-07
Brazilian portuguese translations

igorSilCar:
Updates to Brazilian Portuguese translations of values/strings.xml


Julow:
Thanks!

PR #199, opened 2022-09-12
Add optional NumPad

geroxyz:
First: Thanks for this absolut fabulous keyboard!
In my opinion, its a tiny flow, that in landscape mode the keys
are very wide. My first idea was to add much more horizontal border
to one or both sides. But now I implemented an optional NumPad to
use the extra space.

My changes are divided in 3 commits with further comments:
1) Add attribute 'blind' to KeyboardData.Key
(this change messed up a little bit when switching to Android Studio - sorry)
2) Use 'blind' keys instead of 'spacing'
3) Add optional NumPad

I would appreciate, if you adopt this idea!


res/values/arrays.xml
```diff
@@ -52,6 +52,33 @@
...
+    <item>never</item>
+    <item>landscape</item>
+    <item>always</item>
+  </string-array>
+  <string-array name="pref_show_numpad_entries">
+    <item>@string/pref_show_numpad_never</item>
+    <item>@string/pref_show_numpad_landscape</item>
+    <item>@string/pref_show_numpad_always</item>
+  </string-array>
+  <string-array name="pref_numpad_layout_entries">
```
Julow, line 65: I like customization but I think this is not the right way. The `pref_numeric` case is too crowded and doesn't look good, I don't want to think about that when modifying the numeric layout.

The period vs comma decimal separator is an interesting idea but I think it's not the best way to implement it. I'd like to have that on the regular numeric layout too. A better implementation would be a field in `method.xml` and then to perform arbitrary modifications like `Config.modify_layout`.

I suggest leaving this idea out of the PR for now. I'd prefer to have just one keypad that covers most cases (with the operators, if it's too wide, consider moving things into the corners of the digits).

srcs/juloo.keyboard2/KeyboardData.java
```diff
@@ -40,7 +42,31 @@ public KeyboardData addExtraKeys(Iterator<KeyValue> k)
     addExtraKeys_to_row(rows, k, 1, 3);
     addExtraKeys_to_row(rows, k, 2, 2);
     addExtraKeys_to_row(rows, k, 2, 1);
-    return new KeyboardData(rows, keysWidth, extra_keys);
+    return new KeyboardData(rows, keysWidth, extra_keys, num_pad);
+  }
+
+  public KeyboardData addNumPad()
```
Julow, line 48: "Blind" keys are a nice trick but I think they are not the best way. They show up all over the code and that's some subtle bugs in the future.

I think you can compute the needed "blind" space from here. The first key of each row of the numpad would need `this.keysWidth - row.keysWidth` of extra "shift".

Julow (commented):
I love the idea ! Thanks.

I have some comments on the implementation and I'd prefer to keep the changes smaller at first so we have the opportunity to questions the secondary features one at a time.

gero-streng:
I just added a new branch `gero-streng:optional-numpad-simplified` to my repository.
This solution works without 'blind keys' and has only one fixed NumPad layout as you proposed.

Julow:
Looks good :) Could you open a PR for it ?

The layout cache is important, if it problematic for the decimal separator feature, it's because it shouldn't be implemented that way. The decimal separator should be implemented similarly to `shouldOfferSwitchingToNextInputMethod` (see `Config.modify_layout`). I would prefer if it wasn't queried from system either, adding the information into `method.xml` and storing it in `Config` would be better.

Could you open a different PR for the decimal separator ? So we can discuss the implementation without impacting the first idea.

sandorex:
Not gonna lie it looks quite nice on my tablet
Screenshot_20221017-010353_Samsung Notes

Julow:
I merged `optional-numpad-simplified` without the `decimal_separator` and `grouping_separator` and a slight modification (baec5a7).

Thanks a lot !

PR #211, opened 2022-10-05
Added Hindi language layout

Raj9039852537:
This PR contains the Hindi keyboard layout. I hope everything is okay to merge it.

PR #212, opened 2022-10-06
Added Hindi layout

Raj9039852537:
First sorry for 2 pull requests this one is opened accidentally.
But since it contains one more change of defaulting the layout of Hindi to Hindi instead of QWERTY, I think i should go with it instead of old PR.
new is changes in methods.xml so that selecting hindi layout show hindi keyboard instead of default


Julow:
This is merged in https://github.com/Julow/Unexpected-Keyboard/pull/211

PR #252, opened 2022-12-11
Make the keyboard transparent

ChasmSolacer:
Quite missed this feature from the other keyboard I used in the past along with remote desktop viewers and emulators, so I decided to add it here. This setting is split into 3 options.

This are a few examples what it looks like with the following values:
background opacity: 70%
key opacity: 30%
key pressed opacity: 30%

Settings screen:
settings

Remote desktop app:
remote_desktop

### DIFF
--- a/srcs/juloo.keyboard2/Keyboard2View.java
+++ b/srcs/juloo.keyboard2/Keyboard2View.java
@@ -312,6 +312,7 @@ public class Keyboard2View extends View
     // Set keys opacity
     _theme.keyBgPaint.setAlpha(_config.keyOpacity);
     _theme.keyDownBgPaint.setAlpha(_config.keyActivatedOpacity);
+    _theme.keyBorderPaint.setAlpha(_config.keyOpacity);
     canvas.drawRoundRect(_tmpRect, r, r,
         isKeyDown ? _theme.keyDownBgPaint : _theme.keyBgPaint);
     if (w > 0.f)
--- a/srcs/juloo.keyboard2/Keyboard2View.java
+++ b/srcs/juloo.keyboard2/Keyboard2View.java
@@ -258,6 +258,10 @@ public class Keyboard2View extends View
     updateFlags();
     // Set keyboard background opacity
     getBackground().setAlpha(_config.keyboardOpacity);
+    // Set keys opacity
+    _theme.keyBgPaint.setAlpha(_config.keyOpacity);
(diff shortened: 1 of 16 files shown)


srcs/juloo.keyboard2/Keyboard2View.java
```diff
@@ -309,6 +309,9 @@ void drawKeyFrame(Canvas canvas, float x, float y, float keyW, float keyH,
     float w = isKeyDown ? _theme.keyBorderWidthActivated : _theme.keyBorderWidth;
     float w2 = _theme.keyBorderWidth / 2.f;
     _tmpRect.set(x + w2, y + w2, x + keyW - w2, y + keyH - w2);
+    // Set keys opacity
+    _theme.keyBgPaint.setAlpha(_config.keyOpacity);
+    _theme.keyDownBgPaint.setAlpha(_config.keyActivatedOpacity);
```
Julow, line 314: You might want to make the borders transparent also:

```
_theme.keyBorderPaint.setAlpha(_config.keyOpacity);
```

Also, this needs to be done only once and can be moved to the constructor of the class.

srcs/juloo.keyboard2/Keyboard2View.java
```diff
@@ -258,6 +256,8 @@ public void onLayout(boolean changed, int left, int top, int right, int bottom)
   protected void onDraw(Canvas canvas)
   {
     updateFlags();
+    // Set keyboard background opacity
+    getBackground().setAlpha(_config.keyboardOpacity);
```
Julow, line 260: This can be moved to the constructor as well.

Julow (approved):
Very nice, thanks! I have a few comments but I'll merge anyway for the next release planned today.

ChasmSolacer:
I made it the similar way how label brightness is implemented – it gets called inside drawLabel inside onDraw. When I move the sets to the constructor, I need to close the settings and wait a bit (?) before the changes take effect

ChasmSolacer:
Best I can do is move them to top of onDraw(). Performance has to be sacrificed here, sorry :slightly_smiling_face:

Julow:
The problem of the config not updating immediately is fixed in bf31872. It was the last thing I wanted to fix before a release.

At the beginning of `onDraw` is also fine. The drawing code needs more profound improvements so let's not think too much about it. I'll merge now and release.

PR #253, opened 2022-12-11
qwerty_pl layout is the default for Polish language

ChasmSolacer:
missed this little thing


Julow (approved):
Thanks! Just in time for a bug fix release.

PR #264, opened 2022-12-27
Create makefile.yml

quimsical:


Julow (changes requested):
What is the goal ? This doesn't work.

quimsical:
> 

im brand new to all this anf am just pushing buttons at this point trying to figure out how shot works.

Julow:
There are already some workflows: https://github.com/Julow/Unexpected-Keyboard/tree/master/.github/workflows
The ones you added won't work at all: python and clojure are not used in the project, the android workflow uses gradle which is not used, the makefile workflow is not adapted.

PR #268, opened 2023-01-12
Add modifier key to swap '.' <=> ','

geroxyz:
Add modifier key to swap decimal separator and group separator ('.' <=> ',') on numeric keyboard. The key can be selected in the extra key preferences.

__NOTE:__ the new version of `./srcs/special_font/result.ttf` isn't submitted here, because the creation of this file by `make rebuild_special_font`
fails on my installation (ubuntu) with
```
Import: Wrong number of arguments
Called from...
 build.pe: line 10
make: *** [Makefile:26: rebuild_special_font] Fehler 1
```
So, rebuild `result.ttf` it on your system, or take the version from my `makeshift-resources` branch, which is built by adding glyph 0x15 (`./srcs/special_font/15.svg`) to `master` version with fontforge manually.

### DIFF
--- a/res/xml/numeric.xml
+++ b/res/xml/numeric.xml
@@ -29,7 +29,7 @@
   <row height="0.95">
     <key width="1.5" key0="switch_text" key2="ctrl"/>
     <key width="1.5" key0="0" key3="f11_placeholder" key4="f12_placeholder"/>
-    <key width="0.75" key0="." key1=":" key2="," key3=";"/>
+    <key width="0.75" key0="decimal_separator" key1=":" key2="group_separator" key3=";" key4="loc swap_separator"/>
     <key width="0.75" key0="space" key1="&quot;" key2="'" key4="_"/>
     <key width="1.5" key0="enter" key1="±" key2="action" key3="="/>
   </row>
--- a/res/xml/numpad.xml
+++ b/res/xml/numpad.xml
@@ -20,7 +20,7 @@
     </row>
     <row height="0.95">
         <key key0="0"/>
-        <key key0="."/>
+        <key key0="decimal_separator" key1="loc group_separator" key2="loc swap_separator"/>
         <key key0="="/>
         <key key0="+"/>
     </row>
--- a/srcs/juloo.keyboard2/ExtraKeyCheckBoxPreference.java
+++ b/srcs/juloo.keyboard2/ExtraKeyCheckBoxPreference.java
@@ -36,6 +36,8 @@ public class ExtraKeyCheckBoxPreference extends CheckBoxPreference
     "£",
     "switch_greekmath",
     "capslock",
+    "group_separator",
+    "swap_separator",
     "copy",
     "paste",
     "cut",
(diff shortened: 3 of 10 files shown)


Julow (commented):
What's the use-case for the modifier key ?

geroxyz:
If I have to enter a bulk of data in Excel with german keyboard layout 
in landscape mode with NumPad, its very convenient, not to swipe for 
every numeric input with decimal part.

geroxyz:
A further issue would be: Activate Swap-Lock on german keybord layout 
by default, because that is the standard for german keyboards.

Julow:
With the comma and dot swapped, you wouldn't have to swipe anymore in Excel. Why isn't the option sufficient ?

What do you mean with Swap-Lock ?

geroxyz:
When I open Excel with german settings, the keyboard opens as shown below. 
In this case, the key right of the numeric `0` should have `,` as KeyValue by default.
You can achieve this by activating `.↔,` (SwapLock). Just like `Num↓` (NumLock) 
on a physical keyboard, which is activated initially.

 If this happens at startup, I could even deselect `.↔,` in the extra key preferences. 

Screenshot_20230113-133825_Excel

Julow:
Sorry I wasn't clear. An option in the setting that would swap `,` and `.` seems more suited than a modifier key. Do you often need to have easier access to the `.` ? If no, the ability to swap them with a modifier doesn't seem useful.

geroxyz:
Here some pros for the separator change by modifier key:

- for decimal data in german excel, the `,` as main KeyValue on the numeric keyboard is best.  
But if you have to enter series of date values (format: 31.12.2024) a quick switch to `.` 
is convenient.

- on my first approach for the optional NumPad (PR #199), 
I had the selection of the separator chars in the settings by having various 
NumPad layouts. But now I prefer the modifier variant, especially when running 
several apps with german settings (as Excel) and apps with english settings 
(programming, graphics) concurrently. 
The way over the keyboard settings is just more cumbersome, and increases 
the amount of setting entries.
(Btw: On my phone, I prefer portrait mode. But on my tablet, I use landscape with 
optional NumPad very often)

- the concept of modifier keys is used quite often at the _unexpected keyboard_.
The difference to other modifiers is, that only the 'locked' state makes sense.
The 'lock' state, where the modifier is cleared after the first subsequent 
key press is just useless.  But that's no obstacle for me.

cons:

- the 'swap' feature is somewhat hidden in the settings. Perhaps its better to move it 
more to the top of 'special keys' settings, or set the modifier visible (and activated?) 
by default for languages with swapped separator signs.

Julow:
I'd expect date to be typed rarely ? I'm not convinced the swap key is a better solution:
- A new kind of key, working like the numlock key on PC that I personally hate.
- The `,` and `.` will change place often. Keys on the keyboard shouldn't move too often. Too much thinking is needed for such a small feature.
- The "pointer" mechanism isn't ready for this new kind of key. It will often reset every modifiers. I'm not willing to spent so much time on fixing it for a relatively small feature.

Julow:
I'll close this PR as it was open for a very long time.

Issue #233, opened 2022-10-28
copy, paste, cut, undo, redo, select all - separate keys (not Ctrl + <key>)


OmlineEditor:
I ask you to create separate keys for: copy, paste, cut, undo, redo, select all...
I don't want to use hotkeys consisting of two combinations Ctrl + < key >

it will be very convenient when the user can create the appearance of the keyboard himself https://github.com/Julow/Unexpected-Keyboard/issues/231

AdePelumi:
Here is a sample from Google keyboard for the features he's requesting
![IMG_20221105_143029](https://user-images.githubusercontent.com/78900413/200137711-c28ce694-6ca2-48db-a310-9f68f9411838.jpg)

xi7:
i like the shortcut like computer - using ctrl a x c v   .... new layout looks slow

OmlineEditor:
you can add buttons to the keyboard yourself and it will be fast.

![keybrd](https://user-images.githubusercontent.com/13348933/201476813-130a79a6-9b58-48c1-8098-635095a1a076.jpg)

rancidfrog:
Florisboard has a wonderful implementation of copy paste clipboard etc.
![](https://i.imgur.com/55cDzsv.png)

xi7:
still no like all layout, too much height keyboard.  ctrl + a x c v faster and compact.

xi7:
![1](https://user-images.githubusercontent.com/67331785/201497520-125fe2a4-eb43-4645-9c08-2b855544c251.png)

arrow controlled by space bar key , and we can use a x c v using that empty key

OmlineEditor:
different people need different keyboards. therefore, it is necessary to make it possible for the user to choose independently how to do it. to do this, you need to create a custom keyboard by default https://github.com/Julow/Unexpected-Keyboard/issues/222

xi7:
will add a lot of complexity code and next u will open issue save/open customized backup.

i still prefer compact and faster possible

OmlineEditor:
backup from old versions is not suitable for new versions. This is better than simplifying the program.

Julow:
I've added some editing keys in 2aa98de as a first step for https://github.com/Julow/Unexpected-Keyboard/issues/242 (the goal was to make that fit in the code, 078dbcd). Undo and redo are missing and there's no symbol for them yet (if any of you is interested in continuing the work).
[...]

PR #282, opened 2023-01-20
Add layout 'qwerty_he_il'

RamKromberg:
Hebrew


res/values/arrays.xml
```diff
@@ -42,6 +43,7 @@
     <item>QWERTY (Norwegian)</item>
     <item>QWERTY (Swedish)</item>
     <item>QWERTY (Hungarian)</item>
+    <item>QWERTY (Hebrew)</item>
```
Julow, line 46: Is it commonly called like that ? It doesn't seem like qwerty at all, with punctuation and backspace on the first row and 11 columns.

Julow: Wikipedia calls it "Standard Hebrew keyboard": https://en.wikipedia.org/wiki/Hebrew_keyboard
I think you can rename the layout to "Hebrew" here.

RamKromberg: Physical Hebrew keyboard layouts have QWERTY capitals built-into the shifting / Caps Lock but Google didn't implement it (which is why I moved the shift to a gesture) so you make a valid point.

I guess I'll try and see how much I can implement out of SI-1452-1 first and adjust the naming accordingly.

Ref:
SI-1452-1: http://standards.har-el.com/Projects/kb/PRD_01452-01-00-0-20.pdf (p.10 for the layout picture)
SI-1452-2: http://standards.har-el.com/Projects/kb/Draft%20SI%2001452-02-00-0%20.pdf (p.11 for the layout picture)

Note the latter standard wasn't widely adopted.

Julow: Please add the layout that you would use.

The name "QWERTY (Hebrew)" seems closer to SI-1452-1 and 2, which is not the layout you wrote. I'm not sure such a layout is useful when we have the "secondary" layout option to switch between Hebrew and QWERTY easily, except if it's the layout that you wish use.

RamKromberg:
I ended up doing both to make sure there's as few conflicts as possible between the layouts. The newer one is clearly better but most people will opt for the former out of familiarity so I don't want to force it...

Btw, while trying to add diacritics I realized I need optional label1, label2...label5 entries to match the keys for at least three reasons:
1. Bidi languages (Arabic, Hebrew and Syriac) flip the parenthesis, brackets and greater/lesser-than when typing while leaving the keyboard symbols l2r. So, those layouts should have something like 'key3="[" label3="]"' and the likes.
2. Bidi directional markers (U+200F and U+200E) are invisible so they'll need labels. e.g. 'key3="‫‪&#x200E;‬‬" label3="LRM"'.
3. [Combining Diacritical Marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks) and [Hebrew Diacritics](https://en.wikipedia.org/wiki/Hebrew_diacritics) in particular render incorrectly without a charterer.

I'm not sure if/when/how I'll get this sorted but I thought it's best to let you know before spending time on this.

res/xml/he_il_1452_1.xml
```diff
@@ -0,0 +1,48 @@
...
+    <key key0="ע" width="1.0333"/>
+    <key key0="י" key2="-" key3="_" width="1.0333"/>
+    <key key0="ח" key2="=" key3="+" width="1.0333"/>
+    <key key0="ל" key2="]" key3="}" width="1.0333"/>
+    <key key0="ך" key2="[" key3="{" width="1.0333"/>
+    <key key0="ף" key1=":" key2="," key3="&quot;" width="1.4006"/>
+  </row>
+  <row>
+    <key key0="\\" key1="|" key2="shift" key3="loc capslock"/>
+    <key key0="ז" key1="undo" key3="redo"/>
```
Julow, line 30: `undo`, `redo`, etc.. can be added in the settings. You don't need to specify them on the layout. Further development may make these redundant too (https://github.com/Julow/Unexpected-Keyboard/issues/233).

res/xml/he_il_1452_1.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard bottom_row="false">
```
Julow, line 2: Why do you need a specific bottom row ? This adds maintenance, as the bottom row can evolve independently from the rest of the layout. For example, it changed today: 854eff2

res/xml/method.xml
```diff
@@ -9,6 +9,7 @@
   <subtype android:label="%s" android:languageTag="es" android:imeSubtypeLocale="es_ES" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty_es,extra_keys=accent_grave|accent_aigu|accent_tilde|accent_trema|€"/>
   <subtype android:label="%s" android:languageTag="fr" android:imeSubtypeLocale="fr_FR" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=azerty,extra_keys=accent_grave|accent_aigu|accent_circonflexe|accent_cedille|accent_trema|€"/>
   <subtype android:label="%s" android:languageTag="hu" android:imeSubtypeLocale="hu_HU" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz_hu,extra_keys=accent_aigu|accent_trema|accent_ogonek|€"/>
+  <subtype android:label="%s" android:languageTag="he" android:imeSubtypeLocale="he_IL" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=he_il_1452_1"/>
```
Julow, line 12: Here you can specify keys that will be added to every layouts with `,extra_keys=`. This can be useful for bidi markers if you are regularly switching between two layouts.

Julow (commented):
This can be done by adding keys in `srcs/juloo.keyboard2/KeyValue.java` (near the end of the file).
A "char" key can type a character but display an other on its label. You can add a `bidi-[` and a `bidi-]` key that have the right label. Same for the markers, which can have a specific label.

res/xml/method.xml
RamKromberg: Actually the bidi markers are always used when switching back and forth:
1. Write English; Switch to Hebrew; Write a word and finish with a question mark / exclamation mark; Insert RLM; Switch to English; Write English...
2. Write Hebrew; Switch to English; Write a word; Switch to Hebrew; Write question mark / exclamation mark; Insert LRM; Write Hebrew...

res/xml/he_il_1452_1.xml
RamKromberg: It was done for aesthetics reasons but those aren't relevant anymore after the recent changes so removed.

res/xml/he_il_1452_1.xml
RamKromberg: k removed.

RamKromberg:
K I added the missing accents and hacked a reasonable presentation for them, flipped the affected bidi labels and cleaned up the layouts to something hopefully more maintainable.

There's a "bug" with the FN key alternates where the bracket, parenthesis etc aren't being flipped like how the [hacked bidi versions of those brackets](https://github.com/Julow/Unexpected-Keyboard/pull/282/files#diff-04dc8c9b19a6d78afece4a268d9f2c5b1356079b5dea73bad2fb3d5a02160fddR352) do. I can't think of any easy fix.

p.s.
Btw, I found a Dvorak version of the second layout: http://mikladot.com/dvorak/english.html
It's not part of a standard but it was done by the same guy who developed the standard: http://mikladot.com/english.html

RamKromberg:
Still WIP: I'm trying to do the FN switching but https://github.com/Julow/Unexpected-Keyboard/commit/fecc4dd70ae8e5f2fc40e9679c99b306768cf603 broke the placeholders I need for some of the accents so don't merge :D

I've thrown in some logs around https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L128 for logcat over adb and the issue seems to be that _KeyValue.getKeyByName("f11_placeholder")_ and _KeyValue.getKeyByName("f12_placeholder")_ are never equal to k... But I don't have a good understanding of the code or even a debugger setup (I'm using your nix-shell with make and editing the layouts with a text editor) so take it with a grain of salt ;)

res/xml/method.xml
Julow: That's what `extra_keys` is for.

Julow:
Thanks for noticing, I fixed the placeholder bug in 2539fea.

If specifying an unicode character as the symbol of a key is not enough, you can add a glyph in SVG format in `srcs/special_font`. You can extract these from an open source font using fontforge.

RamKromberg:
I thing I got everything done except:
1. I didn't explore the glyphs option. I admit it's the right thing to do here but it's a lot of work for a minor visual improvement so... Yeah.
2. There's still a minor bug with the FN key where the bidi hacks (e.g. "b[") FN-alts are showing flipped... But I can't imagine it being easy to fix so I'm not even going to try.

So, it's ready as far as I can tell? Let me know if anything needs fixing.

Julow:
To handle FN-alternative of bidi keys, a new kind of key might be needed. For copy/paste/undo/redo, ideally it wouldn't have anything to do with the layout.

Thanks for the work, I'll merge now.

RamKromberg:
Thanks!
> a new kind of key might be needed
Yeah I just couldn't come up with a proper solution for it that didn't call for major surgery. So, I just hacked the minimum with the placeholders so I'll have a keyboard that doesn't suck and moved on :D

FYI, Originally I had all the Hebrew keys as placeholders and when pressing FN, only the accents would show. It was REALLY nice... But it completely defeated the separation between data and code so I cut it out and only kept those last remaining placeholders to keep the rare accents from littering the default unmodified (by the FN modifier...) layout.

> ideally it wouldn't have anything to do with the layout.
I'm surprised you haven't considered replacing the xml layouts with lua scripts like [like how koreader works](https://github.com/koreader/android-luajit-launcher) by now. I mean, a quick look at KeyModifer.java makes it pretty clear the separation of layout data and code is just not working out... And you just know there will always be that one guy asking for a "Share with Whatsapp" shortcut button... :D

Julow:
I don't think of code and data the same way. Consider that FN combinations, keys identified in KeyValue, etc.. are data in code.
I haven't thought of embedding a language but I don't want the keyboard to be complex enough to need this. The KeyModifier class haven't been challenged much until now but I'd prefer to maintain code that is useful rather than sink time in the most general solution. Let's see how the keyboard works for you before changing things.

RamKromberg:
> but I don't want the keyboard to be complex enough to need this...Let's see how the keyboard works for you before changing things.

Oh I'm mostly satisfied with the way things are. I was just referring to how all those different feature requests added up to something fairly Turin-complete sounding to me and wanted to acknowledge the minor defects in my bidi hacks.

Regardless, whether the layouts are done in xml, pure java, a combination of both or in lua means little to me since I just end up picking up the code and compiling it.

If the itch to tinker comes up, I'll limit myself to the SVGs per your suggestion.

Issue #290, opened 2023-02-07
[Feature Suggestion] Accents for Vietnamese keyboards


rVnPower:
Hello! I am very interested in this project and I wanted to contribute too. We are currently translating the app and implementing the Vietnamese layout for this keyboard. Only the accents for some Vietnamese characters are missing from the list. 

- The hook above (u -> ủ)
- The underdot (u -> ụ)
- The horn (u -> ư)

Vietnamese keyboards also use the breve, the acute, the grave, the tilde and the circumflex, which were already included in the list.

Julow:
Thanks for your interest. @K4zoku has implemented the accents and written the layout in https://github.com/Julow/Unexpected-Keyboard/pull/291
Feel free to review the PR, especially the placement of the accents.

PR #291, opened 2023-02-07
Vietnamese keyboard

K4zoku:
Added missing accent (horn, hook above, dot below)
Added layout (~~I am just putting new accents in random position~~ Now it's telex layout)
Added accent font svg (by modified existing accents)

Related issue(s): #290


res/xml/qwerty_vi.xml
```diff
@@ -0,0 +1,37 @@
...
+  </row>
+  <row>
+    <key shift="0.5" key0="a" key1="tab" key2="`" key3="â"/>
+    <key key0="s" key1="loc accent_aigu"/>
+    <key key0="d" key1="loc accent_bar" key2="₫" key3="đ"/>
+    <key key0="f" key1="loc accent_grave"/>
+    <key key0="g" key2="-" key3="_"/>
+    <key key0="h" key2="=" key3="+"/>
+    <key key0="j" key1="loc accent_dot_below" key3="{" key4="}"/>
+    <key key0="k" key1="loc accent_hook" key3="[" key4="]"/>
```
Julow, line 23: There's no `accent_hook`.

res/xml/qwerty_vi.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+  <row>
+    <key key0="q" key1="~" key2="1" key4="esc"/>
+    <key key0="w" key1="loc accent_breve" key2="2" key3="\@" key4="loc accent_horn"/>
+    <key key0="e" key1="!" key2="3" key3="\#" key4="ê"/>
+    <key key0="r" key2="4" key3="$" key1="loc accent_hook_above"/>
```
Julow, line 7: All the accents that you need shouldn't have the `loc ` prefix. Like this, someone who has not the languages enabled on their device but decide to use this layout won't see the accents.

srcs/juloo.keyboard2/KeyModifier.java
```diff
@@ -60,6 +60,9 @@ public static KeyValue modify(KeyValue k, KeyValue.Modifier mod)
       case SLASH: return apply_map_char(k, map_char_slash);
       case BAR: return apply_map_char(k, map_char_bar);
       case ARROW_RIGHT: return apply_combining(k, "\u20D7");
+      case DOT_BELOW: return apply_dead_char(k, '\u0323');
+      case HORN: return apply_map_char(k, map_char_horn);
+      case HOOK_ABOVE: return apply_dead_char(k, '\u0309');
```
Julow, line 65: `DOT_BELOW` and `HOOK_ABOVE` don't work.
It seems that these characters use combining diacritics ? In this case, you should use `apply_combining`.

Julow (commented):
Thanks for the great work!

srcs/juloo.keyboard2/KeyModifier.java
K4zoku: no, I think I need to write a map char for this. Currently, I can type all vietnamese characters, but a little bit janky (e.g: ạ vs ạ is different)

res/xml/qwerty_vi.xml
K4zoku: removed in latest commit

res/xml/qwerty_vi.xml
K4zoku: i forgor 💀
btw, removed in latest commit

srcs/juloo.keyboard2/KeyModifier.java
K4zoku: I'll add map char for those two accent soon

K4zoku: If use `apply_combining`, the other accents should be use it too and the whole code is only for Vietnamese, which is a big deal.

res/xml/qwerty_vi.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+  <row>
+    <key key0="q" key1="~" key2="1" key4="esc"/>
+    <key key0="w" key1="loc accent_breve" key2="2" key3="\@" key4="accent_horn"/>
```
Julow, line 5: Breve is needed ? It shouldn't have the `loc` prefix then. The `loc` prefix is used to place a key but to show it only if it is an "extra key" or is enabled in the settings (eg. like `capslock`)

K4zoku: oh, I forgor again 💀
thanks for figure it out
(it's 4:53 AM now and I haven't sleep yet)

srcs/juloo.keyboard2/KeyModifier.java
Julow: Do you mean that Vietnamese uses combining diacritics for breve, aigu, grave, etc.. ?

K4zoku:
btw, the newly added svgs need to be reposition to get aligned with the others

srcs/juloo.keyboard2/KeyModifier.java
K4zoku: yes
ex: `a á à ả ã ạ`

res/xml/qwerty_vi.xml
K4zoku: Removed in latest commit

srcs/juloo.keyboard2/KeyModifier.java
Julow: In this case, you can simply use `apply_combining` to implement `DOT_BELOW` and `HOOK_ABOVE`.

For the other accents, you'll need to define new versions of them (eg. `AIGU_COMBINING`, etc..) that use the same symbol but are implemented using `apply_combining`.

K4zoku: The combining is janky, I added map char and it now work as I (a Vietnamese guy) expected

K4zoku: if combining, the accent and character before are two characters, meanwhile almost Vietnamese typing engine didn't use that, they use an actual character

K4zoku: The keyboard is perfect for me now. If someone has problem about the Vietnamese layout and accent when using UnexpectedKeyboard, just mention me

K4zoku:
Screenshot_20230209-235331

Screenshot for the latest layout change

srcs/juloo.keyboard2/KeyModifier.java
Julow: That works for me.

Julow (approved):
Thanks a lot for your work!

I pushed a commit to add the missing check boxes. This is a hack that I should remove soon.

PR #293, opened 2023-02-09
[HotFix] Added missing accent that can combine with `horn_accent`

K4zoku:
After couple hours of testing, I found this is missing


Julow (approved):
Thanks!

PR #313, opened 2023-03-09
New Hindi layout

lrvideckis:
Hi! First thank you for creating/maintaining a great keyboard!

Next my credentials: I'm an American (native English speaker) learning Hindi. I have a few friends from Goa, India who are fluent in Hindi, and have given me advice on this layout.

Some misc things:
- I suggest we change it from "Hindi" to "Devanagari". Devanagari is the name of the alphabet/set of characters which Hindi is written in. But Devanagari is more general as some other languages (Nepali, Marathi, Konkani, ...) are written in Devanagari.
- This layout contains all characters from the Devanagari unicode block https://en.wikipedia.org/wiki/Devanagari_(Unicode_block) as well as borrowed punctuation https://blogs.transparent.com/hindi/punctuation-in-hindi/. I compared it to GBoard's Hindi layout, and this layout should be able to type all characters from there.
- I looked at frequency counts https://www.sttmedia.com/characterfrequency-hindi and attempted to place high-frequency characters as a tap (instead of swipe). Although this is not an improvement over the previous layout, as the previous layout also has frequent characters as a tap
- I've been typing on this layout for about a month now, and this is what I've converged on as my "endgame"
- I removed `shift` as it isn't used in Devanagari script (and also caps lock)

---

Some possible downsides of this layout:
- I removed some of the ASCII characters in the previous layout (like `[`, `]`, `(`, ..., `#`, `&`). One of my hindi-fluent friends said these characters are never used in the language, besides borrowed characters https://blogs.transparent.com/hindi/punctuation-in-hindi/ which I did include. Despite this I see your point about these characters being useful for passwords. I did create a version where I crammed in many ASCII symbols, but then the layout looked cluttered overall. I guess it is a preference, and my preference is the cleaner look (without these symbols). I did leave some empty spaces in some keys if you want to re-add these ASCII characters
- My layout doesn't contain any conjunct consonants https://en.wikipedia.org/wiki/Devanagari_conjuncts  (except for like 4 of the most frequent ones). To be clear, this layout is still able to type these, just with more keystrokes (example: न + ् + य = न्य). 
- I didn't include any `loc` characters.


res/xml/devanagari.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+  <row>
+    <key shift="0.35" width="0.9" key0="क" key1="ख" key2="घ" key3="ङ"  key4="ग"/>
```
lrvideckis, line 4: The biggest difference between my layout and the current Hindi layout is the following:

The devanagari characters are organized by mouth/tongue movements https://omniglot.com/writing/devanagari.htm . So for example ग=g, क=k (these are similar tongue movements). I've grouped all similar-tongue-movement characters into the same key

res/xml/devanagari.xml
```diff
@@ -0,0 +1,30 @@
...
+    <key              width="0.9" key0="च" key1="छ" key2="झ" key3="ञ"  key4="ज"/>
+    <key              width="0.9" key0="ट" key1="ठ" key2="ढ" key3="ण"  key4="ड"/>
+    <key              width="0.9" key0="त" key1="थ" key2="ध" key3="न"  key4="द"/>
+    <key              width="0.9" key0="प" key1="फ" key2="भ" key3="म"  key4="ब"/>
```
lrvideckis, line 8: another example: म=m, फ=f, ब=b (these are all lip-movements)

res/xml/devanagari.xml
lrvideckis: I talk to one of my hindi-speaking friends, and he said that this grouping (by tongue movements) is the best way to organize the alphabet (and thus also a keyboard layout)

res/xml/devanagari.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+  <row>
```
lrvideckis, line 3: This row contains only the the consonants https://omniglot.com/language/articles/devanagari.htm

res/values/arrays.xml
```diff
@@ -61,7 +61,7 @@
     <item>Bone</item>
     <item>ЙЦУКЕН (Українська)</item>
     <item>বাংলা</item>
-    <item>हिन्दी</item>
+    <item>देवनागरी (हिंदी)</item>
```
lrvideckis, line 64: Regarding the spelling of Hindi https://forum.unilang.org/viewtopic.php?t=10200

srcs/juloo.keyboard2/KeyModifier.java
```diff
@@ -281,7 +281,62 @@ private static String apply_fn_char(char c)
       case 'ס': return "segol";
       case 'ב': return "hataf_segol"; // reduced segol
       case 'צ': return "tsere";
-      // other
+      // Devanagari symbols
```
lrvideckis, line 284: So I'm using the `fn` key to allow a second "layer". Before, I tried to squeeze all the symbols into a single layer (kinda like the previous hindi layout), but I didn't like how congested/squeezed it turned out.

So for my layout, the original, non-`fn` layer contains the most-used characters, and the `fn` layer contains seldom used characters

res/xml/devanagari.xml
```diff
@@ -0,0 +1,30 @@
...
+    <key              width="0.9" key0="ै" key1="ऐ" key2="ऌ" key4="ॢ"/>
+    <key              width="0.9" key0="ो" key1="ओ" key2="औ" key4="ौ"/>
+  </row>
+  <row>
+    <key key0="्" key2="tab" key4="esc"/>
+    <key key0="ँ" key1="₹" key2="॑" key3="ॖ" key4="॓"/>
+    <key key0="ं" key1="ॐ" key2="ऽ" key3="ः" key4="&quot;"/>
+    <key key0="।" key1="," key2=";" key3="!" key4="\?"/>
+    <key key0="़" key1="॰" key2="०" key3="-" key4="॒"/>
+    <key key0="५" key1="१" key2="३" key3="७" key4="९" key5="४" key6="६" key7="२" key8="८"/>
```
lrvideckis, line 27: I wasn't really sure how to add the numbers, so I just threw them in a number pad

lrvideckis:
Screenshot_20230308-203114
Screenshot_20230308-203158

Julow (approved):
Thanks for your great work! The layout looks nice and well researched.

@Raj9039852537 Do you still use this app ?

In case someone wants the previous "hindi" layout back, how should we call it ?

You need to change the `default_layout` of the hindi subtype (`hi_IN`) in res/xml/method.xml. If that makes sense, you can also add support for the languages you mentionned (Nepali, Marathi, Konkani).

> I removed some of the ASCII characters in the previous layout (like [, ], (, ..., #, &).

That's not a problem. The "secondary layout" option can be used to bring a layout suitable for passwords.

lrvideckis:
> In case someone wants the previous "hindi" layout back, how should we call it ?

Good point, we can just keep the previous name (I'm about to add it back)

Raj9039852537:
> @Raj9039852537 Do you still use this app ?
Yes, I am using it.

As for naming the layout I suggested is inscript layout used vastly in PC. 
>hi_IN subtype 
Providing old layout is usefuI for peoples who are more comfertable with that layout. 

Also if it is possible, a choice to edit layout at runtime (just like Multiling O Keyboard ) will be an awesome addition.

Julow (approved):
Thanks!

lrvideckis:
I just wanted to post an update about this: I still use this layout daily mainly to text my Indian friends.

I did make one tweak:
Screenshot_20231014-095817~2
I changed ; to * so that I can correct my previous text. And I noticed I never use ; (or ,) symbols anyways.

Finally I did write a blog about this https://www.lrvideckis.com/blog/2023/08/21/devanagari_keyboard.html

Julow:
Thanks for the update. Could you open a new PR to update the layout for everyone ?

PR #314, opened 2023-03-09
Add Arabic layout

mostafaayesh:
This request is to add Arabic Keyboard

The layout follows closely IBM PC (https://en.wikipedia.org/wiki/Arabic_keyboard#IBM_PC_Arabic_Keyboard) 
Only alteration is the letter (ذ) usually in the number row was moved to the end of the first row. This layout is the most common from my experience.

The following are a screenshots of the current layout, suggestions are welcome.

Without number row:
without_numrow

With number row:
with_numrow

### DIFF
--- a/res/xml/ar_pc.xml
+++ b/res/xml/ar_pc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <keyboard>
   <row>
-    <key key0="ض" key1="&#1614;" key2="`" key4="esc"/>
+    <key key0="ض" key1="&#1614;" key2="1" key3="`" key4="esc"/>
     <key key0="ص" key1="&#1611;" key2="2" key3="\@"/>
     <key key0="ث" key1="&#1615;" key2="3" key3="\#" key4="loc €"/>
     <key key0="ق" key1="&#1612;" key2="4" key3="$" key4="loc £"/>
--- a/res/xml/ar_pc.xml
+++ b/res/xml/ar_pc.xml
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <keyboard>
   <row>
-    <key key0="ض" key1="&#1614;" key2="1" key3="`" key4="esc"/>
-    <key key0="ص" key1="&#1611;" key2="2" key3="\@"/>
-    <key key0="ث" key1="&#1615;" key2="3" key3="\#" key4="loc €"/>
-    <key key0="ق" key1="&#1612;" key2="4" key3="$" key4="loc £"/>
-    <key key0="ف" key1="&#1604;&#1573;" key2="5" key3="%"/>
-    <key key0="غ" key1="&#1573;" key2="6" key3="^"/>
-    <key key0="ع" key1="&#8216;" key2="7" key3="&amp;"/>
-    <key key0="ه" key1="&#0247;" key2="8" key3="*"/>
-    <key key0="خ" key1="&#0215;" key2="9" key3="("/>
-    <key key0="ح" key1="&#1563;" key2="0" key3=")"/>
+    <key key0="ض" key1="&#1614;" key2="١" key3="`" key4="esc"/>
+    <key key0="ص" key1="&#1611;" key2="٢" key3="\@"/>
+    <key key0="ث" key1="&#1615;" key2="٣" key3="\#" key4="loc €"/>
+    <key key0="ق" key1="&#1612;" key2="٤" key3="$" key4="loc £"/>
+    <key key0="ف" key1="&#1604;&#1573;" key2="٥" key3="%"/>
+    <key key0="غ" key1="&#1573;" key2="٦" key3="^"/>
+    <key key0="ع" key1="&#8216;" key2="٧" key3="&amp;"/>
+    <key key0="ه" key1="&#0247;" key2="٨" key3="*"/>
+    <key key0="خ" key1="&#0215;" key2="٩" key3="("/>
+    <key key0="ح" key1="&#1563;" key2="٠" key3=")"/>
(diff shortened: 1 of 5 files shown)


res/xml/ar_pc.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard>
+  <row>
+    <key key0="ض" key1="&#1614;" key2="`" key4="esc"/>
```
Julow, line 4: If you have `esc`, I think `tab` can be useful.

Julow (approved):
Many thanks!

> Only alteration is the letter (ذ) usually in the number row was moved to the end of the first row.

This makes the keys very small. What do you think of placing it either in the bottom-left corner of `د` (currently taken by `+`) or on the second row ?

Similarly, shift doesn't seem useful and can be removed or made smaller to gain some space.

mostafaayesh:
I appreciate your feedback, I have made adjustments to the layout.

I also added an alternate layout similar to Samsung keyboard. Most of us using Arabic keyboards on computers or phones very rarely use the special characters.

Adjusted layout:
msg819505937-58988

Alternate layout (similar to Samsung keyboard)
msg819505937-58987

Reference Samsung keyboard layout.
msg819505937-58986

mostafaayesh:
I realized I was missing number "1" in the first row so I added that

islam2hamy:
Hello ,
Thank you @mostafaayesh  for your work
For me i like this one

> With number row: with_numrow

Also About (ذ) replace "shift" with it

islam2hamy:
Also , can you please make numbers like this ١ ٢ ٣ ...

res/xml/method.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <input-method xmlns:android="http://schemas.android.com/apk/res/android" android:settingsActivity="juloo.keyboard2.SettingsActivity" android:supportsSwitchingToNextInputMethod="true">
+  <subtype android:label="%s" android:languageTag="ar" android:imeSubtypeLocale="ar" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=ar_pc"/>
+  <subtype android:label="%s" android:languageTag="ar" android:imeSubtypeLocale="ar" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=ar_alt"/>
```
Julow, line 4: There cannot be two entry here. The `default_layout` specifies the layout that people will see by default (until they change it in the settings).

Julow (approved):
Both layouts look good. You can try the debug version of the app here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/4377386493

I'll merge soon, feel free to make more changes to the layouts.

islam2hamy:
> Both layouts look good. You can try the debug version of the app here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/4377386493
> 
> I'll merge soon, feel free to make more changes to the layouts.

Ok , it is look good
P.s parentheses need to be corrected (ex. Next video)

https://user-images.githubusercontent.com/7989030/224122224-788b4f7b-4cef-40eb-bdd0-7c335baf9d4a.mp4

mostafaayesh:
> > Both layouts look good. You can try the debug version of the app here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/4377386493
> > I'll merge soon, feel free to make more changes to the layouts.
> 
> Ok , it is look good P.s parentheses need to be corrected (ex. Next video)

The parentheses seem to be a Right to left (RTL) issue, I will try to look if there is a solution for it.

mostafaayesh:
> > > Both layouts look good. You can try the debug version of the app here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/4377386493
> > > I'll merge soon, feel free to make more changes to the layouts.
> > 
> > 
> > Ok , it is look good P.s parentheses need to be corrected (ex. Next video)
> 
> The parentheses seem to be a Right to left (RTL) issue, I will try to look if there is a solution for it.

I looked into it, the first character is always left to right, if you're using a letter you wouldn't notice an issue, but if you're using a bracket or parenthesis it is the wrong direction. 

In the Hebrew layout "b{" is used but that only fixes the brackets as first character and breaks them for after the direction is changed. The solution was to implement unicodes "lrm" and "rlm" as extra keys so the user can manually switch the direction but it's not very intuitive. I personally prefer not having the "b{" because it only fixes the instance of the character at the beginning of a text field in LTR, but if other people want that I don't see an issue with changing them.

I'm currently looking into a way to implement something similar to AutoCaptilization but for RTL language to fix the direction of the first character. I need some time to get familiar with the source code, if I figure it out I will open a new Pull Request for a fix that should work for all RTL languages.

In the meantime, if the layout works well for most people, I think this can be merged.

islam2hamy:
١ ٢ ٣ ٤ ...
PleasE

mostafaayesh:
> ١ ٢ ٣ ٤ ... Please

It should have the numbers now, only caveat is if you enable the number row, you will have both types.
msg819505937-58996

Julow (approved):
If you need both number scripts, you can use the `Fn` modifier to transform ١..٩ into 1..9. The opposite is also possible by creating new keys `ar_1`..`ar_2` that normally type 1..9 except when transformed by the `Fn` modifier.
Modifying the number row or the numpad is more involved but I would accept the changes.

> I'm currently looking into a way to implement something similar to AutoCaptilization but for RTL language to fix the direction of the first character.

That would be a wonderful addition!

I'll merge this now, thanks for your great contribution!

islam2hamy:
hello @mostafaayesh 
please add (  ّ ) on ( د ) button
thank you

PR #320, opened 2023-04-08
Do not lock modifiers if sliding a key

### LINKED ISSUE #319: Modifier + slide on spacebar (cursor movement) keeps modifier stuck
The behaviour for e.g. `Ctrl` + `Right` differs between the two possible input variants:
* Holding `Ctrl`, slide on _cursor key_ to right.
* Holding `Ctrl`, slide on `Space` to right.

The second variant, i.e. using a slide on the space bar, does not release the `Ctrl` key but locks it instead. This is true for all modifiers.
At least for me this behaviour is unexpected.

Since the slide on the space bar is an relative fresh feature, I assume this is just "configured" wrongly.

doak:
This unconditionally removes all pointers (touches) pressing modifiers which are not already locked, avoiding that the (currently) latched modifier will be locked aventually.

This can be verfified while sliding the space bar to move the cursor left or right.

Closes #319.

### DIFF
--- a/srcs/juloo.keyboard2/Pointers.java
+++ b/srcs/juloo.keyboard2/Pointers.java
@@ -113,6 +113,7 @@ public final class Pointers implements Handler.Callback
       return;
     if (ptr.sliding)
     {
+      clearLatched();
       onTouchUp_sliding(ptr);
       return;
     }


doak:
This seems quite too easy. I am missing some (corner) cases for sure.
It works on my machine™ at least. ;)

Julow (approved):
Thanks a lot! This is how it's supposed to behave and the code looks right.

Merging now.

Issue #329, opened 2023-04-14
Improve layout_of_string and arrays.xml


GrimPixel:
Every time a new layout is added, the layout also need to be added here. Is there any ways to simplify that case switch?

Julow:
This is a long time issue that I would like to fix at some point.
An other important issue to consider is the fragile pairing between `pref_layout_values`/`entries`.

Julow:
This is changed in 8ba82d2 and 15c608b. See the updated guidelines: https://github.com/Julow/Unexpected-Keyboard/blob/master/CONTRIBUTING.md#adding-a-layout

Feedback welcome.

Issue #142, opened 2022-05-10
Change the names of keyboard layouts


GrimPixel:
They are currently a mixture of English and native languages, from word to spelling rules. 
Here are the names I consider consistent:

QWERTY (Brasileiro)
QWERTY (português, Brasil)

QWERTY (Español)
QWERTY (español, España)

QWERTY (Korean)
두벌식 (한국어, 대한민국)

QWERTY (Latvian)
QWERTY (latviešu)

QWERTY (Swedish)
QWERTY (svenska)

ЙЦУКЕН (Русский)
ЙЦУКЕН (русский)

QWERTZ (Hungarian)
QWERTZ (magyar)

Bulgarian (Phonetic Traditional)
ЯВЕРТЪ

In addition, names in “pref_layout_values” can be changed as well:

qwerty_pt
qwerty_pt_br

qwerty_es
qwerty_es_es

qwerty_ko
dubeolsik_ko

qwerty_sv_se
qwerty_sv

ru_jcuken
jcuken_ru

bgph1
iavertie

Issue #325, opened 2023-04-14
Select script first, layout next


GrimPixel:
As the number of layout grows, it becomes necessary to split them into groups with the writing system.

Julow:
Agree! Are you willing to implement this ?

GrimPixel:
I think it requires a menu for the writing system and a secondary menu for the language. I have to learn first.

Julow:
This might be easier to implement since recent changes. The layout selection dialog is now opened here: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/LayoutsPreference.java#L70

PR #331, opened 2023-04-15
Standarised layout names

GrimPixel:
This solves the issue https://github.com/Julow/Unexpected-Keyboard/issues/142

Layout are named in a specific pattern: for “pref_layout_values”, they are prefixed with ISO 15924 and ISO 639-3 codes; for “pref_layout_entries”, they are prefixed with the native language name in the corresponding writing system.


res/values/arrays.xml
```diff
@@ -2,74 +2,74 @@
...
-    <item>Arabic PC</item>
-    <item>Arabic ALT</item>
+    <item>العربية: IBM</item>
+    <item>العربية: IBM [Mostafa Ayesh]</item>
+    <item>বাংলা: [Md Rasel Hossain]</item>
+    <item>български: ЯВЕРТЪ</item>
+    <item>русский: ЙЦУКЕН</item>
+    <item>українська: ЙЦУКЕН</item>
+    <item>हिंदी: InScript</item>
+    <item>हिदी: [Luke Videckis]</item>
```
lrvideckis, line 48: spelling of hindi is wrong: it's missing that dot (it's spelled correctly the line above)

it may not seem like a big deal, but since hindi is phonetic, "हिदी" is pronounced without the 'n' like "hidi"

lrvideckis: more about the dot https://en.m.wikipedia.org/wiki/Anusvara

res/values/arrays.xml
```diff
@@ -2,74 +2,74 @@
...
+    <item>Deutsch: Neo 2</item>
+    <item>Deutsch: QWERTZ</item>
+    <item>English: Colemak</item>
+    <item>English: Dvorak</item>
+    <item>English: QWERTY</item>
+    <item>English: QWERTZ</item>
+    <item>español: QWERTY</item>
+    <item>français: AZERTY</item>
+    <item>latviešu: QWERTY</item>
+    <item>magyar: QWERTY</item>
```
lrvideckis, line 64: should some of these be capitalized? (since we're standardizing)

res/values/arrays.xml
```diff
@@ -2,74 +2,74 @@
...
+    <item>hebr_heb_si_1452_2</item>
+    <item>kore_kor_ks_x_5002</item>
+    <item>latn_ces_qwertz</item>
+    <item>latn_deu_bone</item>
+    <item>latn_deu_neo2</item>
+    <item>latn_deu_qwertz</item>
+    <item>latn_eng_colemak</item>
+    <item>latn_eng_dvorak</item>
+    <item>latn_eng_qwerty</item>
+    <item>latn_eng_qwertz</item>
```
Julow, line 24: This layout is not English. It's German and is a variation of `latn_deu_qwertz`. `qwertz` should be renamed to `latn_deu_qwertz` and `qwertz_de` to `latn_deu_qwertz_2`.

res/values/arrays.xml
```diff
@@ -2,74 +2,74 @@
...
-    <item>bone</item>
-    <item>jcuken_ua</item>
-    <item>bangla</item>
-    <item>devanagari_1</item>
-    <item>devanagari_2</item>
-    <item>he_il_1452_1</item>
-    <item>he_il_1452_2</item>
-    <item>ar_pc</item>
-    <item>ar_alt</item>
+    <item>arab_ara_ibm</item>
```
Julow, line 5: The app being designed for developers, I'd prefer to see programming layouts at the beginning of the list: QWERTY, Colemak, Dvorak.

Julow (commented):
I love the idea! Thanks for the research and implementation work!

res/values/arrays.xml
```diff
@@ -2,74 +2,74 @@
...
+    <item>beng_ben_rh</item>
+    <item>cyrl_bul_yaverti</item>
+    <item>cyrl_ukr_jcuken</item>
+    <item>cyrl_rus_jcuken</item>
+    <item>deva_hin_inscript</item>
+    <item>deva_hin_lv</item>
+    <item>grek_ell_qwerty</item>
+    <item>hebr_heb_si_1452_1</item>
+    <item>hebr_heb_si_1452_2</item>
+    <item>kore_kor_ks_x_5002</item>
```
Julow, line 16: That's `kore_kor_dubeolsik_ks_x_5002` ?

Julow (commented):
A few things:

- There are some references to `R.xml.qwerty` in the code, they should be updated.

- I caused some conflicts by adding a entry to `method.xml` that I fixed by pushing to your branch.

res/values/arrays.xml
GrimPixel: In the orthography of those languages, language names are not capitalised.

res/values/arrays.xml
GrimPixel: Thank you so much. I always used its alternative spelling. Arabic should be diacritised?

res/values/arrays.xml
GrimPixel: Yes. That was a mistake.

res/values/arrays.xml
GrimPixel: Hmm, my original idea was, äöüß are supposed to be on the key0s according to DIN 2137-1 for `qwertz`. It can be perceived as a generic `qwertz` with German support.

Now I think about it, DIN 2137-1 is not the factor that determines whether the keyboard is German or not.

I am cautious about simply naming it `2` because it is not intuitive which one is the number. Maybe adding more letters and diacritics to make it a real generic `qwertz`?

res/values/arrays.xml
GrimPixel: The problem in `method.xml` is `qwertz_hu`. I have found the other `R.xml.qwerty`, too.

GrimPixel:
I have also changed `kore` to `hang`, determined by its input instead of its final output.

res/values/arrays.xml
GrimPixel: It's about sorting with which order. Currently it is a mixture: ISO 15924 first, name second. It needs to be changed for the complexity.

Maybe https://github.com/Julow/Unexpected-Keyboard/issues/325 will help: in addition to writing systems, there is a “[programming]” group including them.

res/values/arrays.xml
Julow: Most of these languages capitalize the first word of a sentence. Also for consistency, I'd like to see every items capitalized.

res/values/arrays.xml
Julow: That's a good point, thanks! Knowing that, `latn_eng_qwertz` works for me.

Julow: Can you write the name of the standards you researched into the `.xml` file ? As comments like this:
```
<!-- Defined in standard DIN 2137-1 -->
```
(some layouts are not defined in standards)

res/values/arrays.xml
Julow: https://github.com/Julow/Unexpected-Keyboard/issues/325 is a good solution but the groups still have to be sorted anyway.

GrimPixel: Yes, it is. _programming_ will be at the top of the first menu together with writing systems, languages will be on the second menu and layouts for that language will be on the third. Then all layouts can be sorted by their codes.

res/values/arrays.xml
GrimPixel: Yes, I can do it.

res/values/arrays.xml
GrimPixel: Hmm, there is a bit of different understanding around “standarisation”.
Probably because of Wikimedia Foundation's approach, in many places where autonyms of the language are used, other people capitalise them all, too.
If they want to make all languages of a writing system follow the same orthography rule, then it brings about another question: capitalise all or uncapitalise all. Outside the influence area of England, Germany, Netherlands, this rule of capitalising language names does not apply.
In my opinion, that Wikimedia Foundation's approach is a kind of anglicisation. So I chose not to force the capitalisation, or uncapitalisation, on other languages.

res/values/arrays.xml
Julow: That's useful, thanks! Don't put too much time on this, I just want the information you already have researched for future reference.

res/values/arrays.xml
Julow: Thanks. In the meantime, you can separate groups of items by adding empty lines. That way, items inside a group can be sorted lexicographically while groups are ordered as we want.

I understand that the list is hard to re-order confidently. I plan to fix https://github.com/Julow/Unexpected-Keyboard/issues/329 at some point to solve the issue of the two arrays going out of sync.

res/values/arrays.xml
GrimPixel: Nope. The definition of a sentence is not like this. It is a noun phrase indeed.

res/values/arrays.xml
GrimPixel: Alright. I will focus on other aspects.

GrimPixel:
> A few things:
> 
>     * There are some references to `R.xml.qwerty` in the code, they should be updated.

I moved that part to a markdown file.

res/xml/define.md
```diff
@@ -0,0 +1,23 @@
+# How to define a layout
+
+A layout is made of keys arranged into rows. The width of keys can be defined with the 'width' attribute; a space can be added to the left side of a key with the 'shift' attribute.
+
+'k0' assigns the symbol in the middle of the key. Other symbols are assigned with 'k1', 'k2', etc.:
+┌─────┐
+│8 1 2│
```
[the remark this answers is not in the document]
GrimPixel: This one is not arbitrary. It is clockwise and easy to memorise. The former one is composed of two zigzags. For the typo-bug, can you find one out?

res/values/arrays.xml
Julow: If you have the time to implement it, that would be wonderful! Please keep this in a separate PR.

res/xml/define.md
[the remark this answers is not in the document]
Julow: I'd appreciate if you could move that to a separate PR.

Julow:
Thanks for all the ideas but I'd appreciate if you could keep the PR focused on renaming the layouts only.

I will not be able to review a PR that big in a reasonable amount of time and it would be a shame to block so much good stuff. I'd like reasons for individual changes too.

res/xml/define.md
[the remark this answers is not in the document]
GrimPixel: Good. I'll do the new changes later.

PR #333, opened 2023-04-15
Updated pl (Polish) translation

ChasmSolacer:


Julow (approved):
Many thanks!

By the way, what's your opinion on the new description ?

ChasmSolacer:
I think the images can describe this keyboard the best way. It's good for the description to be also kept lightweight

Julow:
Noted!

PR #347, opened 2023-04-23
Standarised layout names

GrimPixel:
Current solutions:
* Only script codes, no language codes or names;
* Only when variants exist, distinguish their codes and names with the country/territory;
* Script names are written in English;


res/values/arrays.xml
```diff
@@ -2,76 +2,76 @@
...
-    <item>QWERTY (Vietnamese)</item>
-    <item>ЙЦУКЕН (Русский)</item>
-    <item>QWERTZ</item>
-    <item>QWERTZ (Czech)</item>
-    <item>QWERTZ (Deutsch)</item>
-    <item>QWERTZ (Hungarian)</item>
-    <item>QWERTZ (Slovak)</item>
-    <item>Bulgarian (Phonetic Traditional)</item>
-    <item>Dvorak</item>
+    <item>QWERTY [United States of America]</item>
```
Julow, line 42: This is more commonly called "US". Shorter is better.

res/values/arrays.xml
```diff
@@ -2,76 +2,76 @@
...
-    <item>বাংলা</item>
-    <item>देवनागरी (हिंदी)-1</item>
-    <item>देवनागरी (हिंदी)-2</item>
-    <item>Hebrew SI-1452-1</item>
-    <item>Hebrew SI-1452-2</item>
-    <item>Arabic PC</item>
-    <item>Arabic ALT</item>
+    <item>Dvorak</item>
+    <item>IBM Arabic</item>
+    <item>IBM Arabic (Mostafa Ayesh)</item>
```
Julow, line 46: I don't like naming layouts after their author. This will encourage people to submit too many layouts and discourage people from using them.
This layout is usable for everyone, not just its author.

`PC` and `ALT` were good enough names.

res/values/arrays.xml
```diff
@@ -2,76 +2,76 @@
...
+    <item>Devanagari (Luke Videckis)</item>
+    <item>QWERTY [Hellas]</item>
+    <item>Hebrew</item>
+    <item>Hebrew 2</item>
+    <item>두벌식 [Republic of Korea]</item>
+    <item>AZERTY [France]</item>
+    <item>Bone</item>
+    <item>Neo 2</item>
+    <item>QWERTY</item>
+    <item>QWERTY [Brazil]</item>
```
Julow, line 62: I agree with mentioning only countries. I'd prefer parenthesis, which are more readable. We don't need the two different forms with author names removed.

res/values/arrays.xml
```diff
@@ -2,76 +2,76 @@
...
+    <item>Bengali (Md Rasel Hossain)</item>
+    <item>ЙЦУКЕН [Russia]</item>
+    <item>ЙЦУКЕН [Ukraine]</item>
+    <item>ЯВЕРТЪ</item>
+    <item>InScript Devanagari</item>
+    <item>Devanagari (Luke Videckis)</item>
+    <item>QWERTY [Hellas]</item>
+    <item>Hebrew</item>
+    <item>Hebrew 2</item>
+    <item>두벌식 [Republic of Korea]</item>
```
Julow, line 57: I don't like the inconsistently translated country names. Why is that ? It seems that you already made all the translations in the previous PR.

srcs/juloo.keyboard2/Config.java
```diff
@@ -342,39 +342,39 @@ public KeyboardData layout_of_string(Resources res, String name)
     {
       case "system": case "none": return null;
```
Julow, line 343: There are references to `R.xml.qwerty`:
```
srcs/juloo.keyboard2/Config.java:340:    int id = R.xml.qwerty; // The config might store an invalid layout, don't crash
srcs/juloo.keyboard2/Keyboard2.java:108:      _localeTextLayout = KeyboardData.load(getResources(), R.xml.qwerty);
```

Julow (commented):
Thanks!

res/values/arrays.xml
GrimPixel: Or “USA”, because strictly speaking, “United States” can also indicate Mexico as it is called “United Mexican States”.

res/values/arrays.xml
GrimPixel: This involves the name of “Personal Computer”. It was initially a brand name, then generalised and the word is uncapitalised as “personal computer”.
“PC” is good to use. “ALT” doesn't tell what is changed from the original design. There are layouts based on another derivative, like Colemak-DH ISO and Colemak-DH Ortholinear, so “ALT” isn't enough. I wish the contributors could come up with a name for their own layout.

res/values/arrays.xml
GrimPixel: Yes. I think if the contributors don't name their layouts themselves, we have to do the work ourselves.

res/values/arrays.xml
GrimPixel: Which one did you think is inconsistent? The naming rule is to follow the shortened form of the formal name and if there are more than one country/territory having the same shortened form or causing confusions with a geographical term, use full name.
Here, when “United States of America” is written as “USA”, “Republic of Korea” needs to be renamed “Ro Korea”.

srcs/juloo.keyboard2/Config.java
GrimPixel: Thanks. I renamed them.

res/values/arrays.xml
Julow: "US" is the most used. Let's keep it simple.

res/values/arrays.xml
Julow: This PR is not the right place to change this.

res/values/arrays.xml
Julow: Some of the names are spelled in english while other are not.

I'd also prefer smaller names.

> or causing confusions with a geographical term

I don't care with this one. I want layouts to have slightly different names, that's it.

GrimPixel: Could you explain what that “spelled in english” mean? Did you mean name not abbreviated?
If you don't care abou that geographical term, then “United States” could be called “America” instead and cause confusion with the Latin American variaet.

res/values/arrays.xml
GrimPixel: Contributors need to come up with a name then. They are not supposed to create a layout without naming it properly and simply call them “ALT”, “ALT 2”, “ALT 3”, and so on.

res/values/arrays.xml
Julow: This is way too complicated. "QWERTY (US)" is the most common name.

res/values/arrays.xml
GrimPixel: Then I don't find it meaningful to standarise all the names, as this leaves anything to be in a mess again.

res/values/arrays.xml
GrimPixel: “Hellas” is an English word. It is the native name of Greece and is the noun form of “Hellenic” in the formal name of Greece “Hellenic Republic”.

Julow: I liked translated layout names and countries better (though it was language names before, which I don't prefer). I think this might be what non-english users expect.

Julow (commented):
I still like this change, though the contribution process has been chaotic. 

Please add as many commits as you want. They are useful for me to remember all that happened. The force-pushes and the closing of the older PRs don't help me have a clear idea.

Please focus on one change at a time. You are making and undoing too many things in between my reviews. Good things have been removed and points we are disagreeing on are blocking everything.

Julow:
I'm merging most of your changes in https://github.com/Julow/Unexpected-Keyboard/pull/386
Thanks a lot!

There are a few changes I don't agree with that I didn't take:
- Naming layouts after contributors. I agree the names of some layouts are not good but this is not a solution.
- Changes in the display names of the layouts. The English names are not interesting for non-English speakers and I'd prefer to keep the names chosen by the contributors for now.

Happy to see more of your contributions.

PR #357, opened 2023-05-02
Add Bengali Provat Layout

shmVirus:
I have added Bengali_Provat layout and renamed old Bengali layout to actual name that is Bengali_Provat.

- For key mapping, I considered physical keyboard mapping.
  - [Provat Layout Reference](https://en.wikipedia.org/wiki/Bengali_input_methods#Probhat).
- Kept `~` and `zwj` on top right instead of top left.
  - inspired from 65% keyboards, as these also moves `~` to top right.
- I have also added `zwj` and `zwnj`, these two gives flexibility for ligatures with different scripts.
  - [ZWJ Reference](https://en.wikipedia.org/wiki/Zero-width_joiner).
  - [ZWNJ Reference](https://en.wikipedia.org/wiki/Zero-width_non-joiner).

Finally here's a visualization of added layout.
UnexpectedKeyboard__BengaliProvatLayout


res/xml/bengali_provat.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<keyboard>
+	<row>
+		<key 				key0="দ"	key1="ধ"	key2="১"	key3="!"				key5="esc"	/>
+		<key 				key0="ূ"		key1="ঊ"	key2="২"	key3="‌‌‌‌@"							/>
+		<key 				key0="ী"		key1="ঈ"	key2="৩"	key3="‌‌‌‌#"							/>
```
Julow, line 6: These can be properly escaped:
```suggestion
		<key 				key0="ূ"		key1="ঊ"	key2="২"	key3="\@"							/>
		<key 				key0="ী"		key1="ঈ"	key2="৩"	key3="\#"							/>
```

res/xml/bengali_provat.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<keyboard>
+	<row>
+		<key 				key0="দ"	key1="ধ"	key2="১"	key3="!"				key5="esc"	/>
```
Julow, line 4: Symbols close to the edges of the keyboard can be hard to reach. What do you think of moving `esc` to the Shift key below and `ধ` to `key4` ?

srcs/juloo.keyboard2/KeyValue.java
```diff
@@ -418,6 +418,9 @@ public static KeyValue getKeyByName(String name)
       case "ole_placeholder": return placeholderKey(Placeholder.OLE);
       case "meteg": return charKey("\u05DE\u05BD", '\u05BD', 0); // or siluq or sof-pasuq
       case "meteg_placeholder": return placeholderKey(Placeholder.METEG);
+      /* intending/preventing ligature - supported by many scripts*/
+      case "zwj": return charKey("zwj", '\u200D', 0); // zero-width joiner (provides ligature)
+      case "zwnj": return charKey("zwnj", '\u200C', 0); // zero-width non joiner (prevents unintended ligature)
```
Julow, line 423: `zwj` and `zwnj` could have nicer symbols. They are added to `srcs/special_font` as SVG files and rebuilt with `rebuild_special_font`, which requires `fontforge` installed. The char code is specified by the file name, see how `modifierKey` has a version that takes a `int symbol` instead of a `String symbol`.

I can take care of this later if you don't have the time.

Julow (commented):
Thanks!

res/xml/bengali_provat.xml
shmVirus: Thanks. On my editor indentation was okat, but don't know what happen after commit.

res/xml/bengali_provat.xml
shmVirus: I know, but I think it's better to keep identical as physical layout.

@bdeshi has a suggestion to fix this issue. At the moment, shift key is only working for latin alphbets. To make shift key work in onthet languages, something like this can be a good idea:
# layout definition
...
<key key0="দ" shifted0="ধ"/>
then update the apply_shift() function so that it reads the corresponding shifted# character for pressed key# from the layout file.

srcs/juloo.keyboard2/KeyValue.java
shmVirus: If you want to replace with symbols, you can. But I have seen layouts that have these characters, labeled with zwj and zwnj.

res/xml/bengali_provat.xml
Julow: This is an interesting idea but it might make layouts more complex. The current way would be to add new rules into `map_char_shift`, like this:
```
  case 'দ': return 'ধ';
```

Would that work for you ?

res/values/arrays.xml
```diff
@@ -27,7 +27,8 @@
     <item>neo2</item>
     <item>bone</item>
     <item>jcuken_ua</item>
-    <item>bangla</item>
+    <item>bengali_national</item>
+    <item>bengali_provat</item>
```
Julow, line 31: Are these the unambiguous names for these layouts ?

Layouts are being renamed internally in https://github.com/Julow/Unexpected-Keyboard/pull/347 and good names were not yet found for the first one. Help needed.

Julow (approved):
I'll merge this just before the next release. In the meantime, feel free to make more changes. Thanks!

res/xml/bengali_provat.xml
shmVirus: If shift is applied to a specific language instead of a language, then it won't work. Different layouts map keys differently; for example, in Provat, 'ম' is mapped with 'ঙ' but in Bijoy, 'ম' is mapped with 'শ'.

res/values/arrays.xml
shmVirus: Yes, these Bengali names are unambiguous; these names are provided by layout designers.

If you are talking about layouts above Bengali, I can confirm both `neo2` and `jcuken_ua` are unambiguous but couln't find anything about `bone` layout. References: [neo2](https://en.wikipedia.org/wiki/File:Neo2_keyboard_layout.svg) and [jcuken_ua](https://en.wikipedia.org/wiki/JCUKEN).

Julow: Thanks!

(I was asking for the two bengali layouts, [bone](https://web.archive.org/web/20180721192908/http://wiki.neo-layout.org/wiki/Bone) is an older variant of neo2)

res/xml/bengali_provat.xml
Julow: Interesting!

The `shifted0` is not bad but hard to implement currently. I'll try to come up with a solution soon.

shmVirus: Thanks!

Julow: I finally found some time and implemented something that might help (22d407c):
```xml
<keyboard>
  <modmap>
    <shift a="ম" b="ঙ"/>
  </modmap>

  <!-- The <row>s here -->
</keyboard>
```

This is brand new and might contains bugs.

Julow:
Merging, thanks!

PR #376, opened 2023-06-07
Explicitly use utf-8 encoding while formatting translations on Windows.

9-2-1:
When running sync_translations.py on Windows, an error will show because it use gbk encoding by default. Using utf-8 encoding explicitly can fix it.


Julow (approved):
Many thanks!

Issue #349, opened 2023-04-27
Decluttering Czech Keyboard


frimdo:
As a native czech speaker and user of czech qwertz keyboard, I find the `qwertz_cs.xml` layout cluttered.
There are many accents unused in czech language (ß,ô,ľ,ë,ï,ö,ü,...), greek letters and unneeded math symbols (×,⋅).

Here is image characters that I find useless (as they can be enabled using `Add keys to the keyboard` feature.
![obrazek](https://user-images.githubusercontent.com/10960486/234841563-50b41b8c-1848-4bbe-ae87-2ad5cd1519c0.png)

Also, placing of the letters is unusual compared to czech qwertz keyboard (as seen from e.g. wiki):
![wiki_qwertz](https://upload.wikimedia.org/wikipedia/commons/thumb/9/98/Windows_Czech_keyboard_layout.svg/800px-Windows_Czech_keyboard_layout.svg.png)

Here is image of proposed layout:
![obrazek](https://user-images.githubusercontent.com/10960486/234842653-14b62966-9a02-4e29-97ef-3198e3e432cf.png)
Ans here is proposed layout:

```
<?xml version="1.0" encoding="utf-8"?>
<keyboard>
  <row>
    <key                 key0="q"         key1="1"         key2="esc"       key3="+"          key4="~"      />
    <key                 key0="w"         key1="2"                          key3="|"          key4="\\"     />
    <key                 key0="e"         key1="3"         key2="é"         key3="^"          key4="ě"      />
    <key                 key0="r"         key1="4"         key2="ŕ"                           key4="ř"      />
    <key                 key0="t"         key1="5"                          key3="°"          key4="ť"      />
    <key                 key0="z"         key1="6"                                            key4="ž"      />
    <key                 key0="u"         key1="7"         key2="ú"         key3="\`"         key4="ů"      />
    <key                 key0="i"         key1="8"         key2="í"         key3="("                        />
    <key                 key0="o"         key1="9"         key2="ó"         key3=")"                        />
    <key                 key0="p"         key1="0"         key2="/"         key3="="          key4="%"      />
  </row>
  <row>
    <key shift="0.5"     key0="a"         key1="tab"       key2="á"         key3=";"                        />
    <key                 key0="s"                                                             key4="š"      />
    <key                 key0="d"                                                             key4="ď"      />
    <key                 key0="f"                                           key3="["                        />
    <key                 key0="g"                                           key3="]"                        />
[...]

PR #380, opened 2023-06-10
czech minimal layout added, default czech keyboard changed to minimal

frimdo:
Decluttered czech keyboard introduced in #349. 
Proposed minimal keyboard is set as default czech keyboard.


res/xml/method.xml
```diff
@@ -4,7 +4,7 @@
   <subtype android:label="%s" android:languageTag="be" android:imeSubtypeLocale="be_BY" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=ru_jcuken,extra_keys=ґ|є|і|ї|ў"/>
   <subtype android:label="%s" android:languageTag="bg" android:imeSubtypeLocale="bg_BG" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=bgph1,extra_keys=€"/>
   <subtype android:label="%s" android:languageTag="bn" android:imeSubtypeLocale="bn_BD" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwerty,extra_keys=৳"/>
-  <subtype android:label="%s" android:languageTag="cs" android:imeSubtypeLocale="cs_CZ" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz_cs,extra_keys=a|accent_cedille|accent_circonflexe"/>
+  <subtype android:label="%s" android:languageTag="cs" android:imeSubtypeLocale="cs_CZ" android:imeSubtypeMode="keyboard" android:isAsciiCapable="true" android:imeSubtypeExtraValue="default_layout=qwertz_cs_minimal"/>
```
Julow, line 7: In the issue you mention the Slovak language. Could you add it as a subtype ?

res/xml/qwertz_cs_minimal.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Julow, line 1: This file is encoded as `latin1` but must be UTF-8 to be parsed by android's build tools. I re-encoded it and fixed the escaping of some keys on this branch: https://github.com/Julow/Unexpected-Keyboard/tree/380-reencoded

res/xml/method.xml
frimdo: I have no problem with creating a minimal slovak keyboard, but as I am not a slovak speaker, I can not test by using it for a month (as I did with the czech layout).

The only problem I have is, that I don't know what a subtype is.

res/xml/qwertz_cs_minimal.xml
frimdo: Oops, the encoding is a silly mistake, sorry. The missing escaping is caused by the `custom layout` feature. Adding the layout using the `custom layout` feature requires characters **not** to be escaped. Should I open an issue for this?

res/xml/method.xml
Julow: No need to make a Slovak layout! I'm thinking of using qwerty or qwertz with the appropriate `extra_keys`.
I asked this just in case you knew, don't worry about that :)

A subtype is a line of `method.xml` similar to the one you modified. It specifies the preferred layout for a language and the keys that should be added to other layouts automatically.

res/xml/qwertz_cs_minimal.xml
Julow: Interesting ! Yes, an issue would be useful.

res/xml/method.xml
frimdo: Oh, ok, I already looked into it a little bit, and it seems that [slovak](https://en.wikipedia.org/wiki/File:KB_Slovak.svg) and [czech](https://cs.wikipedia.org/wiki/Soubor:Windows_Czech_keyboard_layout.svg) layouts are different to such extend, that slovak would deserve its own layout.

res/xml/qwertz_cs_minimal.xml
frimdo: The issue is opened #381

Julow: Thanks!

res/xml/method.xml
Julow: Thanks!

res/xml/qwertz_cs_minimal.xml
Julow: Thanks! Do you wish that I push the re-encoding commit to your branch directly ?

frimdo: Yes please :)

Julow: There was some more encoding problems, so I pushed a second commit. I also made the layout available in the option.

res/xml/qwertz_cs_minimal.xml
```diff
@@ -0,0 +1,37 @@
...
+<keyboard>
+  <row>
+    <key                 key0="q"         key1="1"         key2="esc"       key3="+"          key4="~"      />
```
Julow, line 4: Keys on the edge of the screen (`key1` and `key3` here) are harder to type. It would be best to move these elsewhere. Some keys from the first row could be pushed down to the second row as there are some free spots.

On qwerty, this was solved by moving the digit close to the edge of the screen to the opposite position (inconsistent with the other digits). The escape key is on `key4`.

res/xml/qwertz_cs_minimal.xml
```diff
@@ -0,0 +1,37 @@
...
+    <key                 key0="w"         key1="2"                          key3="|"          key4="\"     />
+    <key                 key0="e"         key1="3"         key2="�"         key3="^"          key4="�"      />
+    <key                 key0="r"         key1="4"         key2="�"                           key4="�"      />
+    <key                 key0="t"         key1="5"                          key3="�"          key4="�"      />
+    <key                 key0="z"         key1="6"                                            key4="�"      />
+    <key                 key0="u"         key1="7"         key2="�"         key3="`"         key4="�"      />
+    <key                 key0="i"         key1="8"         key2="�"         key3="("                        />
+    <key                 key0="o"         key1="9"         key2="�"         key3=")"                        />
+    <key                 key0="p"         key1="0"         key2="/"         key3="="          key4="%"      />
```
Julow, line 13: More keys on the edge of the screen.

frimdo:
Solved keys on the edge of the screen and also switched buttons `<` and `>`.
I am a little bit worried about the encoding issues, as after your fix, the letters are not displayed correctly, not even in github:
obrazek
should be 
```xml
key4="~"/>
key4="\\"/>
key4="ě"/>
key4="ř"/>
key4="ť"/>
key4="ž"/>
key4="ů"/>
key4=")"/>
key4="%"/>
```

frimdo:
An attempt to solved the issue with encoding is provided in [frimdo:frimdo-encoding-utf8](https://github.com/frimdo/Unexpected-Keyboard/tree/frimdo-encoding-utf8)

Julow:
I did not realize the language used `ů` and broke the layout. `frimdo-encoding-utf8` works fine with a small fix:

```diff
diff --git a/res/xml/qwertz_cs_minimal.xml b/res/xml/qwertz_cs_minimal.xml
index 1f9bdc8..498335b 100644
--- a/res/xml/qwertz_cs_minimal.xml
+++ b/res/xml/qwertz_cs_minimal.xml
@@ -29,7 +29,7 @@
     <key key0="x" key3="\#"/>
     <key key0="c" key3="&amp;" key4="č"/>
     <key key0="v" key3="\@"/>
-    <key key0="b" key1="&lt;" key2="&gt" key3="{" key4="}"/>
+    <key key0="b" key1="&lt;" key2="&gt;" key3="{" key4="}"/>
     <key key0="n" key1="\?" key2="." key3="," key4="ň"/>
     <key key0="m" key1=":" key2="*" key3="-" key4="_"/>
     <key width="1.5" key0="backspace" key2="delete"/>
```

If you merge this branch and apply the fix, this should be ready to be merged !

frimdo:
The fix is added.

Julow:
No, I meant to merge `frimdo-encoding-utf8` into the branch of this PR (`master`). (if you want)

But now, there are also conflicts. I fixed them and pushed directly to your branch because the conflicts are due to unrelated changes and were complicated.

Don't worry about the complicated history that would result from that, I'll squash it.

frimdo:
To be honest, I am a bit lost now. Should I do something now?

Julow:
Nothing to do about the branches, sorry for the confusion. 

There's huge conflicts, as I made big changes to layouts recently. I'll fix the conflicts just before merging, which I'll do soon.

I intent to rename the layouts though:
- `qwertz_cs` -> `qwertz_cs_multifunctional` (as named by @Validbit )
- `qwertz_cs_minimal` → `qwertz_cs`

Julow:
@Validbit Not needed, I'll merge @frimdo's layout directly.

Julow:
Thanks! I'll merge now.

frimdo:
Thanks! I'm happy to help and happy to have my first open-source contribution merged!

Issue #204, opened 2022-09-24
Feature request: Compose key


Julow:
Related issues: https://github.com/Julow/Unexpected-Keyboard/issues/30 https://github.com/Julow/Unexpected-Keyboard/issues/171 https://github.com/Julow/Unexpected-Keyboard/issues/191 https://github.com/Julow/Unexpected-Keyboard/issues/164

The [compose key](https://en.wikipedia.org/wiki/Compose_key) is a good way to type rarely needed characters. For example, `Compose, -, >` types `→`, `Compose, ', e` types `é`.

Key combinations are generally intuitive but the keyboard has the opportunity to show the possible combinations during typing, making them easy to discover.

This feature is called `Multi_key` in XKB and the list of [combinations is huge](https://cgit.freedesktop.org/xorg/lib/libX11/plain/nls/en_US.UTF-8/Compose.pre)

Julow:
@Spike-from-NH suggested to use Hacker's Keyboard's compose key sequences in https://github.com/Julow/Unexpected-Keyboard/issues/523

parkerlreed:
What sets the icon for the compose key? I can't tell what it's supposed to be. It's just a circle with a vertical line which looks weird in the middle of the arrow keys.

Julow:
I started from the [ISO keyboard symbol](https://en.wikipedia.org/wiki/File:ISO_7000_-_Ref-No_2021.svg) and removed the square part because the symbol was too wide. I don't have other suggestions and no one complained until now.

PR #383, opened 2023-06-20
Split (decluttered) Czech layouts + Updated translation

Validbit:
In response to https://github.com/Julow/Unexpected-Keyboard/issues/349, https://github.com/Julow/Unexpected-Keyboard/pull/380 and additional feedback, changes to Czech layout were made and minimal (default) + aosp/requested layout was created. New Czech strings,xml were translated and two Fn shortcuts added.


Julow (commented):
Many thanks for the translations !

I was not really happy with the idea of having 2 slightly different layouts but 4 is a stretch.
I would prefer if you could collaborate with @frimdo on https://github.com/Julow/Unexpected-Keyboard/pull/380 and together converge to 1, ideally, or 2 layouts.

Validbit:
It'd be 3, if accepted, but my main point is....
_Mobile virtual keyboard will be always greatly limited compared to a normal keyboard, but this one proved it might as well not be (without sacrificing much speed/accuracy). _

Some **physical** layouts (Czech default) utilize right Alt (Alt Gr) to write symbols of common specialized use cases (that, when needed, can be typed fast without ever switching). 

Here (without multiple layouts) it's still good, but...
as some of us use our phones more and more for organisation, writing (email/text/document/blog/KB), school, programming (learning, queries) and mass communication on-the-go, there are simply more use-cases that "ask for" effortless typing.

Example uses of quick-access special symbols:
- communication/document shortcuts (Δ = change, ø = diameter of pizza, ~ = estimation, § = reference/law)
- text-expansion (custom)
- hobby/edu communication ( ♭, ÷)
- passwords
- systematic writing (blog, KB, wiki, ..) (×. `, •)

- school (rest of symbols in current Multifunctional (- arrows))
- communication itself (multiple languages, understanding without resorting to english)
- language learning (multiple languages)

Fn is a solution but has only one possibility per character (e.g. Fn+P (pi) cannot be assigned)

One of **possibilities** I thought of would be to have 
- (existing) single language layout toggle 
(one/locale) [space top]
- variant toggle (added layout of symbols[^2])
(e.g. Basic, Compatibility*, Universal, (user)) [space bottom]

(imo) **to get rid of future requests** for different layouts per language. (it still is far from universal (effortless typing) but it may as well be a solution)

If you (group) really desire one without options, we can of course simplify to the requested layout but (apart from some people not finding the keyboard useful enough) I personally would like to see the keyboard utilized to it's full potential while making it easy for newcomers to set it up. 

(The editor is also viable solution - could be better utilised if there was and official guiding in-app (for others) how to use it. maybe later implement the variants functionality to switch/share them better.)

*layout(s) similar to other keyboards they're used to (AOSP, Swiftkey, ...) 
 
[^2]: Static (user-layouts) would be prefferable (and easier to make) since in automatic placement pose difficulties to setup rules (e.g. pair: ˋ((key2="[" && key3="]")||((key4="[" && key3="]"))ˋ). Without them the placement of certain symbols (mostly brackets) would be too difficult to get used to, if applied to certain languages)

Julow:
> Some physical layouts (Czech default) utilize right Alt (Alt Gr)

This app uses several dead-keys for this purpose. Adding new deadkeys is better than cluttering layouts.

> Example uses of quick-access special symbols:

These symbols are not specific to the Czech language.
I recommend adding all the ASCII punctuations to latin layouts because the app is initially intended to programmers but I do not recommend cluttered layouts.

Users of other layouts would benefits from these symbols in other places:
the `Fn` modifier, the number board, the "greekmath" board.

> variant toggle (added layout of symbols)

I'm not against such an idea. An other idea for adding a huge number of symbols is the [compose key](https://github.com/Julow/Unexpected-Keyboard/issues/204).

I don't want to get rid of future requests for new layouts. I want to avoid layouts used by few people. Layouts backed by a standard are always welcome.
I wouldn't use the cluttered czech layout and an other person mentionned that too. Let's focus on that.

Julow:
I intent to merge https://github.com/Julow/Unexpected-Keyboard/pull/380 soon as it fixes an immediate problem.

With this in mind, this PR propose to increase the number of layouts for Czech to 4. Two of which have no user.
I think this is not the right way forward. I prefer to focus on the current two layouts (the one currently released and https://github.com/Julow/Unexpected-Keyboard/pull/380), so I'll close this PR.

Translations are very valuable, would you mind opening a PR for them ?

PR #395, opened 2023-07-11
Some changes in characters' positions in persian.xml

RZHSSNZDH:
Changes some characters' places like `!`.

### DIFF
--- a/res/values/layouts.xml
+++ b/res/values/layouts.xml
@@ -50,7 +50,7 @@
     <item>Arabic Alt</item>
     <item>Arabic PC</item>
     <item>Arabic (کوردی)</item>
-    <item>Persian PC</item>
+    <item>Persian Alt</item>
     <item>বাংলা (জাতীয়)</item>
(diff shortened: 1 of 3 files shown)


RZHSSNZDH:
I am not an android app developer, so I do not have installed its tools, so make commands to add the layout does not work and I get error below:
`Makefile:46: *** ANDROID_HOME not set.  Stop.`

Julow (commented):
What do you think of modifying the existing layout instead ? I'd prefer one well-maintained by several people rather than several slightly different layouts.

RZHSSNZDH:
> What do you think of modifying the existing layout instead ? I'd prefer one well-maintained by several people rather than several slightly different layouts.

You are right, but when I clone the project, I could not find `persian.xml` in `/res/xml`. So I created a new persian xml file.

Julow:
Layouts have all been renamed to `<script>_<layout name>_<country>`, Persian is now `/res/xml/arab_pc_ir.xml`.

PR #407, opened 2023-07-22
Update arab_pc_ckb.xml

GoRaN909:
Correct layout name to Kurdish because Kurdish and Arabic are totally two different language.


Julow (approved):
Thanks!

Issue #191, opened 2022-08-24
(Feature Request) A way to write Unicode Characters


YeaminSarder:
Please allow enter Unicode.
you can provide a key that will behave like 'alt codes' but uses unicode.
Let me clearify.
I want to hold a key and enter unicode hex, when i release the key the unicode character should appear.

guobao2333:
Recently I've been troubled by this issue too. Is the delayed support for this feature because Android doesn't support continuous input of alt + codes? Are we forced to use `modmap` for implementation?

If possible, perhaps try using `compose + a key` to lock subsequent key inputs? This could achieve Unicode output or other functions—specifically behaving as `compose + a key + codes`
Alternatively, maintain the original `alt + codes` logic but add a `compose` key press after alt to lock the sequence, then output upon pressing Enter or releasing the lock?

Julow:
I think it's possible to implement this in the keyboard currently (either via `compose` or `alt`). Anyone wants to do it ?

PR #409, opened 2023-07-26
Add dagger, double dagger, section mark, and pilcrow to all QWERTY-like layouts

syskill:
I use the QWERTY (US) layout and I miss these punctuation marks, which can be found in other Android keyboards including the LineageOS stock keyboard US layout. So I've added them to all QWERTY-like keyboard layouts, according to the following rules:

1. If § was already in the layout, then I added † next to it in the same position. Usually this put † on the D key.
2. If < and . were already in the 2nd and 3rd position on the C key, as in the US layout, then I put § and † in the 2nd and 3rd position on the X key.
3. Otherwise, I put § and † on the S and D keys, in the same position.

The result is about as consistent as I think is possible, without moving any existing characters, in that § and † are in almost the same place on almost all of the QWERTY-like layouts (except AZERTY and ЯВЕРТИ).

### DIFF
--- a/res/xml/cyrl_yaverti.xml
+++ b/res/xml/cyrl_yaverti.xml
@@ -8,7 +8,7 @@
     <key key0="т" key2="5" key3="%"/>
     <key key0="ъ" key2="6" key3="^" key4="€"/>
     <key key0="у" key2="7" key3="&amp;" key4="§"/>
-    <key key0="и" key2="8" key3="*"/>
+    <key key0="и" key2="8" key3="*" key4="†"/>
     <key key0="о" key2="9" key3="(" key4=")"/>
     <key key0="п" key2="0" key3="f11_placeholder" key4="f12_placeholder"/>
   </row>
--- a/res/xml/latn_azerty_fr.xml
+++ b/res/xml/latn_azerty_fr.xml
@@ -32,7 +32,7 @@
     <key key0="x"/>
     <key key0="c" key1="accent_cedille" key3="," key4="\?"/>
     <key key0="v" key3=";" key4="."/>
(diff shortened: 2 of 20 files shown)


sdrapha:
As an user, I would rather not have those keys showing up all the time. I would prefer to have them available as optional extra keys instead.

Julow:
Thanks for experimenting ! I also do not like adding more symbols to layouts. People need a wide variety of these rarely used symbols. (recently, https://github.com/Julow/Unexpected-Keyboard/issues/349)
Layouts will feel complicated and ugly to the majority of people that do not use them.
Also, maintaining all the layouts doesn't work. I prefer that layouts are maintained by their users, which results in some layouts with very few symbols and some that are completely full. Users of these layouts might have a different opinion about these new symbols.

For these reasons, I'd prefer to insert these symbols in layouts by other means.
The "Add keys to the keyboard" option would be a better choice. The list can be edited [here](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/ExtraKeysPreference.java#L18).
In the next release, this list will be editable by users but I'd prefer to have used symbols there.

The inconvenient of this approach is that the symbol will be added to arbitrary and changing place. You can place it on the layouts you use with `loc §` and `loc †`, it will appear there only if it's enabled in the settings.

Opinion welcome on agnostic and more efficient ways to type less used symbols: [unicode entry](https://github.com/Julow/Unexpected-Keyboard/issues/191), [compose key](https://github.com/Julow/Unexpected-Keyboard/issues/204).

syskill:
Thank you for the feedback. I'll submit another request using the extra keys feature.

PR #410, opened 2023-07-27
Add dagger, double dagger, section mark, and pilcrow as extra keys on QWERTY-like layouts

syskill:
I use the QWERTY (US) layout and I miss these punctuation marks, which can be found in other Android keyboards including the LineageOS stock keyboard US layout. So I added section mark and dagger as user-selectable extras on the S and X keys (except for the layouts that already had the section mark). Double dagger can be entered as a modified dagger.


Julow:
Thanks! This is not adding pilcrow as the title say. Did you mean to bind it to `Fn+§` ?

Julow:
It was already! I'll merge.

PR #418, opened 2023-08-05
docs: mention Calculator++ (fixes #417)

### LINKED ISSUE #417: Mention similar input app Calculator++
There is a calculator implementing the same idea of sliding to the keys corners to input more complex characters, and it is hands down the best calculator I have ever used. I actually first discovered this UX concept first with Calculator++, and only recently found about UnexpectedKeyboard and loved it instantly!

Likewise, I guess some UK users may also find CPP a very nice companion app, so I suggest to mention it in the readme.

Here is the repo: https://github.com/Bubu/android-calculatorpp

And it's on f-droid: https://f-droid.org/en/packages/org.solovyev.android.calculator/

lrq3000:


Julow (approved):
Thanks!

PR #419, opened 2023-08-05
Updated pl (Polish) translation

ChasmSolacer:


Julow (approved):
That was fast! Thanks a lot!

Julow:
I've added more unplanned strings in ddc4ff1

PR #421, opened 2023-08-06
Updated Czech translation (customary)

Validbit:


res/values-cs/strings.xml
```diff
@@ -43,11 +43,11 @@
   <string name="pref_autocapitalisation_title">Automatická kapitalizace</string>
   <string name="pref_autocapitalisation_summary">Stiskne Shift na začátku věty</string>
   <string name="pref_switch_input_immediate_title">Přepnout na posledně užívanou klávesnici</string>
-  <string name="pref_switch_input_immediate_summary">Jak bude klávesa pro přepnutí klávesnice reagovat</string>
+  <string name="pref_switch_input_immediate_summary">Namísto okna výběru klávesnice (u klávesy pro změnu klávesnice)</string>
```
Julow, line 46: This might be too long on some devices.

Julow (approved):
Many thanks!

Julow:
I've pushed a conflicting change, it's only removing strings: 1eea9e2

Julow:
I've added more strings in ddc4ff1.

res/values-cs/strings.xml
Validbit: I rephrased it so it's even clearer what it does.

Should i revert this line and create new PR?

Julow: You can modify this branch. You also need to pull the master branch. Happy to do it if the conflicts I created are too annoying to fix.

Validbit: I see there were more changes after the translation so I manually updated the file and I'm enclosing it here
[values-cs.zip](https://github.com/Julow/Unexpected-Keyboard/files/12283274/values-cs.zip)

Julow: I've pushed directly to your branch, thanks a lot and sorry for the conflicts.

Validbit: no problem, thank you for the progress :)

PR #422, opened 2023-08-06
Add extra keys

sdrapha:
Add optional extrakeys to the settings [ Superscript, Subscript ]
Add ordinals extrakeys [ ª , º ] to pt-br layout


Julow (approved):
Thanks!

`Sup` can't do ª and º, you can add more combinations in https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L716

sdrapha:
the ordinals ª and º are NOT the superscript forms of regular `a` and `o`, they are differently chars. I just combined two different toppics ( commits ) into a single PR for conveniency, but they are unrelated.
Since I added ª and º ordinals to the optional extra keys specifc to ptbr layout, I can also add them to the general optional extrakeys list, I'll do just that in a moment.

sdrapha:
I could also expand the list of available supercript and subscript letters, but that was not exactly my goal today. I can do that as a follow up addition in another PR later.

sdrapha:
I rebased on top of your latest commits, also added description strings, and the rest of the available superscript and subscript letters.
Also translated the new strings you just created today.

Sorry that the PR became a big one, that was not the original intent. I like to do it little by little in separate steps. Meanwhile, getting done with this at once will avoid further merging conflicts while you work with some other ideas.

Issue #448, opened 2023-08-26
Development Environment


deftkHD:
Hi, I really appriciate your project and your efford. I would like to use this keyboard as my daily-driver, but there are some aspects, that I would really miss and since I have some time, I would like to contribute. What are your reasons not to use Gradle and Android Studio? I did some testing locally and apart from some localization issues, Gradle and Android Studio seem to work just fine. Or let me ask the other way around: Which development enrionment do you recommend to work on this project?

Also I am interested in hearing your oppion about integrating more modern features such as Kotlin code or AndroidX libraries. Is the goal to keep the minimum API level as low as possible or do you think this would be a good idea? As an example I would like to promote the AndroidX Emoji2 library, but you already mentioned it in #384.

Maybe as a background: I already did some work locally and made a new Emoji picker, which uses a combination of the AndroidX library for displaying the emojis and the Keyboard2View to display a custom keyboard layout at the bottom so the wheel gets not reinvented and haptic feedback and everything works automatically. But the changes I made to run the project in Android Studio are too large to contribute these changes, so I was wondering what you recommend.

Julow:
Thanks for experimenting with the code :)

I needed a reproducible build system that works offline and is quick to set up on my machine and found that Gradle didn't allow any of that.
I recognize that the current build system is not ideal and I'm open to Gradle, though I insist that using visual studio shouldn't be a requirement.
I personally develop with `nix-shell ./shell.nix` and `make` and use no LSP.

The API level is kept as low as possible on purpose, as many users are stuck on very old versions in other parts of the world.
By how much would you need to increase it ?

An other goal is to keep the size of the APK at a minimum. It's currently ridiculously low and could be increased for new features but I didn't feel like libraries would be useful in an app that is mostly one custom view.

deftkHD:
Thanks for the feedback :)

I never had any significant issues when setting up Gradle since the GradleWrapper is nearly always shipped together with the source code, but I don't know how this works on NixOS with their package manager. I forked the repo and set up a basic, but (I think) fully functional build system with Gradle. Maybe you could try if it works by invoking `./gradlew assembleDebug`, which will build the debug APK without the need of any LSP at all. The catch however is you need to run Gradle with Java 17, but from what I can see this should be adjustable in the shell.nix file.

[...]

PR #452, opened 2023-08-27
Use Gradle

deftkHD:
As discussed in #448 this pull request uses Gradle as the default build system. This pull request is not yet ready to be merged because of the following issues:

- [x] Documentation needs to be updated
- [ ] F-Droid might have to be noticed about changes in the build system, ~~I'm not sure however~~
- [ ] Update shell.nix (Feedback required. Does it work?)
- [x] Merge upstream changes


sdrapha:
For the github CI, what about caching and retrieving the keystore from github secrets? 
I could not find this feature in the new code. And that is important, so we don't need to keep unninstalling the app to test a new vesion if the signature is always changing.
I'm not that familiar with gradle yet, if you would to point me where in the code you implemented that feature?

sdrapha:
Also, looking at you fork, it seems that gradle based CI to compile the apks takes a bit more time to run, 1m40sec against previously ~27sec

deftkHD:
Oh, I tought of something different when I read that code and I thought this would be handled by Gradle, which is not the case. When I understand correctly, a user has a local keystore which he uploads to GitHub secrets and this keystore is then used to sign the debug apk for each individual user in CI?
As you are not that familar with Gradle yet, I am not that familar with GitHub Workflows so I'll do my best to implement this. Right now Gradle uses a system-wide keystore, but that defeats the purpose. I'm working on this.

sdrapha:
Yep. I think you understood it now. You can save a ascii encoded version of your debug keystore into your github secrets, so then the resulting CI compiled apk will have the same signature as your local compiled apk. every single run, it'll will be consistent. So it's way easier intall an apk on top of an older version.
Those github secrets, or automatically generated(but cached) are unique per user. Each fork has its own, doesn't matter the branch.
Note: Similar aproach could be further developed into creating a CI for automatically generating a Signed RELEASE version of the apk, but that was not my origibal goal, just consistent debug apks would be enough.

deftkHD:
The commit [ba77fec](https://github.com/Julow/Unexpected-Keyboard/pull/452/commits/ba77fec63e1f8a4ddac449ba9c159e6958dfd390) should now fully implement signing of debug and release APKs in GitHub CI and locally with the same keystores. Debug builds should work the same way they did before, but it would be great to have some input on that.

Releases can also be automatically build and signed by adding the secret RELEASE_KEYSTORE (same principle as DEBUG_KEYSTORE). Additionally the secrets RELEASE_KEYSTORE_PASSWORD, RELEASE_KEY_ALIAS and RELEASE_KEY_PASSWORD are required to successfully sign the release APK.
I have to note however that it is theoretically possible to leak the password of the release keystore since it is used as an argument for gpg and thus could be observed by using ps or something like that. The line I talk about is https://github.com/deftkHD/Unexpected-Keyboard/blob/ba77fec63e1f8a4ddac449ba9c159e6958dfd390/.github/workflows/make-apk.yml#L47C1-L47C19. Maybe someone has a better idea to implement this. GitHub has a note about that in their docs: https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow.

> Also, looking at you fork, it seems that gradle based CI to compile the apks takes a bit more time to run

Yes, that is definitly the case, but the CI action you mentioned actually built two APKs, additionally installed FontForge and built the required special font, which was not done by the old build script.
The first two steps take a significant amount of time and while I made building the second (release) APK now optional, FontForge is required or otherwise the font binary has to be pushed to the repo and the Gradle build task for the font has to be made optional. As you can see for the lastest run on my fork, installing FontForge took 13s and building the release APK took 15s. What do you think?

app/src/debug/res/values-cs/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Julow, line 1: What happened to the "product" attribute ?

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
Julow, line 1: Why not merge this gitignore with the main one ?

app/check_layout.py
```diff
@@ -85,13 +85,14 @@ def check_layout(layout):
...
 
 for fname in sys.argv[1:]:
-    if fname in KNOWN_NOT_LAYOUT:
+    layout_id, _ = os.path.splitext(os.path.basename(fname))
+    if layout_id in KNOWN_NOT_LAYOUT:
         continue
     layout = parse_layout(fname)
     if layout == None:
-        print("Not a layout file: %s" % fname)
+        print("Not a layout file: res/xml/%s" % layout_id)
```
Julow, line 93: I'd prefer not to construct fake values.
```suggestion
        print("Not a layout file: %s" % layout_id)
```

app/check_layout.py
```diff
@@ -85,13 +85,14 @@ def check_layout(layout):
...
     else:
-        print("# %s" % fname)
+        print("# res/xml/%s" % layout_id)
```
Julow, line 95:
```suggestion
        print("# %s" % layout_id)
```

app/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
Julow, line 1: If proguard is not used, let's get rid of this file.

.github/workflows/check-layouts.yml
```diff
@@ -12,8 +12,9 @@ jobs:
   check-layouts:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout repo
-      uses: actions/checkout@v3
-    - run: python3 gen_layouts.py
-    - name: Check that the generated layouts.xml is uptodate, run python3 gen_layouts.py otherwise
-      run: git diff --exit-code
+      - name: Checkout repo
```
Julow, line 15: What auto-formatter do you use for yaml ?

app/build.gradle
```diff
@@ -0,0 +1,162 @@
...
+            minifyEnabled false
+            shrinkResources false
+            debuggable true
+            applicationIdSuffix ".debug"
+            signingConfig signingConfigs.debug
+        }
+    }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
```
Julow, line 59: The app used to build on Java 1.7.

app/build.gradle
```diff
@@ -0,0 +1,162 @@
...
+    dependsOn += "buildKeyboardAssets"
+}
+
+tasks.register('buildKeyboardAssets') {
+    dependsOn 'copyKeyboardFont'
+    dependsOn 'buildKeyboardLayoutXml'
+}
+
+tasks.register('buildKeyboardFont') {
+    println "Executing fontforge to build custom font\n"
```
Julow, line 77: This is not caching the built font in the source tree. This was important as the appropriate version of fontforge can't be installed in every distros (for example in f-droid build scripts).

Julow (commented):
Thanks for this huge work! This will help keeping the app maintenable on the long term.

> F-Droid might have to be noticed about changes in the build system, I'm not sure however

I can take care of that.

> Yes, that is definitly the case, but the CI action you mentioned actually built two APKs

Why build two APKs ? Only the debug APK is wanted as the wrongly signed release APK could be confused with an official release build.

> additionally installed FontForge and built the required special font, which was not done by the old build script.

The font file is part of the source and doesn't need to be built in CI.

app/build.gradle
deftkHD: I made this task optional so it can be invoked manually, but just if needed. The built font file is included now.

app/proguard-rules.pro
deftkHD: Agree. Removed.

app/check_layout.py
deftkHD: I implemented it that way so the current check_layout.output file would be compatible. Now just the layout_id gets printed.

.github/workflows/check-layouts.yml
deftkHD: I use Android Studio set to two spaces. Were tabs used before? I am confused.

deftkHD:
> Thanks for this huge work! This will help keeping the app maintenable on the long term.



> Why build two APKs ? Only the debug APK is wanted as the wrongly signed release APK could be confused with an official release build.

The earlier CI builds were just for testing and also separated on a second branch, so right now the CI action will just build a single APK file. This behaviour was quickly disabled and since then I thought of https://github.com/Julow/Unexpected-Keyboard/pull/452#issuecomment-1694731158 and implemented it since I was already on it. This could be used to automatically receive releases signed by a custom key, but it is only optional and the RELEASE_KEYSTORE secret has to be set. If there is no need for this, I can remove it again, but right now it doesn't slow down anything if you don't want to.

> The font file is part of the source and doesn't need to be built in CI.

I removed the CI for that and the automatic trigger in Gradle. I also pushed the resource file.

app/src/debug/res/values-cs/strings.xml
Julow: I see. I'd prefer not to complicate the translation process if possible. This is intended to add the " (Debug)" string in the name of the app, which is shown in several places.
Do you think the app name could be defined programmatically instead ? eg. define `app_name`, which is translated as usual and `debug_suffix`, which is either "" or " (Debug)" and set the app name to the concatenation of the two.

.github/workflows/check-layouts.yml
Julow: I did not enforce any style to the yml files before and was just curious.
The style you set is fine.

CONTRIBUTING.md
```diff
@@ -4,64 +4,67 @@ Thanks for contributing :)
...
-Fortunately, there's not many dependencies:
-- OpenJDK 8
+Fortunately, there are not many dependencies:
+- OpenJDK 17
+- Python 3
 - Android SDK: build tools (minimum `28.0.1`), platform `30`
-- Make sure to have the `$ANDROID_HOME` environment variable set.
+
+If you don't use Android Studio, you probably have to inform Gradle about the location of your Android SDK.
+To do this, create the file `local.properties` in the directory of this repository and paste `sdk.dir=<location_of_android_home>` e.g. `sdk.dir=/home/user/Android/Sdk`. Android Studio does this automatically.
```
Julow, line 15: Is here an environment variable to accomplish this ? This would make the `shell.nix` easier to write.

Otherwise, the file could be created with a commented-out `sdk.dir` line to ease setup ?

app/build.gradle
```diff
@@ -0,0 +1,147 @@
...
+        println "$bos"
+    }
+    copy {
+        from "$buildDir/special_font.ttf"
+        into "src/main/assets"
+    }
+}
+
+tasks.register('buildKeyboardLayoutXml') {
+    dependsOn 'checkKeyboardLayoutXml'
```
Julow, line 93: These two tasks don't depend on each other.

`gen_layouts.py` generates the `layouts.xml` file, which contains lists of layouts that are used in the settings or for lookups. Perhaps the task should be named `genLayoutsList` ?

`check_layout.py` checks some optional properties about layouts to ease reviewing. Perhaps it can be renamed into "lint" ?

app/build.gradle
```diff
@@ -0,0 +1,147 @@
...
+            workingDir = projectDir
+            commandLine("python", "check_layout.py", *layouts)
+            standardOutput = bos
+        }
+
+        new File(projectDir, "../check_layout.output").write(bos.toString())
+    }
+}
+
+tasks.register('generateTranslationTemplates') {
```
Julow, line 121: `syncTranslations` ?

Julow (commented):
Could the `app` directory be removed ? I don't like that scripts are inside the `app` directory and have to struggle with their working directory. I can't think of a purpose and the extra level is annoying when navigating the code.

> This could be used to automatically receive releases signed by a custom key, but it is only optional and the RELEASE_KEYSTORE secret has to be set. If there is no need for this.

I do not wish to automatically build release builds so I think this can be removed, less code is better in github actions.

app/build.gradle
deftkHD: Yes, I'll rename it

CONTRIBUTING.md
deftkHD: I realized the Android Gradle Plugin respects the ANDROID_HOME environment variable, so the file is just required when this variable is not set. I'll update the documentation, but since I don't use NixOS and am not able to properly test it, it would be great if you could do that.

.github/workflows/check-layouts.yml
deftkHD: Okay, that's good

app/src/debug/res/values-cs/strings.xml
deftkHD: It would be possible to define two different app names per strings.xml file, e.g. `app_name_debug` and `app_name_release` and Gradle can replace `app_name` with one of those two depending on the current build type. From my testing this didn't work with one resource just named `app_name`, so they could be named `app_name_debug` and `app_name_release` or `app_name_default` or something like that. This would be easier I think.

deftkHD:
> Could the app directory be removed ? I don't like that scripts are inside the app directory and have to struggle with their working directory. I can't think of a purpose and the extra level is annoying when navigating the code.

I have not seen any Android Gradle project so far without that project structure. I can definitely move the scripts into the root directory of the project, but I can't give any promises about the Gradle submodule.

> I do not wish to automatically build release builds

Okay, then I misunderstood. I removed the code.

app/build.gradle
deftkHD: There are two predefined Gradle tasks named `test` and `lint`. Right now nothing depends on anything regarding our custom tasks, but should the `lintKeyboardLayouts` or `genLayoutsList` tasks get triggered when e.g. running the already existing `test` task? Otherwise everything will stay as it is right now and both `lintKeyboardLayouts` and `genLayoutsList` will only run if requested manually.

Julow: We can make these task run on `test` for now and change later if it's not working. Perhaps `lintKeyboardLayouts` can be renamed `checkKeyboardLayouts` to avoid confusion ?

deftkHD:
The scripts are now located in the project root directory and the Gradle submodule was merged with the root project.
What about the shell.nix file?

.github/workflows/make-apk.yml
```diff
@@ -1,47 +1,48 @@
-name: Make Apk CI
+name: Make APKs CI
```
Julow, line 1: Previous name was better.

.gitignore
```diff
@@ -2,3 +2,21 @@
...
+/.idea/navEditor.xml
+/.idea/assetWizardSettings.xml
+/.idea/shelf/
+/.idea/workspace.xml
+.DS_Store
+/build
+/captures
+.externalNativeBuild
+.cxx
+local.properties
```
Julow, line 21: Included twice.

gradle.properties
```diff
@@ -0,0 +1,23 @@
+# Project-wide Gradle settings.
+# IDE (e.g. Android Studio) users:
+# Gradle settings configured through the IDE *will override*
+# any settings specified in this file.
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
```
Julow, line 9: Is the memory argument useful ? Otherwise, I'd keep only the encoding argument.

Also, I'd prefer it the comments in this file were removed.

settings.gradle
```diff
@@ -0,0 +1,17 @@
+pluginManagement {
+    repositories {
```
Julow, line 2: Are any of these repositories used ?

src/main/java/juloo.keyboard2/Keyboard2.java
Julow, line 1: Is the `src/main/java` directory needed ? Couldn't the sources reside in `srcs/` as before and the resources in `res/` ?

gen_layouts.py
```diff
@@ -59,13 +59,13 @@ def mk_array(tag, name, strings_items):
     values_items, entries_items = zip(*(system_item + layouts + custom_item)) # unzip
     ids_items = map(lambda s: "@xml/%s" % s if s not in ["system", "custom"] else "-1", values_items)
     root = XML.Element("resources")
-    root.append(XML.Comment(text="DO NOT EDIT. This file is generated, see gen_layouts.py."))
+    root.append(XML.Comment(text="DO NOT EDIT. This file is generated, see app/gen_layouts.py."))
```
Julow, line 62:
```suggestion
    root.append(XML.Comment(text="DO NOT EDIT. This file is generated, see gen_layouts.py."))
```

gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
Julow, line 1: Are these two scripts needed to run Gradle ? It seems optional according to Gradle's documentation.
Does the Android SDK come with Gradle ?

Julow (commented):
I'll try to come up with a `shell.nix`.

src/main/java/juloo.keyboard2/Keyboard2.java
deftkHD: The project layout is specified in the `build.gradle` file under the section `sourceSets`. Gradle separates the different build variants and their source code. This is why `src/main` and `src/debug` exist. Right now I can see two different solutions:

- We use a different layout, which doesn't exactly match the old one, because it has to consider both `debug` and `main` source sets, but is similar. Maybe you have an idea for that, but I think it is not possible to share the `res` folder between resources and assets. So maybe `src/main/<java code>`, `res/main/<resources>` and `res/debug/<debug resources>`?
- I could remove the last resource file in `src/debug/res` by letting Gradle take care of setting the boolean `debug_logs` to true. A debug sourceSet would not be required anymore, we can just use one sourceSet and store the files in `srcs/` and `res/` and hopefully we don't need a second sourceSet in the future.

What do you prefer?

And just btw: In the sourceSets in `build.gradle` file the terms "res" and "resources" are used. While "res" refers to the Android resources, "resources" refers to Java resources. Java resources are not used in this project, but I specified them to make it complete. This should be safe to remove.

settings.gradle
deftkHD: The block `pluginManagement` as the name says is required to provide sources for the Gradle plugins. Without this block Gradle doesn't know where to get e.g. the Android Gradle Plugin, so it is required. It is possible to remove individual repositories, but you never know about dependencies of Gradle plugins and to reduce headaches in the future, I would prefer to leave those repositories as they are. Just Jitpack could be useless.

The block `dependencyResolutionManagement` is more interesting. While the block `pluginManagement` specifies the repositories for the Gradle plugins, this block specifies the repositories for dependencies specified in the `dependencies` block in the build.gradle file. Since there are no dependencies right now, it is currently safe to remove. But adding any additional dependencies in the future will fail because Gradle doesn't know where to search for them.

gradle.properties
deftkHD: It is useful, but not mandatory. When building, Gradle will use its own JVM instance and this argument specifies how much memory it will get. Increasing the memory will improve the build performance, but I think my computer is not a good benchmark for that. As specified by https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory the default memory size is 512M. This project is rather small and efficient so I think it won't have a big impact (if at all) when this flag gets removed.
Additionally I will remove the Kotlin codestyle property since Kotlin isn't used.

settings.gradle
Julow: It's fine if dependency resolution fail when adding. It should be easy to add the required repositories.

deftkHD: I noticed during testing that when removed the build fails due to unmet dependencies. It seems like the block `dependencyResolutionManagement` is also required to manage the dependencies of the Gradle plugins. I can safely remove Jitpack, but besides that and Maven Central for the `pluginManagement` section, every other repository seems mandatory.

Julow: Ok, thanks for trying out.

build.gradle
```diff
@@ -0,0 +1,158 @@
+plugins {
```
Julow, line 1: I think it would be useful to indent this file like the source code and resources.
The source code is indented with 2 spaces, brackets have their own lines and empty lines between large blocks.

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+                keyPassword "$System.env.RELEASE_KEY_PASSWORD"
+            }
+        }
+    }
+    buildTypes {
+        release {
+            minifyEnabled true
+            shrinkResources true
+            debuggable false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt')
```
Julow, line 54: This file is gone.

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+                keyAlias "$System.env.RELEASE_KEY_ALIAS"
+                keyPassword "$System.env.RELEASE_KEY_PASSWORD"
+            }
+        }
+    }
+    buildTypes {
+        release {
+            minifyEnabled true
+            shrinkResources true
+            debuggable false
```
Julow, line 53: I'm not a fan of enabling minifyEnabled and disabling debug. The app is already small enough and these help get better stacktraces when a crash happens.

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt')
+            resValue "string", "app_name", "@string/app_name_release"
+            if (System.env.RELEASE_KEYSTORE) {
+                signingConfig signingConfigs.release
+            }
```
Julow, line 58: Shouldn't the release build fail if this isn't set ?

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+        }
+        println "$bos"
+    }
+    copy {
+        from "$buildDir/special_font.ttf"
+        into "assets"
+    }
+}
+
+tasks.withType(Test).configureEach {
```
Julow, line 101: `syncTranslations` should also be added to tests.

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+        targetCompatibility JavaVersion.VERSION_1_7
+    }
+}
+
+dependencies {
+
+}
+
+tasks.register('buildKeyboardFont') {
+    println "Executing fontforge to build custom font\n"
```
Julow, line 81: Gradle's logs seem to have the empty line first:

```suggestion
    println "\nBuilding assets/special_font.ttf"
```

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+tasks.withType(Test).configureEach {
+    dependsOn 'genLayoutsList'
+    dependsOn 'checkKeyboardLayouts'
+}
+
+tasks.register('genLayoutsList') {
+    println "Executing python script to generate keyboard layouts"
```
Julow, line 107:
```suggestion
    println "\nGenerating res/values/layouts.xml"
```

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+            workingDir = projectDir
+            commandLine "python", "gen_layouts.py"
+            standardOutput = bos
+        }
+        println "$bos"
+    }
+}
+
+tasks.register('checkKeyboardLayouts') {
+    println "Executing python script to check keyboard layouts\n"
```
Julow, line 119:
```suggestion
    println "\nChecking layouts"
```

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+            commandLine("python", "check_layout.py", *layouts)
+            standardOutput = bos
+        }
+
+        new File(projectDir, "check_layout.output").write(bos.toString())
+    }
+}
+
+tasks.register('syncTranslations') {
+    println "Executing python script to generate translation templates"
```
Julow, line 135:
```suggestion
    println "\nUpdating translations"
```

build.gradle
```diff
@@ -0,0 +1,158 @@
...
+            workingDir "$projectDir/special_font"
+            def svgFiles = workingDir.listFiles().findAll {
+                it.isFile() && it.name.endsWith(".svg")
+            }
+
+            commandLine("fontforge", "-lang=ff", "-script", "build.pe", "$buildDir/special_font.ttf", *svgFiles)
+            standardOutput = bos
+        }
+        println "$bos"
+    }
```
Julow, line 94:
```suggestion
    exec {
        workingDir "$projectDir/special_font"
        def svgFiles = workingDir.listFiles().findAll {
            it.isFile() && it.name.endsWith(".svg")
        }

        commandLine("fontforge", "-lang=ff", "-script", "build.pe", "$buildDir/special_font.ttf", *svgFiles)
    }
```
We can leave standard output alone. Same for all the others except `checkKeyboardLayouts`.

Julow (commented):
I've made a working `shell.nix` at this branch that you can cherry-pick: https://github.com/Julow/Unexpected-Keyboard/tree/452_shell_nix

Julow:
Thanks for this huge amount of work :) This will hopefully make the app more maintainable.

I usually "squash and merge" but as this PR is large and touch many parts, I'll manually squash commits.

deftkHD:
I think it will still unfortunately take some time (1-2 weeks) until I can work on this again, but I did not forget it.

build.gradle
deftkHD: This statement loads the default proguard configuration file for Android which does not exist in the directory of the project. I updated this statement when removing the empty proguard file as can be seen in #452 7e8a417. Proguard is used to optimize the APK built by Gradle and https://developer.android.com/build/shrink-code recommends to leave this as it is right now.

build.gradle
```diff
@@ -0,0 +1,145 @@
...
+        workingDir = projectDir
+        commandLine "python", "gen_layouts.py"
+    }
+}
+
+tasks.register('checkKeyboardLayouts') {
+    println "\nChecking layouts"
+    new ByteArrayOutputStream().withStream { bos ->
+        exec {
+            def layouts = new File(projectDir, "res/xml").listFiles().findAll {
```
Julow, line 113: Layouts must be sorted alphabetically for reproducibility.

Julow:
I've picked the changes in the special font files in 3d36ecb to make the diff smaller in this PR.

Julow:
I've made some changes to reduce the diffs and I've merged 851d22da6 separately. This is ready to be merged.

Thanks for this large contribution that I hope will make the app more maintainable by more people.

PR #477, opened 2023-10-14
Resolve layout warnings in neo2

matthiakl:
The update to the bottom row did not get into neo2. This resolves the warnings found by check_layout.py:
- Replace "-" with "—" on the "ß" key
   "-" should be U+2011 (non-breaking hyphen), but this mostly is not recognized by apps, —" (U+2014, EM Dash) is in the second layer of the same key and was not available before.
- ~"$" and "€" are available again ("$" is not available from the additional key menu, is this intentional)~
  edit: $ was already present, I reverted this point
- Update bottom row special keys

https://github.com/Julow/Unexpected-Keyboard/blob/5b4345088d419ab2b26df11fbc80bc265181df93/check_layout.output#L65-L68

### DIFF
--- a/res/xml/latn_neo2.xml
+++ b/res/xml/latn_neo2.xml
@@ -7,8 +7,8 @@
     <key key0="l" key2="3" key4="["/>
     <key key0="c" key2="4" key4="]"/>
     <key key0="w" key2="5" key4="^"/>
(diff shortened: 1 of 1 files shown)


Julow (approved):
Thanks for maintaining the layout :)

PR #479, opened 2023-10-16
Add Bulgarian BDS layout

lyubomirv:
I  have added the BDS layout for the Bulgarian language, which is widely used and preferred by many people


Julow (commented):
The third row is too wide and that doesn't look right:
shot

It would look better with the `width` attributes removed in the third row and `shift="1" ` added to the first key of the first row.
It's possible to keep the width of shift and delete by moving delete to the first row.

res/xml/cyrl_ueishsht.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard name="УЕИШЩ (Български, БДС)" script="cyrillic">
+  <row>
+    <key key0="у" key2="1" key4="esc"/>
+    <key key0="е" key1="!" key2="2" key3="\@" key4="#"/>
```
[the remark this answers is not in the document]
lyubomirv: I'll fix it. I'll also move the right-most letter from the bottom (б) to the top row, as it is common on mobile keyboards (now I know why).

res/xml/cyrl_ueishsht.xml
```diff
@@ -0,0 +1,41 @@
...
+    <key key0="б" key2="€"/>
+  </row>
+  <row>
+    <key key0="ь" key1="tab" key2="`"/>
+    <key key0="я"/>
+    <key key0="а"/>
+    <key key0="о" key2="-" key3="_"/>
+    <key key0="ж" key2="=" key3="+"/>
+    <key key0="г" key4="}" key3="{"/>
+    <key key0="т" key2="~" key3="[" key4="]"/>
```
Julow, line 23: `~` is present twice.

Julow (commented):
Looks good :) I have one more comment and then it's good to merge.

res/xml/cyrl_ueishsht.xml
lyubomirv: Thanks for noting it! I've just fixed it.

Issue #496, opened 2023-11-27
[Partial Solution][Revisiting Closed Feature Request] To forcibly show persistent keyboard at any time


LostRuins:
Greetings. I was a long-time user of Hacker Keyboard, and one of the most essential features was the "Permanent Notification" aka "Show Keyboard at any time". Unfortunately, that feature has been broken since Android 12, and that project is abandoned, so I came here instead.

Now, I understand that you've mentioned in https://github.com/Julow/Unexpected-Keyboard/issues/71 https://github.com/Julow/Unexpected-Keyboard/issues/392 and https://github.com/Julow/Unexpected-Keyboard/issues/69 that due to the crippling of `SHOW_FORCED`, it is no longer possible to invoke the keyboard at will in the same way as before. However, there might be a workaround.

There is an absolutely [horrible ad-infested unfree app on the play store called iGame Keyboard](https://play.google.com/store/apps/details?id=com.infinityplus.igamekeyboard&hl=en_US)  that achieves this by utilizing the `Display over other apps` permission to add a small floating button that is rendered over all other screens. This button is displayed when a notification is selected, and only has one purpose, to trigger the keyboard when pressed. I can confirm **this works even when targeting Android 13 (API 33)**.

After installing that app, I was then able to change my active keyboard **back** to Unexpected-Keyboard. Then, I am able to **invoke Unexpected-Keyboard** anywhere by pressing the overlay floating button that opens *a* keyboard. I can confirm that the arrow keys in unexpected keyboard do indeed work in navigating the menus of other applications without text inputs.

Here is a screen recording of this approach: 
![proof1](https://github.com/Julow/Unexpected-Keyboard/assets/39025047/f69b4ae6-3a72-43c0-8b80-6cb8bc23e4c9)

https://github.com/Julow/Unexpected-Keyboard/assets/39025047/9eb59319-6cfa-45ec-9695-28838b27d957

Looking around I can see that on-demand keyboards are a dearly desired feature, so I think many including myself would be thrilled if you decide to implement it. I would much rather use an implementation within your program that have to deal with a third-party ad-infested workaround,

Do let me know if you need any additional information. Thanks and appreciate your hard efforts.

JackDSRS:
Ahem...unkosher, but L ⊃ c k y P @ t c h ∃ r re: ads

Issue #392, opened 2023-07-05
How to always have the navigation bar button to bring up the keyboard?


madorian:
I've found in setting a way to always have it there, but that doesn't quite work;)

I'm on an s22 ultra and there are many times that button is not there to bring it up. 

Any pointers as to what I can try?;)

Mispille:
Hi and welcome

Could you be more precise on your request?

Looks like it depends on Samsung more than any keyboard?

Julow:
This have been requested before and I decided not to implement it because it can't work on Android 12 and above.

Though, I'm not against the idea if someone can make it work :)

Mispille:
Sorry @Julow I don't know what it's about, can't help on this one !

Issue #71, opened 2022-02-13
Add a way to forcibly invoke the keyboard


Poopooracoocoo:
Hacker's Keyboard is probably the only keyboard that allows you to type in input fields that don't automatically invoke the keyboard. It has a persistent notification that once tapped on opens the keyboard. If Unexpected Keyboard had an optional persistent notification or a quick setting toggle to do the same, that'd be be great!

Julow:
Thanks for the suggestion. There's an other issue about this: https://github.com/Julow/Unexpected-Keyboard/issues/69

Issue #69, opened 2022-02-10
[feature] forcibly show the keyboard at any time


GUUser91:
I want to open up the Unexpected-Keyboard app even if there is no text field in focus.

For example, the Hacker's Keyboard app has this feature. It has a notification option that if I tap, the keyboard will pop up. Here's a video of me opening the Hacker's Keyboard app to use the keyboard shortcuts on the mpv app.

https://user-images.githubusercontent.com/77461922/153512355-075112bf-dcc2-4dbf-bddf-68dfe484a88a.mp4

07416:
Was it ever possible to use Home/End and PG UP/DN keys in a web browser on Android? Do the Android browsers support hotkeys in the first place?

Julow:
That's an interesting idea, thanks!

@07416 I think at least browser extensions will be able to interpret the key presses (eg. vimium). Other android interfaces have hotkeys (even though they are generally of bad quality).

dilipkn007:
In addition to that make it floatable

Efreak:
@07416 I regularly use a media remote with my tablet. The arrow keys are implemented as dpad buttons, and the page keys are just pgup/pgdn. This works perfectly in Firefox/Lynket as well as several text editors and other apps; with the ability to cast the screen, I'd very much expect standard keyboard navigation to work in any other browser as well.

The entire Android interface does have basic hotkey support (ctrl+zxcva works in any standard text field, alt+tab switches apps, arrow keys navigate lists and the home screen, etc), which is rather expected imo since Android *does not require a touchscreen*--while this isn't useful for personal devices, I've seen kiosks built on Android that use a button, trackball and keyboard, with a standard display behind heavy plastic. Much cheaper than replacing a touchscreen every time it gets vandalized.

mysori:
this hackers keyboard feature is no longer working on Android 12 and the hackers Keyboard project is dead and the developer is not interested in updating it anymore, so it has become important that this feature gets implemented in unexpected Keyboard.
This feature is more useful for educational purposes than  gaming or entertainment purposes as there is no built support in Android to achieve this.

Julow:
Unfortunately, the mechanism for doing that has been removed in Android 12 (the [SHOW_FORCED](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#SHOW_FORCED) flag has been deprecated and has changed meaning at the same time).

[...]

PR #497, opened 2023-11-30
Adds the ability for users to invoke the soft keyboard anywhere, at any time, even on screens that do not normally allow keyboards. This approach works even on Android 13 (API 33)

LostRuins:
I went ahead and did it myself, because I really needed the functionality. I am quite certain no other FOSS keyboard has this ability working in Android 13.

This PR adds the ability for users to invoke the soft keyboard anywhere, at any time, even on screens that do not normally allow keyboards. This approach works even on Android 13 (API 33), and is an essential replacement for the now-defunct Hacker Keyboard's Permanent Notification, which had a similar capability but stopped functioning after API 30 (https://github.com/klausw/hackerskeyboard/issues/897).

Closes https://github.com/Julow/Unexpected-Keyboard/issues/496, Closes https://github.com/Julow/Unexpected-Keyboard/issues/71, Closes https://github.com/Julow/Unexpected-Keyboard/issues/392 and Closes https://github.com/Julow/Unexpected-Keyboard/issues/69 

How to use:
1. Launch Unexpected Keyboard, and click `Show Persistent Keyboard`
2. Grant the `Draw Over Other Apps` permission, allowing overlays to be displayed.
3. Now, whenever you require a keyboard, launch the app and press `Show Persistent Keyboard`
4. A button overlay will appear in the top left corner
5. Navigate back to the app you want to use that does not support keyboards.
6. Click the `Show` button in the top left corner. The overlay will be closed, and at the same time the soft keyboard will be invoked.
7. Enjoy your ability to use a keyboard anywhere, like Richard Stallman intended.

demo

### DIFF
--- a/res/layout/launcher_activity.xml
+++ b/res/layout/launcher_activity.xml
@@ -6,4 +6,5 @@
   <TextView style="@style/paragraph" android:text="https://github.com/Julow/Unexpected-Keyboard" android:autoLink="web" android:linksClickable="true"/>
   <TextView android:id="@+id/launcher_tryhere_text" style="@style/paragraph" android:text="@string/launcher_tryhere"/>
   <EditText android:id="@+id/launcher_tryhere_area" style="@style/paragraph" android:inputType="text"/>
+  <Button style="@style/paragraph" android:text="Show Persistent Keyboard" android:onClick="startOverlayService" android:layout_width="wrap_content"/>
 </LinearLayout>
--- a/res/layout/overlay_button.xml
+++ b/res/layout/overlay_button.xml
@@ -2,18 +2,23 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="wrap_content"
     android:layout_height="wrap_content"
+    android:padding="10dp"
     android:orientation="horizontal">
 
-    <Button
-        android:id="@+id/invokeKeyboardButton"
-        android:layout_width="72dp"
-        android:layout_height="wrap_content"
-        android:text="Show" />
 
-    <Button
+    <ImageView
+        android:id="@+id/invokeKeyboardButton"
+        android:layout_width="50dp"
+        android:layout_height="50dp"
+        android:background="@android:color/darker_gray"
+        android:gravity="center"
+        android:scaleType="fitCenter"
+        android:src="@drawable/ic_launcher_foreground" />
+
+    <ImageView
         android:id="@+id/closeOverlayButton"
-        android:layout_width="32dp"
-        android:layout_height="32dp"
+        android:layout_width="25dp"
+        android:layout_height="25dp"
         android:background="#FF8282"
-        android:text="x" />
+        android:src="@android:drawable/ic_menu_close_clear_cancel" />
 </LinearLayout>
--- a/res/layout/overlay_button.xml
+++ b/res/layout/overlay_button.xml
@@ -1,0 +1,19 @@
+<!-- overlay_button.xml -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:orientation="horizontal">
+
+    <Button
+        android:id="@+id/invokeKeyboardButton"
+        android:layout_width="72dp"
+        android:layout_height="wrap_content"
+        android:text="Show" />
+
+    <Button
+        android:id="@+id/closeOverlayButton"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
+        android:background="#FF8282"
+        android:text="x" />
+</LinearLayout>
--- a/srcs/juloo.keyboard2/LauncherActivity.java
+++ b/srcs/juloo.keyboard2/LauncherActivity.java
@@ -35,7 +35,7 @@ public class LauncherActivity extends Activity
   private static final int REQUEST_CODE_OVERLAY_PERMISSION = 1601;
   private void requestOverlayPermission() {
     // Request the permission using an Intent
-    Intent intent = new Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + getPackageName()));
+    Intent intent = new Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.fromParts("package", getPackageName(), ""));
     startActivityForResult(intent, REQUEST_CODE_OVERLAY_PERMISSION);
   }
 
@@ -58,6 +58,7 @@ public class LauncherActivity extends Activity
       // Check if the permission was granted
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && Settings.canDrawOverlays(this)) {
         Toast.makeText(this, "Overlay permissions are granted. You can now launch persistent keyboards.", Toast.LENGTH_SHORT).show();
+        startService(new Intent(this, OverlayService.class));
       } else {
         Toast.makeText(this, "Overlay permissions are required for launching persistent keyboard.", Toast.LENGTH_SHORT).show();
       }
--- a/srcs/juloo.keyboard2/LauncherActivity.java
+++ b/srcs/juloo.keyboard2/LauncherActivity.java
@@ -2,6 +2,8 @@ package juloo.keyboard2;
 
 import android.app.Activity;
 import android.content.Intent;
+import android.net.Uri;
+import android.os.Build;
 import android.os.Build.VERSION;
 import android.os.Bundle;
 import android.provider.Settings;
@@ -9,6 +11,7 @@ import android.view.KeyEvent;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.TextView;
+import android.widget.Toast;
 
 public class LauncherActivity extends Activity
 {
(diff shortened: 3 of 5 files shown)


LostRuins:
Feel free to modify or change anything, I am not an android dev so I might not be following best practices.

Julow (commented):
This is fantastic, thanks! I'll make a proper review and merge this soon.

Only APKs built by the CI are trustworthy, here's the latest for this PR: https://github.com/Julow/Unexpected-Keyboard/actions/runs/7049282572?pr=497

LostRuins:
Worth noting that on android 13, it seems that there must be at least one navigable ui element for the keyboard to remain displayed. Otherwise, it gets dismissed. If you would have been able to plug a real keyboard and navigate that app via arrow keys, then it would probably work.

Spike-from-NH:
I'm glad you took the initiative!
① Can I drag this button somewhere else if it occludes something important?
② The "X" box is to dismiss the button? Can you jam them together to better show they are parts of the same control? If dismissed, can they be re-enabled in the way they were originally?
③ If Draw Over Other Apps is not granted, do you ask for it?
④ Add the enable button to the Settings menu as well?
⑤ Does it look good in all Themes?

res/layout/overlay_button.xml
```diff
@@ -0,0 +1,19 @@
...
+        android:layout_width="72dp"
+        android:layout_height="wrap_content"
+        android:text="Show" />
+
+    <Button
+        android:id="@+id/closeOverlayButton"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
+        android:background="#FF8282"
+        android:text="x" />
```
Julow, line 18: The close button takes a lot of place and might not be useful. The show button could turn into the close button while the keyboard is showing.

res/layout/launcher_activity.xml
```diff
@@ -6,4 +6,5 @@
   <TextView style="@style/paragraph" android:text="https://github.com/Julow/Unexpected-Keyboard" android:autoLink="web" android:linksClickable="true"/>
   <TextView android:id="@+id/launcher_tryhere_text" style="@style/paragraph" android:text="@string/launcher_tryhere"/>
   <EditText android:id="@+id/launcher_tryhere_area" style="@style/paragraph" android:inputType="text"/>
+  <Button style="@style/paragraph" android:text="Show Persistent Keyboard" android:onClick="startOverlayService" android:layout_width="wrap_content"/>
```
Julow, line 9: Opening the app and clicking this button is a lot of work. Any hope of having a permanent notification instead ?

res/layout/overlay_button.xml
```diff
@@ -0,0 +1,19 @@
...
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:orientation="horizontal">
+
+    <Button
+        android:id="@+id/invokeKeyboardButton"
+        android:layout_width="72dp"
+        android:layout_height="wrap_content"
+        android:text="Show" />
```
Julow, line 11: This buttons shows up on top of every other apps. "Show" doesn't mean much without context. The app's icon or a keyboard symbol would be more intuitive.

srcs/juloo.keyboard2/LauncherActivity.java
```diff
@@ -26,6 +29,39 @@ public void onCreate(Bundle savedInstanceState)
     if (VERSION.SDK_INT > 28)
       _tryhere_area.addOnUnhandledKeyEventListener(
           this.new Tryhere_OnUnhandledKeyEventListener());
+
+  }
+
+  private static final int REQUEST_CODE_OVERLAY_PERMISSION = 1601;
+  private void requestOverlayPermission() {
+    // Request the permission using an Intent
+    Intent intent = new Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + getPackageName()));
```
Julow, line 38: `Uri.parse` can be replaced by `Uri.fromParts("package", getPackageName(), "")`.

srcs/juloo.keyboard2/LauncherActivity.java
```diff
@@ -26,6 +29,39 @@ public void onCreate(Bundle savedInstanceState)
...
+
+  @Override
+  protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+    super.onActivityResult(requestCode, resultCode, data);
+
+    //handle response for overlay permission request
+    if (requestCode == REQUEST_CODE_OVERLAY_PERMISSION) {
+      // Check if the permission was granted
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && Settings.canDrawOverlays(this)) {
+        Toast.makeText(this, "Overlay permissions are granted. You can now launch persistent keyboards.", Toast.LENGTH_SHORT).show();
```
Julow, line 60: We could start the OverlayService here.

Julow (commented):
This doesn't work on my device, the keyboard opens then closes immediately on top of any app. Any idea on how to debug this ?

LostRuins:
Hi @Julow, I have a bit of bad news. It seems like I was mistaken on the specifics this approach.

If the *user's device* is below Android 13, it will work in _all_ cases.
If the *user's device* is at or above Android 13, and the *target SDK* of the app is below API 33, it will _also_ work.
However, if the APK target SDK exceeds 33, **and** the device is **also** android 13 or above, then this approach does not work anymore, as the keyboard will be closed the moment the view loses focus.

You can try by either using any older device, or downloading some random older app from fdroid. I did not notice this at first, as many of the apps installed on my device are older APKs. 

With this disappointing news, I am not sure if this warrants closing this PR or not, it's your choice... perhaps it could still be of some use either for older apps or older devices like in my own use case.

Nevertheless, I have tidied up the PR a bit, so do take a look. 
- The button has been changed to an ImageView with the app logo as an icon.
- The dismiss button has been resized and changed to be more intuitive looking.
- They overlay is now persistent and does not disappear on a single use. Instead, the keyboard is invoked with this flow:
  1. Overlay view starts as non-focusable (`FLAG_NOT_FOCUSABLE`), allowing normal interaction with the underlyring screens.
  2. The invoke keyboard button is pressed by the user.
  3. The overlay view layout params is switched to `FLAG_NOT_TOUCH_MODAL`, allowing it to be focused.
  4. The overlay view requests focus, then the soft keyboard is invoked with `inputMethodManager.toggleSoftInput`
  5. The overlay view is switched back to non-focusable, allowing normal interaction with the underlyring screens.
- The overlay button remains on screen and re-usable until manually dismissed, which allows the keyboard to be triggered at will multiple times.
- Applied all the other changes you have suggested.

Here is another video where I attempt to use this new version on multiple apps on my Android 13 device, you can see it works for any app below target SDK 33 (the bottom group), while failing for those above (the top group). I did also test it on my Android 9 device and it works fine in all cases. I also demonstrate the new interface, switching between apps and invoking the keyboard with touches visible. 

https://github.com/Julow/Unexpected-Keyboard/assets/39025047/1da21704-e30e-469e-a355-42376025e24c

Also replying @Spike-from-NH : 
1. Not at the moment
7. Adjusted
8. Yes
9. See (5)
10. Let's see if its still worth including first considering the new limitations found.

diejuse:
What about this solution??? https://www.youtube.com/watch?v=OCTiacmVu_Y&ab_channel=ArtTV%3AAGorgeousPlace
This keyboard is permanent and works even on Android 14. I just discovered it and made a comment on the video. I think it's a perfect solution but you need to have more options.

RetrogisusDEV:
I think that a good solution would be to use accessibility, to invoke the keyboard without the need for any text field, and it would be more shareable with previous APIs, than how accessibility is implemented? I don't know, but it would be the most viable solution for previous api (estimated up to api 19)

diejuse:
> I think that a good solution would be to use accessibility, to invoke the keyboard without the need for any text field, and it would be more shareable with previous APIs, than how accessibility is implemented? I don't know, but it would be the most viable solution for previous api (estimated up to api 19)

Yes that's how it is! I don't understand why there isn't any keyboard of this type! (apart from this). It would be perfect to have a floating icon above the rest of the apps and be able to display it whenever you want to send letters.
It would even be usable as a gamepad!

doak:
I am very much looking forward to this feature.

diejuse:
Anyone knows a always visible keyboard like this for Android? I need it.

LostRuins:
I will be closing this PR due to lack of interest. 

For future readers, this approach only works for apps with target sdk < 33.
For anyone who wants a fully working solution on Android 13+, use Game Keyboard 2.

diejuse:
> to invoke the keyboard without the need for any text field,

Couldn't you try adding the keyboard functionality you mentioned to your keyboard so we could use your wonderful keyboard at any time? 
That wat the keyboard could be to invoke without the need for any text field, and it would be more shareable with previous APIs, as said @RetrogisusDEV

diejuse:
> Sorry for the lack of interest. The target sdk constraint is a big problem as Google Play would not accept upgrades and would lower the rank of the app.
> 
> Though, I'd accept to release the app with this on F-Droid and Github, for a bit of time. It'd need to be made optional.
> 
> This do not work on my device (running Android 13): the keyboard seems to open and close immediately, no logs. Do you think there's a way to detect when it wouldn't work and not offer the feature in this case ?

Couldn't you try adding the keyboard functionality you mentioned to your keyboard so we could use your wonderful keyboard at any time?
That wat the keyboard could be to invoke without the need for any text field, and it would be more shareable with previous APIs, as said @RetrogisusDEV

RetrogisusDEV:
> > Sorry for the lack of interest. The target sdk constraint is a big problem as Google Play would not accept upgrades and would lower the rank of the app.
> > 
> > Though, I'd accept to release the app with this on F-Droid and Github, for a bit of time. It'd need to be made optional.
> > 
> > This do not work on my device (running Android 13): the keyboard seems to open and close immediately, no logs. Do you think there's a way to detect when it wouldn't work and not offer the feature in this case ?
> 
> @Julow Couldn't you try adding the keyboard functionality you mentioned to your keyboard so we could use your wonderful keyboard at any time?
> That wat the keyboard could be to invoke without the need for any text field, and it would be more shareable with previous APIs, as said @RetrogisusDEV

In fact, I've been trying and I'm already giving up xD

PR #505, opened 2023-12-18
Update strings.xml

polyctena:
The recent 8 missing strings in German. 
The "translations" issue is quite convenient :)


Julow (approved):
Thanks :)

Issue #373, opened 2023-06-03
Translation updates


Julow:
A comment will be added to this issue every time the translations need to be updated.

If you wish to be notified when translations need to be updated,
click on the "Subscribe" button in the right column of this page.

Guidelines on how to contribute translations [are here](https://github.com/Julow/Unexpected-Keyboard/blob/master/CONTRIBUTING.md#translations).

If you have comments or questions, please open a different issue to avoid spamming the contributors.

Julow:
New strings are added in 69e0b4c.

Julow:
Many thanks to all of you ! Your quick response was awesome :)

The next release adds many strings:
- Custom extra keys preferences (6747669 and 458e17b)
- Added option (fce8ff7)
- Layouts preference (dad5f57)
- Key descriptions (ae791ab)

Here's the current state of the translations:

```
res/values-ro/strings.xml: missing 14 strings
res/values-ko/strings.xml: missing 52 strings
res/values-de/strings.xml: missing 14 strings
res/values-tr/strings.xml: missing 49 strings
res/values-vi/strings.xml: missing 22 strings
res/values-pt/strings.xml: missing 14 strings
res/values-pl/strings.xml: missing 9 strings
res/values-it/strings.xml: missing 45 strings
res/values-cs/strings.xml: missing 9 strings
res/values-fr/strings.xml: uptodate
res/values-zh-rCN/strings.xml: missing 9 strings
res/values-fa/strings.xml: missing 14 strings
res/values-lv/strings.xml: missing 18 strings
res/values-es/strings.xml: missing 7 strings
res/values-ru/strings.xml: missing 14 strings
```

Julow:
I plan to release soon. Here's the current state of the translation:
```
res/values-ro/strings.xml: missing 34 strings
res/values-ko/strings.xml: missing 71 strings
res/values-de/strings.xml: missing 8 strings
res/values-tr/strings.xml: missing 3 strings
res/values-vi/strings.xml: missing 42 strings
res/values-pt/strings.xml: missing 8 strings
res/values-pl/strings.xml: missing 8 strings
res/values-it/strings.xml: missing 64 strings
res/values-cs/strings.xml: missing 8 strings
res/values-fr/strings.xml: uptodate
res/values-zh-rCN/strings.xml: missing 29 strings
res/values-fa/strings.xml: missing 34 strings
res/values-lv/strings.xml: missing 8 strings
[...]

PR #525, opened 2024-01-18
Add Ukrainian translation

Sestowner:


Julow (approved):
Awesome, thanks!

Don't feel obliged to maintain the translation over time but if you do, you can subscribe to the notifications of this issue where I post before each releases: https://github.com/Julow/Unexpected-Keyboard/issues/373 (I don't see who is subscribed)

PR #546, opened 2024-02-05
Update CI to node.js-20

sdrapha:
node.js v16 is end-of-life and github actions needed an update to node.js v20


Julow (approved):
Thanks!

PR #549, opened 2024-02-07
Update latn_qwerty_tly.xml

vedamanavi:
Removing "g" as it was used two times as a main key


Julow (commented):
You should add `shift="0.5"` on the changed `<row>`. It currently looks like this:
shot

vedamanavi:
@Julow Thanks for your guidance. Done.

Julow (approved):
Looks good!
shot

PR #553, opened 2024-02-11
Update README.md

BogdanLata:
I made more clearer how to use the keys


Julow (approved):
That's very useful, thanks!

I feel like the store description is not optimal. Do you have an idea on how to improve it ? https://github.com/Julow/Unexpected-Keyboard/blob/master/res/values/strings.xml#L6

PR #557, opened 2024-02-16
Add QWERTY GB layout

RyanGibb:
I had a go at adapting the latn_qwerty_us for my UK keyboard muscle memory and thought it might be useful for others


Julow (approved):
Thanks!

PR #560, opened 2024-02-18
neo2: Update bottom row

matthiakl:
Update bottom row in neo2 layout to match default and added new keys.
Resolves layout warnings.


Julow (approved):
Thanks for maintaining the layout!

PR #562, opened 2024-02-18
Gujarati phonetic layout

Yogesh-B:
Added Gujarati phonetic layout.

- [x] added layout xml
- [x] added changes in `KeyModifier.java` for shift key combinations
- [x] re-generated `values/layouts.xml` 
- [x] re-generated `check_layout.output`

Tested debug APK build(gradle and java), working fine on my device.

### DIFF
--- a/srcs/layouts/guj_phonetic_in.xml
+++ b/srcs/layouts/guj_phonetic_in.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<keyboard name="ગુજરાતી ફોનેટિક - Gujarati Phonetic" script="gujarati">
+  <row>
+    <key key0="1" key1="`" key2="૧" key3="!" key4="esc"/>
+    <key key0="2" key1="~" key2="૨" key3="\@" key4="૱"/>
+    <key key0="3" key2="૩" key3="\#"/>
+    <key key0="4" key1="₹" key2="૪" key3="$"/>
+    <key key0="5" key2="૫" key3="%"/>
+    <key key0="6" key2="૬" key3="^"/>
+    <key key0="7" key2="૭" key3="&amp;"/>
+    <key key0="8" key2="૮" key3="*"/>
+    <key key0="9" key1="_" key2="૯" key3="(" key4=")"/>
+    <key key0="0" key1="+" key2="૦" key3="-" key4="="/>
+  </row>
+  <row>
+    <key key0="ટ" key1="“" key4="”"/>
+    <key key0="ડ"/>
+    <key key0="ે" key1="એ"/>
+    <key key0="ર" key1="ઋ" key3="ૃ"/>
+    <key key0="ત" key1="ટ"/>
+    <key key0="ય"/>
+    <key key0="ુ" key1="ઉ"/>
+    <key key0="િ" key1="ઇ" key3="ઁ"/>
+    <key key0="ો" key1="ઓ"/>
+    <key key0="પ" key1="૰" key3="॒"/>
+  </row>
+  <row>
+    <key shift="0.5" key0="ા" key1="tab" key4="`" key3="અ"/>
+    <key key0="સ"/>
+    <key key0="દ" key1="ડ"/>
+    <key key0="્" key3="ૄ" key1="ૠ"/>
+    <key key0="ગ" key2="-" key3="_"/>
+    <key key0="હ" key2="=" key3="+"/>
+    <key key0="જ" key3="{" key4="}"/>
+    <key key0="ક" key2="ૡ" key3="[" key4="]" key1="ૢ"/>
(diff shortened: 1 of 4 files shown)


srcs/layouts/guj_phonetic_in.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<keyboard name="ગુજરાતી ફોનેટિક - Gujarati Phonetic" script="gujarati">
+  <row>
```
Julow, line 3: The top row doesn't look good in combination to the built-in number row (that can be enabled in the settings). Do you think every users of this layout will want to have this row ?

Also, the layout is very high due to top row. It takes almost half of the screen on my device. You can make the top row a bit smaller with `height="0.75"` (like the built-in number row).

srcs/layouts/guj_phonetic_in.xml
```diff
@@ -0,0 +1,49 @@
...
+    <key key0="1" key1="`" key2="૧" key3="!" key4="esc"/>
+    <key key0="2" key1="~" key2="૨" key3="\@" key4="૱"/>
```
Julow, line 5: The layout contains a lot of ASCII symbols like `~`, `@`, `^`, etc.. Are these really used ?

Julow (approved):
Looks good :) I have left some comments but I'd be happy to merge it like this if you think it's ready.

srcs/layouts/guj_phonetic_in.xml
Yogesh-B: mostly not much useful frequently but kept it as default layout has it and most keyboards contains these.

srcs/layouts/guj_phonetic_in.xml
Yogesh-B: we can do something like this to remove the top row. [**as you see in the Screen Shot, numbers are shown properly in keyboard but output is not in devanagari. I feel like there is a bug. It devanagari numbers work fine in normal view, just this number view is not working proper]

srcs/layouts/guj_phonetic_in.xml
Julow: Makes sense.

srcs/layouts/guj_phonetic_in.xml
Julow: The devanagari digits are specified here: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L1060
Are they wrong ?

To change the numpad digits when your layout is visible, you'd need to add a new similar map and call it from here: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L100

Yogesh-B: The first screen shot just shows the layout in Devanagari(or any other language), but when typing, it puts output as english numbers. I think it is using values from https://github.com/Julow/Unexpected-Keyboard/blob/master/res/xml/numpad.xml .  

I tried adding Gujarati number in the java file you suggested, but had same issue as mentioned above.

Yogesh-B:
I could not find option to enable script specific numeric row, hence currently adding first row permanent and have made changes to row height so that it does not take up space.

Now the layout looks quite good and feels almost same as experience on PC. No issue in merging it.

Numeric layout related issue can be done in another pull/issue.

srcs/layouts/guj_phonetic_in.xml
Julow: You are right, there's a serious bug in the numpad. I fixed it in 8c7559d and fixed the number row in 8b5d971.

You can now remove the number row from your layout and enable the number row in the settings.

Note that the layout must also contain the numerals for people that don't use the number row option.

Julow:
Now that the number row and numpad are fixed on master, could you pull master and update the layout ?

I've added support for the gujarati numerals in this branch: https://github.com/Julow/Unexpected-Keyboard/tree/gujarati_numpad_script
Though, I'm not sure I got them right so I'd prefer if you could push that to your branch or push the implementation you had earlier.

srcs/layouts/guj_phonetic_in.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<keyboard name="ગુજરાતી ફોનેટિક - Gujarati Phonetic" script="gujarati">
+  <row>
+    <key key0="ટ" key1="`" key2="1" key3="!" key4="esc"/>
```
Julow, line 4: The western numerals are on the keyboard instead of the gujarati ones. Is this intended ?

Julow (approved):
The layout looks nice :) I have a comment but I would merge before next release anyway.

Many thanks for helping improve the keyboard :)

PR #578, opened 2024-03-03
Update pl/strings.xml

ChasmSolacer:


Julow (approved):
Incredibly fast response! Thanks

PR #579, opened 2024-03-03
Update German strings.xml

polyctena:


Julow (approved):
Thanks! 

Due to my bad scheduling, this will not be part of the release. Sorry!

PR #583, opened 2024-03-08
Update to Unicode Emoji v15.1

### LINKED ISSUE #572: Missing emojis
There are some emojis missing from the keyboard, only one im aware of so far is the melting head one

tmqCypher:
Resolves #572

### DIFF
--- a/res/raw/emojis.txt
+++ b/res/raw/emojis.txt
@@ -1,54 +1,62 @@
-156
-:grinning: 😀 grinning face
-:smiley: 😃 grinning face with big eyes
-:smile: 😄 grinning face with smiling eyes
-:grin: 😁 beaming face with smiling eyes
-:satisfied: 😆 grinning squinting face
-:sweat_smile: 😅 grinning face with sweat
+168
+:u1F600: 😀 grinning face
+:u1F603: 😃 grinning face with big eyes
+:u1F604: 😄 grinning face with smiling eyes
+:u1F601: 😁 beaming face with smiling eyes
+:u1F606: 😆 grinning squinting face
+:u1F605: 😅 grinning face with sweat
 :u1F923: 🤣 rolling on the floor laughing
-:joy: 😂 face with tears of joy
+:u1F602: 😂 face with tears of joy
 :u1F642: 🙂 slightly smiling face
 :u1F643: 🙃 upside-down face
-:wink: 😉 winking face
-:blush: 😊 smiling face with smiling eyes
-:innocent: 😇 smiling face with halo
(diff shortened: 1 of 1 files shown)


res/raw/emojis.txt
```diff
@@ -1,54 +1,62 @@
-156
-:grinning: 😀 grinning face
```
Julow, line 2: The first part was used to name emojis internally and was used when saving the most commonly used emojis. This will make some of the saved emojis disappear.

Though, I don't want to slow this down, instead I'd prefer to improve the way emojis are handled internally.
Here's my idea:
- The emoji names would be extracted from the previous version of this file and would be saved a new file. This will be used to migrate the saved information.
- The "last used emojis" feature would save the emojis characters directly and not their names.
- This file is kept as it is for easier upgrade in the future (or changed to be easier to upgrade) but is no longer embedded in the app. A smaller file is generated at test-time containing the data that should be embedded in the app (the grouped emojis without name and description). This should make the app slightly smaller and faster.
- The names and descriptions of emojis are otherwise not needed internally and should not be loaded by the app. `srcs/juloo.keyboard2/Emoji.java` is removed, `KeyValue` is used instead.
  The emoji descriptions could be added to the app later if they are needed.

What do you think ?

Julow (approved):
Many thanks!

tmqCypher:
> Many thanks!

No, thank you! This app and your work on it have been indispensable to me for years now. I'm just happy to finally make a small contribution.

res/raw/emojis.txt
tmqCypher: Switching to a more compact emoji implementation should be pretty straightforward. I'll review Unicode's documentation to make sure it's reasonably future proof.

As for the recent emojis, I think it would suffice to hardcode a map from the old names to the characters and remove it at a later date when most users' SharedPreferences have been updated.

Thanks for the insight! It might be a week or two until I can get around to it, but I'll open a new PR when I have something to show.

Julow: Thanks for looking into it :)

> As for the recent emojis, I think it would suffice to hardcode a map from the old names to the characters and remove it at a later date when most users' SharedPreferences have been updated.

Yes!

Julow: Also, the old names might be useful data for future features.

My preferred way to build a map in Java is to use a `case` when the map is accessed infrequently with different keys. Similarly to KeyValue: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L345

PR #592, opened 2024-03-16
Add Shaw Imperial layout

la-ninpre:
I see that there are no options for typing in shavian on android besides the keyman app. It is okay, but somewhat slow and it's not great switching back and forth between keyboard apps all the time. After I learned that one can add a custom layout to Unexpected keyboard, I immediately thought about finally unifying all my layouts in one app, so i don't have to switch all the time.

Hope it will be useful for somebody.

And thanks for this great project!

### DIFF
--- a/srcs/layouts/shaw_imperial_en.xml
+++ b/srcs/layouts/shaw_imperial_en.xml
@@ -41,12 +41,12 @@
     <key key0="𐑲"/>
     <key key0="𐑴"/>
     <key key0="𐑰"/>
-    <key key0="𐑚" key2="&lt;" key3="."/>
-    <key key0="𐑝" key2="&gt;" key3=","/>
-    <key key0="𐑟" key2="\?" key3="/"/>
-    <key key0="𐑒" key2=":" key3=";"/>
-    <key key0="𐑢" key2="&quot;" key3="'"/>
-    <key key0="𐑛"/>
+    <key key0="𐑚"/>
+    <key key0="𐑝" key2="&lt;" key3="."/>
+    <key key0="𐑟" key2="&gt;" key3=","/>
+    <key key0="𐑒" key2="\?" key3="/"/>
+    <key key0="𐑢" key2=":" key3=";"/>
+    <key key0="𐑛" key2="&quot;" key3="'"/>
     <key key0="backspace" key2="delete"/>
   </row>
 </keyboard>
--- a/srcs/layouts/shaw_imperial_en.xml
+++ b/srcs/layouts/shaw_imperial_en.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Shaw Imperial layout for shavian alphabet, see https://www.shavian.info/keyboards/ -->
+<keyboard name="Shaw Imperial" script="shavian">
+  <row>
+    <key key0="𐑶" key2="1" key4="esc"/>
+    <key key0="𐑬" key1="~" key2="2" key3="\@"/>
+    <key key0="𐑫" key1="!" key2="3" key3="\#"/>
+    <key key0="𐑜" key2="4" key3="$"/>
+    <key key0="𐑖" key1="𐑠" key2="5" key3="%"/>
+    <key key0="𐑗" key2="6" key3="^"/>
+    <key key0="𐑙" key2="7" key3="&amp;"/>
+    <key key0="𐑘" key2="8" key3="*"/>
+    <key key0="𐑡" key2="9" key3="(" key4=")"/>
+    <key key0="𐑔" key2="0" key3="f11_placeholder" key4="f12_placeholder"/>
+  </row>
+  <row>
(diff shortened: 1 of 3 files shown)


srcs/layouts/shaw_imperial_en.xml
```diff
@@ -0,0 +1,52 @@
...
+    <key key0="𐑜" key2="4" key3="$"/>
+    <key key0="𐑖" key1="𐑠" key2="5" key3="%"/>
+    <key key0="𐑗" key2="6" key3="^"/>
+    <key key0="𐑙" key2="7" key3="&amp;"/>
+    <key key0="𐑘" key2="8" key3="*"/>
+    <key key0="𐑡" key2="9" key3="(" key4=")"/>
+    <key key0="𐑔" key2="0" key3="f11_placeholder" key4="f12_placeholder"/>
+  </row>
+  <row>
+    <key shift="0.5" key0="𐑭" key1="tab" key2="𐑸" key3="`"/>
```
Julow, line 17: The alternating shift avoids having a grid layout but make the keys smaller. What do you think of a grid layout ?

srcs/layouts/shaw_imperial_en.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Shaw Imperial layout for shavian alphabet, see https://www.shavian.info/keyboards/ -->
+<keyboard name="Shaw Imperial" script="shavian">
+  <row>
```
Julow, line 4: 4 rows is a lot and corners are underutilized.
Any way of removing one by moving some characters to the corners of other keys ?
The "Shaw QWERTY" layout manages to have only 3 rows.

srcs/layouts/shaw_imperial_en.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Shaw Imperial layout for shavian alphabet, see https://www.shavian.info/keyboards/ -->
+<keyboard name="Shaw Imperial" script="shavian">
```
Julow, line 3: The numpad and the optional number row (that can be enabled in the settings) can be configured to match the layout's script by modifying this function: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L100

Julow (approved):
Nice! I wrote some suggestions but I would merge once you say it's ready!

srcs/layouts/shaw_imperial_en.xml
la-ninpre: I thought about that, yes. I would argue that it makes it a bit different, but maybe compactness matters more here, I'm fine with making it a grid.

srcs/layouts/shaw_imperial_en.xml
la-ninpre: Yes, I agree, that four rows make it rather tall, but I don't know if it is okay to move a whole row to corners of adjacent row. Shavian just has more letters than Latin. I think this change would make it a different layout... I'll test to see if I can come up with something that works well.

srcs/layouts/shaw_imperial_en.xml
la-ninpre: Hmm, I don't think I understand. If I got it correctly, this function converts regular numbers to their equivalents in a certain script, but shavian doesn't have any special numbers of its own, they stay the same.

srcs/layouts/shaw_imperial_en.xml
Julow: Matching the official layout is also nice! That was just a suggestion, 4 rows is fine if it's the layout that you would use. Same for the alternating shift, choose what you'd prefer using.

srcs/layouts/shaw_imperial_en.xml
Julow: My mistake!

srcs/layouts/shaw_imperial_en.xml
la-ninpre: Well, I came up with something that works okay. I'm thinking this: will it be okay if I add both versions? So it would be like "Shaw Imperial" and "Shaw Imperial (Compact)".

I could add Shaw QWERTY too, btw.

Julow: The layout selector is not ergonomic enough to handle too many layouts so I don't accept small variations and layouts without users.
But feel free to post any number of layout to https://github.com/Julow/Unexpected-Keyboard-layouts, which have a chance to be included in the app when the settings improve.

Julow:
Please use it for a while and then confirm when it's ready for merging.

la-ninpre:
I tested a few different options and for me the current one is the most comfortable to type with. I mean, the four-row version without staggering. You can merge it, if you have no other things to add.

Julow (approved):
Perfect, let's merge!

Of course, feel free to make more modifications to the layout in the future.

PR #594, opened 2024-03-16
Normalization of Korean characters

Julow:
This changes how some characters are sent to the editor. Characters specified in 'should_normalize_char' are not committed immediately to the editor but instead added to the composing text.

The composing text is NFCKC-normalized. Resulting characters that are not in 'should_normalize_char' committed, the rest is the composing text.

Typing a character not specified here do not change: commitText is used immediately.


Julow:
Closed in favor of https://github.com/Julow/Unexpected-Keyboard/pull/595

PR #600, opened 2024-03-25
feat: add a bunch of new themes

byomess:
I made it for my self, but raising this PR just in case you guys are interested


scripts/themes.json
```diff
@@ -0,0 +1,1113 @@
+[
```
Julow, line 1: Why not generating the arrays, etc.. from the `themes.xml` instead of introducing this file ?
The format the same and we already use xml a lot in the project.

srcs/juloo.keyboard2/Config.java
```diff
@@ -385,14 +385,57 @@ private int getThemeId(Resources res, String theme_name)
   {
     switch (theme_name)
```
Julow, line 386: This switch statement needs to go! Linear lookup is the best way, see the layout lookup as an example (it uses a third generated array containing the resources ids) https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/prefs/LayoutsPreference.java#L53

res/values/arrays.xml
```diff
@@ -20,25 +20,111 @@
   </string-array>
   <string-array name="pref_theme_entries">
     <item>@string/pref_theme_e_system</item>
-    <item>@string/pref_theme_e_dark</item>
-    <item>@string/pref_theme_e_light</item>
-    <item>@string/pref_theme_e_black</item>
+    <!-- THEMES_ENTRIES_ARRAY_START -->
```
Julow, line 23: Arrays could be outputted to a new file, to make the generator script more maintainable. This is how it's done for layouts, this file is generated: https://github.com/Julow/Unexpected-Keyboard/blob/master/res/values/layouts.xml

res/values/strings.xml
```diff
@@ -59,14 +59,57 @@ This application contains no ads, doesn't make any network requests and is Open
   <string name="pref_character_size_summary">Size of characters displayed on the keyboard (%.2fx)</string>
   <string name="pref_theme">Theme</string>
   <string name="pref_theme_e_system">System settings</string>
-  <string name="pref_theme_e_dark">Dark</string>
-  <string name="pref_theme_e_light">Light</string>
-  <string name="pref_theme_e_black">Black</string>
+  <!-- THEMES_STRINGS_START -->
```
Julow, line 62: The strings could also be outputted to the new file I suggest above but they need to be processed by `sync_translations.py`. It could be modified to also look at the new file when calculating the `baseline` (the English strings).

scripts/themes.json
```diff
@@ -0,0 +1,1113 @@
...
+    {
+        "name": "AlabasterDark",
+        "display_name": "Alabaster Dark",
+        "parent_theme": "BaseTheme",
+        "is_light_theme": false,
+        "attribs": {
+            "colorKeyboard": "#1f1f1f",
```
Julow, line 8: Where do the themes come from ?

Julow (commented):
I love the idea!

scripts/themes.json
byomess: I just asked ChatGPT to generate all them haha

byomess:
@Julow appreciate your feedback. I'll take care of addressing those changes sometime this week

Julow:
I would really like to merge this someday, without the added themes. This is a nice improvement to the code.

Julow:
I love the idea of generating the themes definitions but I cannot accept the AI generated themes. I'll close this for now but feel free to open a new PR with just the high quality part.

Issue #173, opened 2022-06-29
[Feature Request] Emoji menu search bar


sharkaccino:
Hi :) I've been using this keyboard for a couple days, and i've absolutely fallen in love with it. Getting used to this swiping thing is a bit of a challenge, but I can already tell this will end up being a much better experience for me. Having quick access to all these symbols is HUGE.

That said, I'm coming from Gboard, and one of the features I really miss is being able to search for emojis by text. 

Some screenshots for those who haven't used it before:

![Screenshot_20220629-023801_Simple Notepad_1](https://user-images.githubusercontent.com/31686191/176406294-8d82bd58-15d5-4788-8738-a2ebc9dd7836.png)

![Screenshot_20220629-023813_Simple Notepad_1](https://user-images.githubusercontent.com/31686191/176406299-659c1f66-d84d-457d-9cef-955f65986866.png)

![Screenshot_20220629-023834_Simple Notepad_1](https://user-images.githubusercontent.com/31686191/176406302-2baa0c70-f4b7-4665-a90c-0c945b8abe11.png)

Unfortunately, I'm not really an Android dev, and more importantly, I have no idea how exactly this is done in Gboard. If someone could give me some pointers in that area, I'd gladly make a fork of this project to try and implement it myself for a PR.

Otherwise, I'd love to see this added some time :) Thanks for your work!

tonyaajjackson:
Seconded - this feature would be awesome to have. I remember the name of the emoji way more readily than its position in a menu. I often use the built-in feature in slack or discord of being able to type :horse: and would love to be able to search for emoji by name from the keyboard

TPS:
I'd like to request expanding this to _all_ included characters, since it's not always just emoji that's hard to locate.

Rn, I end up copy/pasting from [UnicodePad](https://github.com/Ryosuke839/UnicodePad), which has _excellent_ search functionality.

Russell-Jones-OxPhys:
> I'd like to request expanding this to _all_ included characters, since it's not always just emoji that's hard to locate.
> 
> Rn, I end up copy/pasting from [UnicodePad](https://github.com/Ryosuke839/UnicodePad), which has _excellent_ search functionality.

Not license compatible, sadly. If there's a GPL 2 or 3 emoji input app, it may be possible to crib from that.

TPS:
> Not license compatible, sadly. If there's a GPL 2 or 3 emoji input app, it may be possible to crib from that.

[...]

PR #612, opened 2024-04-18
Emoji Update

tmqCypher:
Finally got a chance to take a look at updating the emoji backend, apologies for the wait!

I made what seemed to be the most necessary changes talked about in #583:
- Removing unused information (names and descriptions) from the Emoji class
- Creating a Gradle task that generates a more efficient `res/raw/emojis.txt` file from the most recent Unicode standard
- Saving recently used emoji preferences as emoji values rather than names
- Migrating old user preferences to the new system

I kept the Emoji class because it keeps EmojiGridView and EmojiGroupButtonsBar cleaner and keeps space for potential features like #173, which would require more emoji-specific functionality than is offered by KeyValue.
<br>

The code does all of the above as it is now, but as part of the update I'd like to suggest saving recently used preferences as individual string-integer entries rather than a string set. We would no longer have to parse the integers out when the preferences are loaded or redundantly save unchanged values for every keystroke.

If this sounds like a good idea, or if any other preference changes should be made, they should definitely be implemented now to avoid a much more headache-inducing multi-stage migration process in the future.

Let me know what you think!

### DIFF
--- a/srcs/juloo.keyboard2/Emoji.java
+++ b/srcs/juloo.keyboard2/Emoji.java
@@ -7,26 +7,15 @@ import java.io.IOException;
 import java.io.BufferedReader;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 
 public class Emoji
 {
-  private final String _name;
   private final KeyValue _kv;
-  private final String _desc;
 
-  private static HashMap<String, Emoji> emojis_by_name = new HashMap<String, Emoji>();
-
-  protected Emoji(String name, String bytecode, String desc)
+  protected Emoji(String bytecode)
   {
-    _name = name;
-    _kv = new KeyValue(bytecode, KeyValue.Kind.String, 0, 0);
-    _desc = desc;
-    emojis_by_name.put(name, this);
-  }
-
-  public String name()
-  {
-    return _name;
+    this._kv = new KeyValue(bytecode, KeyValue.Kind.String, 0, 0);
   }
 
   public KeyValue kv()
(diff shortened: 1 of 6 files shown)


srcs/juloo.keyboard2/EmojiGridView.java
```diff
@@ -110,6 +111,33 @@ SharedPreferences emojiSharedPreferences()
     return getContext().getSharedPreferences("emoji_last_use", Context.MODE_PRIVATE);
   }
 
+  private void migrateOldPrefs()
```
Julow, line 114: This does some work that is re-done quickly after and loose the result (`emojiSharedPreferences`, `getStringSet`).
What do you think of changing that into a function that returns the last used (eg. `Set<String> loadLastUsedFromPrefs()`) and would be called from `loadLastUsed()` ?
That would make this function the only place where the prefs are loaded, ensuring migration is always checked.

srcs/juloo.keyboard2/EmojiGridView.java
```diff
@@ -110,6 +111,33 @@ SharedPreferences emojiSharedPreferences()
...
+    if (lastUsed != null && !prefs.getBoolean(MIGRATION_CHECK_KEY, false))
+    {
+      SharedPreferences.Editor edit = prefs.edit();
+      edit.clear();
+
+      Set<String> lastUsedNew = new HashSet<>();
+      for (String entry : lastUsed)
+      {
+        String[] data = entry.split("-", 2);
+        lastUsedNew.add(Integer.parseInt(data[0]) + "-" + Emoji.mapOldNameToValue(data[1]));
```
Julow, line 132: `mapOldNameToValue` can throw an exception and it's not fine to crash the app for such a small feature.
I think catching the exception here and not adding the problematic emoji to `lastUsedNew` is a nice way to handle it.

srcs/juloo.keyboard2/Emoji.java
```diff
@@ -7,81 +7,782 @@
...
+      // Read emoji (until empty line)
+      while (!(line = reader.readLine()).isEmpty())
+      {
+        Emoji e = new Emoji(line);
+        _all.add(e);
+        _stringMap.put(line, e);
+      }
 
-  public static int num_groups = 0;
+      // Read group indices
```
Julow, line 50: This format works but I think could be slightly simpler with the group indices on the first line. That way, the list don't have to be delimited.
The code that interprets the indices can still be done after the entire list has been read.

srcs/juloo.keyboard2/Emoji.java
```diff
@@ -7,81 +7,782 @@
...
+  public static int getNumGroups()
+  {
+    return _groups.size();
+  }
 
-  public static Emoji getEmojiByName(String name)
+  public static Emoji[] getEmojisByGroup(int groupIndex)
   {
-    return emojis_by_name.get(name);
+    return _groups.get(groupIndex).toArray(new Emoji[0]);
```
Julow, line 68: The List representation seems ideal for saving memory avoiding copies. Especially that this function is called for every group just to get the first emoji of each groups to make the tabs.
I feel like the rest of the code can be made to work with List with small modification, what do you think ?

Julow (commented):
This is fantastic ! The code is very well done.

The APK do not decrease in size (and in fact increases by 85K), which is probably due to the to-be-removed migration function (the huge switch is 121K).

> The code does all of the above as it is now, but as part of the update I'd like to suggest saving recently used preferences as individual string-integer entries rather than a string set. We would no longer have to parse the integers out when the preferences are loaded or redundantly save unchanged values for every keystroke.

You mean having one StringSet representing the emojis that were used previously, and an extra integer preference for each of them ?
I think that could be a good idea !

`LayoutsPreference` also deals with serializing complex values and it uses [JSONObject](https://developer.android.com/reference/org/json/JSONObject) in order to rely less on `SharedPreference`'s methods.

There's also [getAll()](https://developer.android.com/reference/android/content/SharedPreferences#getAll()) but it's very unsafe and worrying about `LAST_USE_PREF` and `MIGRATION_CHECK_KEY` seems like a step backward.

srcs/juloo.keyboard2/Emoji.java
tmqCypher: Agreed. For the initial draft I wanted to make as few changes outside of Emoji.java as possible, but I can definitely refactor the rest to work with Lists

tmqCypher:
> You mean having one StringSet representing the emojis that were used previously, and an extra integer preference for each of them ?

> There's also [getAll()](https://developer.android.com/reference/android/content/SharedPreferences#getAll()) but it's very unsafe and worrying about `LAST_USE_PREF` and `MIGRATION_CHECK_KEY` seems like a step backward.

My idea was to store only the integer preferences, use `getAll()` to retrieve the emoji-integer pairs, and copy the values to `_lastUsed`. This wouldn't require any de-serialization and would also remove the need for `MIGRATION_CHECK_KEY` since migration status could be determined simply by checking for the old string set with `prefs.contains(LAST_USE_REF)`.

All of the preferences would have a single, known type and nothing would be getting modified, so it doesn't seem unsafe to use `getAll()` in this case. I'm new to Android development though, so I'll trust your judgement if you'd still rather avoid it.

srcs/juloo.keyboard2/EmojiGridView.java
tmqCypher: Whoops, thanks for noticing that. I made it throw for that exact purpose and forgot to actually catch it

Julow:
Sorry for the slow reply. This should not be a draft anymore, I would like to merge it!

> My idea was to store only the integer preferences, use getAll() to retrieve the emoji-integer pairs, and copy the values to _lastUsed.

That works for me :) Though, the PR is working, so I suggest to merge it like this now and make further improvements later.

Julow (approved):
This will make our lives easier! Thanks a lot :)

Julow:
The new preference layout you propose seems good to me but will require yet an other migration function as I want to avoid leaving this PR open for too long. I plan to release soon and need the updated emojis.

PR #617, opened 2024-04-20
Update Simplified Chinese translation

CloneWith:
The "Home" "Compose" "End" strings are intentionally kept, since we usually don't translate it on modern Chinese keyboards.


Julow (approved):
Sorry for the slow reply and thanks a lot :)

PR #625, opened 2024-04-30
add settings button to launcher app

alotbsol555:
not pretty, but it works.


Julow (commented):
This button is very useful but I'm worried that it could make the installation procedure more confusing by being placed next to the other buttons.

What do you think of putting a gear icon in the title bar instead ?

Julow:
This looks great! Please open a PR :)

It's unfortunate that the SDK do not contain a gear icon. Perhaps this one would do ? https://pictogrammers.com/library/mdi/icon/cog-outline/ (in the download menu, choose "Download XML Vector Drawable for Android")

PR #626, opened 2024-04-30
allow to modmap the fn key

alotbsol555:
introduce `<fn ...>` the same way as `<shift ...>` inside of `<modmap>` to map keys for the fn-key via xml without changing the source code.

seems to work straight forward. but maybe [this commit](https://github.com/Julow/Unexpected-Keyboard/commit/3f6b6fd23253e5326d9ba29cac10be36305b3019) broke something even without this pull request:

when building with all current commits in Julow/Unexpected-Keyboard/master, the modmap is only read once on startup of the app. after editing the layout and the modmap xml you have to force quit the app and restart. then everything is up to date. so that is not due to this pull request.

### DIFF
--- a/srcs/juloo.keyboard2/KeyModifier.java
+++ b/srcs/juloo.keyboard2/KeyModifier.java
@@ -158,6 +158,12 @@ public final class KeyModifier
 
   private static KeyValue apply_fn(KeyValue k)
   {
+    if (_modmap != null)
+    {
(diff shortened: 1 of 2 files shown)


Julow (approved):
Awesome :) Thanks!

> maybe this commit broke something

Thanks for the debugging! I fixed that in c7ed2bc

PR #628, opened 2024-05-03
Compass-point synonyms for edge keys in layouts

Spike-from-NH:
Adds eight compass-point synonyms for key1 through key8

### DIFF
--- a/srcs/juloo.keyboard2/KeyboardData.java
+++ b/srcs/juloo.keyboard2/KeyboardData.java
@@ -430,6 +430,7 @@ public final class KeyboardData
       KeyValue[] ks = new KeyValue[9];
       int keysflags = 0;
       keysflags |= parse_key_attr(parser, "key0", ks, 0);
+      /* key1-key8 define swipe gestures (diagram above) */
       keysflags |= parse_key_attr(parser, "key1", ks, 1);
       keysflags |= parse_key_attr(parser, "key2", ks, 2);
       keysflags |= parse_key_attr(parser, "key3", ks, 3);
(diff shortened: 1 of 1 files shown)


srcs/juloo.keyboard2/KeyboardData.java
```diff
@@ -438,6 +439,15 @@ public static Key parse(XmlPullParser parser) throws Exception
       keysflags |= parse_key_attr(parser, "key6", ks, 6);
       keysflags |= parse_key_attr(parser, "key7", ks, 7);
       keysflags |= parse_key_attr(parser, "key8", ks, 8);
+      /* Compass-point synonyms for swipe gestures */
+      keysflags |= parse_key_attr(parser, "nw", ks, 1);
```
Julow, line 443: I like that the compass-points override the key* attributes but there's a bug: if both `key1` and `nw` are set, the key specified by `nw` is used but the flags of both keys are OR-ed.

Here's a suggestion to fix this bug:
Make a function that checks for one attribute then the other and return that attribute's value. Change `parse_key_attr` to receive a key name instead of an attribute name. Here's how it'll be called:

```java
      keysflags |= parse_key_attr(parser, get_key_attr("nw", "key1"), ks, 1);
```

Julow (commented):
Nice idea :)

Spike-from-NH:
I don't see it; the net result in that case seems to be:
 `keysflags |= (F_LOC<<1) | (F_LOC<<1)`
—the same bit gets set twice?

I can try to write (by imitation) a function that takes parameters `attr1` and `attr2`. (Or better, modify parse_key_attr() to take two and behave as before if `attr2` is empty?) (Then need another function if we stumble over a sensible third synonym?)

Julow:
>  —the same bit gets set twice?

Not if the flags are different for the two conflicting keys.

> Or better, modify parse_key_attr() to take two and behave as before if attr2 is empty?

I prefer having smaller functions that combine well rather than a bigger function that takes more arguments. But if you have an implementation in mind, go ahead.

Spike-from-NH:
> Not if the flags are different for the two conflicting keys.

But two synonyms have the same flag! So I'm still not convinced there is a bug. If there is not, just listing synonyms is better than creating a brittle function based on there being exactly two synonyms.

By the way, I don't care which prevails if the user specifies both `nw=` and `key1=`, though ideally, your new pop-up diagnostic should warn that the user may be confused.

Julow:
> But two synonyms have the same flag! So I'm still not convinced there is a bug.

Here's an example of where this bug can show up: Start from the default layout:

```
    <key key0="d" key1="loc accent_grave" key2="loc £" key3="loc accent_aigu"/>
```

Add `nw="¡"` to this key. The added key will not show up.

> If there is not, just listing synonyms is better than creating a brittle function based on there being exactly two synonyms.

The implementation I have in mind is not brittle.

> ideally, your new pop-up diagnostic should warn that the user may be confused.

That would also be a good way to do it! This definitely requires a new function.

Spike-from-NH:
> Add `nw="¡"` to this key. The added key will not show up.

Well, because the definition now includes two contradictory specifications for the northwest swipe (one with a LOC bit and one without). Again, I don't care what results in this case but, yes, a pop-up should warn the user.

How shall we proceed? I finally put a modern browser on a big tablet (so I won't get "Something went wrong!") but I am slow at writing Java, have not studied your diagnostic routine, and have no way to test my code before submitting.

Julow:
Then, an error is definitely the best solution!

Spike-from-NH:
I've made more edits to this file. I saw a warning I had unsaved changes and blew past it. I leave it to @Julow to issue the pop-up error if both synonyms are defined incompatibly.

Julow (approved):
Thanks for this awesome contribution :)

I've implemented the error in case of conflicting attributes and made that a separate function.

Spike-from-NH:
Thank you for correcting, completing, merging, and yet calling it awesome!

PR #640, opened 2024-05-23
Circle and round trip gestures

Julow:
This implements clockwise/anticlockwise circle and round trip gestures inspired by Messagease.

The circle gestures start after a small threshold to avoid making the regular swipe too hard to aim.

The gestures do:

- clockwise circle: Shift variant of the center symbol
- anticlockwise circle: Fn variant of the center symbol
- round trip: Fn variant of the targetted side symbol

PR #644, opened 2024-05-25
Configure anticircle gesture per-key

Julow:
This adds the new 'anticircle' attribute to layouts '<key>' elements that configure the key to send when doing a anti-clockwise circle gesture on it.


Julow:
Follow-up on https://github.com/Julow/Unexpected-Keyboard/pull/640. Feedback welcome

Julow:
Thanks for your review again :) The bug you mentioned is fixed in 3043752.

PR #647, opened 2024-05-29
Add support for Android 12+ dynamic colors

TadaCZE:
This pull request adds three Monet themes for a light, a dark and a system dependent theme. (tested on a real device for both dark and light mode)


Julow (commented):
That's awesome !

The dark theme looks like this on my side:
shot

The side symbols looks too accented and not working together with the center symbols. Also, the "dimmer" symbols color being based on the center symbol color make it looks inconsistent.
What do you think of using the accent color (darkened) as the background of keys instead ?

TadaCZE:
Okay, will try that!

Julow:
I've changed the dark theme as I suggested earlier and it looks like this:

shot

Many thanks for your work :) I'll merge this now, feel free to make improvements with subsequent PRs.

Issue #650, opened 2024-06-01
/doc/*.md typos


ChiefMikeK:
https://github.com/Julow/Unexpected-Keyboard/blob/cf1aaa3bdfcb25388dd0ced888d2ad024fe4a029/doc/Possible-key-values.md?plain=1#L38  
do > `does` not perform editing in response to these keys.

https://github.com/Julow/Unexpected-Keyboard/blob/cf1aaa3bdfcb25388dd0ced888d2ad024fe4a029/doc/Possible-key-values.md?plain=1#L86  
`ᵢ` not displaying/showing any difference here?

Spike-from-NH:
I'll fix these.

Julow:
Fixed in https://github.com/Julow/Unexpected-Keyboard/pull/651

PR #651, opened 2024-06-03
Flesh out custom layout doc; fix typos in #650

Spike-from-NH:


doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
 
-### Modmap
-
-The `<modmap>` tag encloses custom mappings for the Shift and Fn key’s behavior.
+## Row
+The `<row>`...`</row>` pair encloses one row on the keyboard. It has only one optional property:
+* `height`: The height of the row: a positive floating-point value.
 
-Each entry contains two mandatory properties: `a` for the base character before the modifier is applied, and `b` for the modified character.
+A row's standard height is 1.0. The `height` property makes the row taller or shorter than the others. For example, if you define a 5-row keyboard but the top row has `height="0.7"`, then the keyboard's total height is 4.7 units.
```
Julow, line 47:
```suggestion
A row's default height is 1.0. The `height` property makes the row taller or shorter than the others. For example, if you define a 5-row keyboard but the top row has `height="0.7"`, then the keyboard's total height is 4.7 units.
```

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
 
-For example, to make the `I` key behave as in Turkish:
-
-    <modmap>
-        <shift a="i" b="İ" />
-    </modmap>
-
-There can be as many of these tags inside `<modmap>` as needed.
+You set the physical height of the complete keyboard in Settings as a percentage of the total height of the screen, which can be different between portrait and landscape. Unexpected Keyboard scales the "units" so all the rows fit in that alloted space. Thus, defining a shorter row gives space to the other rows.
```
Julow, line 49: That's not true. The percentage set in the setting is only true for a 4 row layout where every rows have a height of `1.0`. Adding or removing a row or changing a row's height will change the keyboard apparent size and will not scale the keys.

The interaction between unusual layouts and this option has never be thought of so I suggest dropping this paragraph.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
 
-Shift and Fn modmaps also affect the clockwise circle and the roundtrip gestures.
+## Key
+The `<key />` tag defines a key on the keyboard. Its position in the sequence of keys inside `<row>`...`</row>` indicates its position in the row from left to right. It must have the following property:
+* `key0`: What the key does when tapped. Unexpected Keyboard provides a legend in the middle of the key's area.
```
Julow, line 53: This property is not mandatory! It's reasonable to have a key without a center symbol, for example a directional pad.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
 key1 | key7 | key2
-:--: | :--: | :--:
+:-: | :--: | :-:
 key5 | key0 | key6
 key3 | key8 | key4
 
-The following properties are optionally supported:
+You can define a swipe only once with either compass-point or numeric notation. Unexpected Keyboard automatically puts a small legend in that direction from the center of the key.
+
+* `slider`: If `slider="true"`, and the key also has `w` and `e` properties, then the key tracks horizontal finger motion precisely and sends the `w` and `e` keystrokes repeatedly. In built-in layouts, this makes the space bar send left and right characters as the user slides on the space bar.
```
Julow, line 75: The internal mechanism is more complicated than that so let's not reveal unnecessary details.
```suggestion
* `slider`: If `slider="true"`, and the key also has `w` and `e` properties, then the key tracks horizontal finger motion precisely and sends the `w` and `e` keystrokes repeatedly. In built-in layouts, this is used to implement the space bar slider.
```

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
-* `indication`: An extra label to show under the main label, intended to be used as a legend for 2A typing (e.g. `<key key0="2" indication="ABC" />`). Caution: if you have `key8` defined, it overlaps!
-* `slider`: If set to `true`, the keys `w` and `e` are sent repeatedly when the key is being slid on. Intended to be used on the space bar, and in fact used on the default space bar.
-* `anticircle`: The key value to send when doing an anti-clockwise circle gesture on the key. The clockwise circle and round-trip gestures are not configurable that way.
+### Extra legend
+* `indication`: An optional extra legend to show under the main label. For example, `<key key0="2" indication="ABC" />` displays ABC at the bottom of the 2 key, as on a pinpad or some telephones. If the key also defines a downward swipe with `s` or `key8`, the legends overlap.
 
 ## Possible key values
+The properties that define the effects of tapping or swiping a key may be one of the built-in strings documented on [this page](Possible-key-values). For example, `se="cut"` says that a southeast swipe produces the "cut" character (Ctrl-C).
```
Julow, line 89: I'd rather not use "character" for talking of anything else than a piece of text containing only one visual character. Internally, the "cut" key is an "editing key" but I'd prefer not to over explain these internal details. "Key" is fine.
```suggestion
The properties that define the effects of tapping or swiping a key may be one of the built-in strings documented on [this page](Possible-key-values). For example, `se="cut"` says that a southeast swipe produces the "cut" key (Ctrl-C).
```

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
-Special values for the keys are documented in [this page](Possible-key-values).
+If the string defining a tap or a swipe is anything other than one of the built-in strings, the defined string is output *verbatim.* This is what most of the taps and swipes on a typical keyboard do. So `key0="a"` simply outputs the letter a.
 
-### `loc ` prefix
+## Modmap
+The `<modmap>`...`</modmap>` pair encloses custom mappings for modifier keys.
 
-Keys prefixed with `loc ` do not appear by default, and are only visible when they are enabled through the "Add keys to keyboard" option in the settings menu, or the language installed on the device is detected to require it.
+A modmap can contain the following tags, each of which must have an `a` and a `b` property:
+* `<shift a="`...`" b="`...`" />` —This says that, if the Shift modifier is on (or the user made an anti-clockwise gesture on a key), and if the key would normally generate the "a" sequence, it must instead generate the  "b" sequence.
```
Julow, line 99: "Sequence" here is unclear, it can only be one key.
```suggestion
* `<shift a="`...`" b="`...`" />` —This says that, if the Shift modifier is on (or the user made a clockwise or round-trip gesture on a key), and if the key would normally send the key specified by "a", it must instead send the key specified by "b".
```
Same below for `fn`.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
+* `<fn a="`...`" b="`...`" />` —This says that, if the Fn modifier is on (or the user made a round-trip gesture on a key), and if the key would normally generate the "a" sequence, it must instead generate the  "b" sequence.
```
Julow, line 100: The gestures are affected by both shift (preferably) and fn (as a fallback if shift did nothing).

Julow (commented):
Thank you for this! I've left many comments but I really appreciate your work and I think the doc looks better now :)

Spike-from-NH:
Thanks for this detailed review! You make good points. The only one I disagree with is the one on the modmap; my modmap maps "the " to "The ". (Perhaps it only affects the "t"?) And I don't understand the final one—Are you saying that `<fn` transforms a key only if no `<shift` did?

It looks like you want me to make suitable edits to my fork before accepting the merge request. Okay, but this process is confusing; where you have made an exact suggestion (I agree with these) and Github is asking me to "Commit suggestion", does that perform an edit on my file? Does the issue then go away, or must I keep track of which issues I have and haven't dealt with?

Spike-from-NH:
Regarding new lines 47 and 49, you seem to be saying that the percentage of the screen height that Settings lets the keyboard occupy is 4.0 units (rather than my guess that you scale row heights to comply with Settings)—so a keyboard with other than 4.0 will have a height other than that specified in Settings. If so, we ought to say so, in order to say what "1.0" means.

Spike-from-NH:
And on your first two comments: I added this Intro to explain why the document exists, by contrasting custom layouts to built-in ones. And it might be read not only by users who have decided to make a custom layout but by users who want to know what their options are. I think the built-in layouts, though not covering a large percentage of languages, will satisfy a large percentage of users, but you're right, the document oughtn't try to talk them out of a custom layout.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
 # Custom layouts
+Unexpected Keyboard includes key layouts for many natural languages. Unexpected starts out with a key layout likely to suit your device's language. You can select a different layout by calling up the Settings page (swipe the gear icon) and, at the top of the page, either tapping an existing layout or tapping _Add an alternate layout_. This displays a menu of available layouts.
```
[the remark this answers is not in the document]
Spike-from-NH: Should be addressed by my revision. I explain the need for such an intro in my follow-on post.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
 
-Unexpected Keyboard allows custom layouts to be defined, loaded, and used in the app. These layouts are defined in XML.
+If no built-in layout is what you want, you can define your own layout by choosing _Custom layout_ at the bottom of this menu. Unexpected Keyboard now displays code in the XML format. You make changes by replacing this with different code and tapping OK.
 
-## Existing keyboard layouts
+Start out in one of these ways:
+* Copypaste the displayed code into a text file (named something like `MyChanges.xml`) to maintain it independently of Unexpected Keyboard.
```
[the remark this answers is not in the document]
Spike-from-NH: Should be resolved by my revision.

doc/Custom-layouts.md
```diff
@@ -1,112 +1,114 @@
...
+* Make a copy of one of the built-in layouts found in [`/srcs/layouts`](https://github.com/Julow/Unexpected-Keyboard/tree/master/srcs/layouts).
+* Use the [web-based editor](https://unexpected-keyboard-layout-editor.lixquid.com/). You can use touch gestures to define keys and swipes and move keys to desired positions, and it will write the XML code for you.
```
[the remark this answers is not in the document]
Spike-from-NH: Should be resolved by my revision (rewording). I mistakenly assumed the web page let you drag keys into position.

doc/Custom-layouts.md
Spike-from-NH: Agreed, a more conventional term.

doc/Custom-layouts.md
Spike-from-NH: Should be resolved by my revision.

doc/Custom-layouts.md
Spike-from-NH: Yes, fewer details on the flow of characters is better.

doc/Custom-layouts.md
Spike-from-NH: Okay. Ctrl-C is a control "character" (or "code") but I agree, reader wants to deal with keys, not bytes.

Spike-from-NH:
I revised the file generally. In the 4 cases you offered wording, I agreed with 3. (I made 2 a "batch" but was never asked to do anything with it.) My remaining confusion is confined to keyboard height and modmap.

Julow:
> The only one I disagree with is the one on the modmap; my modmap maps "the " to "The ". (Perhaps it only affects the "t"?)

That's a single key outputting several characters. The word "sequence" is ambiguous in the sense that's it's not possible to specify a sequence of several keys (eg. it's not possible to specify "selectAll cut").

> Are you saying that <fn transforms a key only if no <shift did?

For the gestures only, yes.

> Github is asking me to "Commit suggestion", does that perform an edit on my file?

Yes, Github will modify the file in your fork and make a commit. If you have the code locally, you'll have to pull.
To simplify things a bit, you can copy paste the suggestions. Or even, take inspiration from them and change the doc in an other way.

> so a keyboard with other than 4.0 will have a height other than that specified in Settings

Yes. But I think it's not necessary to specify that for the reason I mentionned above: That's not intended. No users of 3-rows or 5-rows layouts have complained but this behavior has not been made on purpose.

Spike-from-NH:
The document now seems to include the 2 of your edits I put in the "batch."

Regarding keyboard height, I'll document it as I now understand it. If you feel it still restricts future development options, we can delete it.

I've filed the Pull Request because I can't tell if I missed anything. If you accept it but I have missed something, I'll promptly do another round.

doc/Custom-layouts.md
```diff
@@ -1,113 +1,124 @@
 # Custom layouts
+Unexpected Keyboard includes key layouts for several natural languages. Unexpected starts out with a key layout likely to suit your device's language; if there is none, Unexpected starts with a traditional QWERTY layout. You can select a different layout by calling up the Settings page (swipe the gear icon) and, at the top of the page, either tapping an existing layout or tapping _Add an alternate layout_. This displays a menu of available layouts.
```
Julow, line 2: I really don't think this paragraph is useful and I'd prefer if it was removed.

doc/Custom-layouts.md
```diff
@@ -1,113 +1,124 @@
...
 
-Unexpected Keyboard allows custom layouts to be defined, loaded, and used in the app. These layouts are defined in XML.
+You can define your own layout by choosing _Custom layout_ at the bottom of this menu. Unexpected Keyboard now displays code in the XML format. You make changes by replacing this with different code and tapping OK.
 
-## Existing keyboard layouts
+We recommend you keep your work in a file outside Unexpected Keyboard (named something like `MyChanges.xml`). If you installed a new version of Unexpected from a different website (with a different Build ID), then your Settings, including work you did solely by editing the XML inside Unexpected, would be lost.
```
Julow, line 6:
```suggestion
We recommend you keep your work in a file outside Unexpected Keyboard (named something like `MyChanges.xml`). If you installed a new version of Unexpected from a different website (with a different signature), then the work you did solely by editing the XML inside Unexpected, would be lost.
```

Julow (commented):
The `height` part and the contributing section looks correct. Just a few comments and then I'd merge.

Julow (approved):
Awesome, many thanks !

Issue #655, opened 2024-06-09
modmap in custom keyboard layout


Joppla:
Hello,

Realy nice keyboard. Compliments.

I was experimenting a bit with the custom keyboard. I try to use modmap to use the fn key in other way. But till now it doesn't what I suspect.

Questions
Can I use the modmap tag within the custom keybord?
Where to place the modmap tag?

Thanks for reaction.

JapanYoshi:
Currently, I think `<modmap />` only works with Shift. I wish I could define custom Fn states, or even custom Compose sequences, too.

Julow:
Hi,

The modmap is placed at the same levels as the rows, for example starting from the qwerty US layout:
```
<keyboard name="QWERTY (US)" script="latin">
  <modmap>
    <fn a="~" b="≈"/>
    <fn a="=" b="≠"/>
  </modmap>
  <row>
    <key key0="q" key2="1" key4="loc esc"/>
    <key key0="w" key1="~" key2="2" key3="\@"/>
    <key key0="e" key1="!" key2="3" key3="\#" key4="loc €"/>
    <key key0="r" key2="4" key3="$" key1="loc ₪"/>
    <key key0="t" key2="5" key3="%"/>
    <key key0="y" key2="6" key3="^"/>
    <key key0="u" key2="7" key3="&amp;"/>
    <key key0="i" key2="8" key3="*"/>
    <key key0="o" key1="loc accent_macron" key2="9" key3="(" key4=")"/>
    <key key0="p" key2="0"/>
  </row>
  ... The rest
</keyboard>
```

@JapanYoshi It's possible to map Fn since last release.

Joppla:
Hi,
Thank for reaction.
I thought it should be something like that, but only the original replacement with æ en ₱ for a and b is shown.

This is my code:
[code]
<?xml version="1.0" encoding="utf-8"?>
<keyboard name="QWERTY (US)" script="latin">
  <modmap>
    <fn a="~" b="≈"/>
    <shift a="=" b="≠"/>
  </modmap>
  <row>
    <key key0="q" key2="1" key3="!" key4="esc"/>
    <key key0="w" key1="~" key2="2" key3="\@"/>
    <key key0="e" key1="!" key2="3" key3="\#" key4="loc €"/>
    <key key0="r" key2="4" key3="$" key1="loc ₪"/>
    <key key0="t" key2="5" key3="%"/>
    <key key0="y" key2="6" key3="^"/>
[...]

PR #666, opened 2024-06-13
Modmap: Where to put it; cancelling built-in mods (#655)

Spike-from-NH:


doc/Custom-layouts.md
```diff
@@ -95,23 +95,30 @@ If the string defining a tap or a swipe is anything other than one of the built-
...
-### Example
+### Examples
 Turkish keyboards use the Latin alphabet, but when "i" is shifted, it should produce "İ". This is achieved with the following modmap: 
 
-
     <modmap>
         <shift a="i" b="İ" />
     </modmap>
+    
+On built-in layouts for European locales, `Fn`+`e` produces the € character. This can be inhibited with the following modmap, which maps the modified `e` key to itself:
```
Julow, line 115: That's the case on every locales. On every layouts that have `Fn` and `e`, `Fn+e = €`.

Julow (commented):
Awesome, thanks! As this is documentation, I've left one nick-pick. Otherwise, that's good to go.

doc/Custom-layouts.md
Spike-from-NH: I've deleted the qualification.

Spike-from-NH:
Separately, is this section tight enough?
- I say that "Typically" a layout has only one modmap; should this be mandatory?
- I say the modmap is inside `<keyboard>` but outside any `<row>`. Should it be before any row?
Also: In **Swipes**, the boldfacing and shading of certain of the compass points suggests there is a distinction when there isn't. At most, the center cell should be highlighted. I don't know Markdown well enough to fix this.

Julow:
> I say that "Typically" a layout has only one modmap; should this be mandatory?

Currently, the second modmap would erase the first one. This should be made into an error though and should be described as such.

> I say the modmap is inside <keyboard> but outside any <row>. Should it be before any row?

No constraint on that.

> In Swipes, the boldfacing and shading of certain of the compass points suggests there is a distinction when there isn't.

Table have a mandatory header row and this is a limitation of Github Markdown. Perhaps this can be fixed by writing a raw HTML table ?

Julow (approved):
Thanks! The HTML tables look much better.

Spike-from-NH:
> Thanks! The HTML tables look much better.

But all tables seem to have a default style of shading every other row, which is inappropriate here—It implies there's something special about `w` and `e`.

Julow:
That's fine !

Issue #224, opened 2022-10-21
[Feature Request] split layout, a là Swift Keyboard?


formula-spectre:
title.
![image](https://user-images.githubusercontent.com/61557475/197217961-a2ac0d94-bb05-4e08-baec-ae09d4702ae4.png) 
something like this would be cool, but I understand if this is rejected

moya2162:
I also would like to see this layout as an option.

It would be nice to have it as a toggle option similar to "Show NumPad", where we can have it with the following options:

1.  Never
2.  Only in Landscape Mode
3.  Always

allan-null:
This can already be accomplished with a custom layout.

![Screenshot_20230125_082259_GitHub.png](https://user-images.githubusercontent.com/11065022/214551835-e7f6b18a-0124-488f-b0a0-b9108b50ef96.png)

All you need to do is grab your favorite layout xml from the source-code on the folder `/res/xml/,` add the property `shift="3"` (could be any number) to the key you want to add space to the **left**, and then paste the **xml contents** on the `Custom layout` field under `Advanced options`.
Then, to quickly toggle between split and not split you can setup your secondary layout as the custom one and the primary layout as your favorite one.

**Be warned the repository owner said this is an experimental feature and our custom layouts could break with updates.**

madrabbit90:
Cool! That was easier than I thought it would be! Is it possible to split the bottom row as well?

Julow:
You can optionally re-define the bottom row in your layout, see for example how the Neo2 layout does that: https://github.com/Julow/Unexpected-Keyboard/blob/master/res/xml/latn_neo2.xml#L3
This attribute is set `<keyboard bottom_row="false" ...>` and an extra `<row>` is added. Be warned that the bottom row changes often and you might miss out on new features.

Julow:
Why did you close this issue? This is a commonly requested feature that I would like to have one day.

kingIZZZY:
+1 yes please 🙏

Spike-from-NH:
In #491, I illustrate and give the XML of a layout (Landscape) that has keyboards on each side. I don't claim it is what you want, but do claim you could get what you want from it by imitation, if you work out the horizontal spacing. Reminders:
① Put the XML into effect by copying into clipboard, swiping Unexpected "gear" icon, specifying Custom layout, and pasting everything into the prompt.
② Make very gradual edits and try out, because there is no error reporting yet.

slashdevslashurandom:
[...]

PR #668, opened 2024-06-14
Create latn_qwertz_de_tablet.xml

wamdam:
Added german layout keyboard optimized for tablets in landscape mode. Make sure you also enable numpad for best experience.

### DIFF
--- a/srcs/layouts/latn_qwertz_de_tablet.xml
+++ b/srcs/layouts/latn_qwertz_de_tablet.xml
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This file defines the QWERTY layout.
+
+A layout is made of keys arranged into rows. Keys can be made bigger with the
+'width' attribute and blank space can be added on the left of a key with the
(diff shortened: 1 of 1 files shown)


Julow (commented):
Thanks for your contribution. Unfortunately, the app is not ready to accept variations of layouts as the menu is not ergonomic enough and the maintenance is costly.
Could you add this layout to https://github.com/Julow/Unexpected-Keyboard-layouts so it can be used by others and eventually added to the app later?

My idea for this kind of layouts (variant for tablet or landscape orientation but also for programmers) is to transform the base layouts automatically. The goal is to avoid maintaining all the variations independently and instead to have one definition and a set of rules.

wamdam:
Done, thanks for the great explanation. Fully understood.
https://github.com/Julow/Unexpected-Keyboard-layouts/pull/6

Julow:
I implemented something like this. If you wish to try it, the link is in this comment: https://github.com/Julow/Unexpected-Keyboard/issues/224#issuecomment-4735522256

PR #681, opened 2024-06-29
Clipboard pane

Julow:
This adds the clipboard pane, which allows to save an arbitrary number of clipboards and to paste them later. The key can be disabled in settings.

Checking the "Recently copied text" checkbox will cause the keyboard to keep a temporary history of copied text. This history can only contain 3 elements which expire after 5 minutes. If this is unchecked, only the current system clipboard can be seen.

Ready for testing.


Julow:
Thanks for your review :) That really helped me make the feature better.

The reason for putting the recent entries at the top is that the list of pinned entries is unbounded. Also, it's nice to quickly see whether the keyboard is collecting the history as some people might want to leave the checkbox disabled most of the time.
If it was at the bottom, the workflow for pinning the current clipboard would be to scroll the long list before finding it. The history is bounded in length (3 entries) and in time (5 minutes).
I'd prefer to keep it like that until I gathered more feedback.

I've made the pane a little taller but I like the idea of a handle at the top. This is definitely out of scope of this PR as it's a large task.

The keyboard switching keys have indeed no use and if it was fixed to work, it wouldn't be very intuitive, so I removed it.

I like the idea of an insert button but I'm a bit afraid to add a textbox dialog within the keyboard. Perhaps that would be solved by switching to a full activity but that seems a bit disrupting. Perhaps a new page in the settings. I'll leave that for later as well.

What do you think of it now ?

Julow:
That makes total sense! The remove button had a hidden feature, which is to clear the system clipboard if the current system clipboard is removed from the history. That was not very intuitive and I think would be better as a dedicated button.

Thanks again for your help ! I'll merge this now and do incremental improvements over time.

PR #685, opened 2024-07-02
"loc" prefix not valid in modmap

Spike-from-NH:
Also—
① New **Place-holders** subsection; put last (it is an exception). Give the exception for modmap.
② Partly revert @Julow, **Modmap** readers should refer above, not only to other .md file.


Julow:
I do not want to mention the modmap before it is introduced. Also, this exception do not depict how keys and modmaps works and can only cause confusion.
Here's what I mean: a2eea76

PR #690, opened 2024-07-06
Redesigned icon and support for monochromatic icon.

Pindrop12344321:
I have changed the icon, and not deleted the previous one from the project, instead added new folder for the new "svg" icon.
**Here are the samples**

### DIFF
--- a/res/drawable/ic_launcher_background.xml
+++ b/res/drawable/ic_launcher_background.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
-  <solid android:color="#ff101010"/>
+  <solid android:color="#220059"/>
 </shape>
--- a/res/drawable/ic_launcher_foreground_new.xml
+++ b/res/drawable/ic_launcher_foreground_new.xml
@@ -1,0 +1,180 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="1280dp"
+    android:height="1280dp"
+    android:viewportWidth="2000"
+    android:viewportHeight="2000">
+  <path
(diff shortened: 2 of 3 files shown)


Julow (commented):
Thanks! I think I'd prefer if the old icon was gone and replaced with an image version of the new icon (for compatibility). The old icon was made quickly and doesn't look very good.

ChasmSolacer:
The old icon shows these small chars in key corners, I don't remember if I noticed this app because of this, but I think it quickly shows off the keyboard's main feature.
I'd rather insert random symbols in some corners and maybe remove the `_` key and enlarge the other keys a bit to make them visible. What do you think?

Pindrop12344321:
You are right 👍

ChasmSolacer:
Honestly idk, if it needs replacement, maybe a key that doesn't appear on most keyboards like arrows or Fn.
Otherwise, the icon you propose looks elegant. That's just a random remark

Pindrop12344321:
> Honestly idk, if it needs replacement, maybe a key that doesn't appear on most keyboards like arrows or Fn. Otherwise, the icon you propose looks elegant. That's just a random remark

** I think that was a great remark **, going to replace the underscore

Julow:
I agree with @ChasmSolacer that the icon should show symbols in the corner of keys.

ValerieLehane:
I would think that, the name being "unexpected keyboard," you'd want to transmit the idea "this keyboard does more than you'd expect, and differently than you'd expect it to."

Therefore, I would think you'd want to feature absolutely off the wall, random-seeming keys (like an interrobang!), perhaps displaying some of the categories of keys that are supported (math, boxes, arrow symbols, arrow keys, home/end/pgup/pgdn, composeables), in addition to the directional swipe features.

I'm stuck on Android because I'm poor -- that's how I got here! -- but I'll try to produce a concept attempt, in Krita, the best I can. I don't understand how to use the path tools, with only a touch screen, so I'll try to keep the raster traceable for SVG, even at whatever the minimum icon size is, which I'll try to research, but maybe someone here can tell me what to aim for.

Or, if you're pretty sure what you want to go for, I can stand down and save my time. :) I just happened by, and saw something I figured I can present a unique perspective on (but don't we all have one of those?), and felt like attempting art. I will NOT be offended if you're okay with what you have, though! ❤️

ValerieLehane:
Well, nevermind my dreams of artistic fame through icon design, lol. Krita is crashing. :( Life is pretty difficult with Android 11 and 2GB RAM. I'm sure y'all have got this, anyway.

PR #694, opened 2024-07-07
Update values-de/strings.xml

polyctena:
Translated new strings for clipboard, long press timeout/key repeat, Monet theme.


Julow (approved):
Thanks !

PR #695, opened 2024-07-07
New Spanish translations

Spike-from-NH:
Using clipboard → portapapeles
Using gesture → gesto
If Android uses different terms, it's preferable to align with those


Julow (approved):
Thanks!

PR #696, opened 2024-07-08
Clerical edits related to locale_extra_keys

Spike-from-NH:


Spike-from-NH:
Separately, the intro Custom-layouts.md should contain everything the reader needs to get started. So I continue to believe the general rule on values that are **not** built-in strings should be moved or copied back from Possible-key-values.md. Whereas the `loc ` prefix is obscure enough to be moved to Possible-key-values.md.

Julow:
> Separately, the intro Custom-layouts.md should contain everything the reader needs to get started. So I continue to believe the general rule on values that are not built-in strings should be moved or copied back from Possible-key-values.md. Whereas the loc prefix is obscure enough to be moved to Possible-key-values.md.

You could add a mention in Custom-layouts.md if you think that's better.
The `loc` prefix cannot be moved into Possible-key-values.md, I've said that already.

Spike-from-NH:
> The `loc` prefix cannot be moved into Possible-key-values.md, I've said that already.

`loc` is not just another built-in string, but I wish you'd reconsider from a reader's point of view. This paragraph on `loc` is confusing where we have it, but it relates directly to the material on the other page.

PR #709, opened 2024-07-19
Change width of backspace in persian layout and change the place of چ

RZHSSNZDH:


Julow (commented):
I think you also want to add `shift="0.25"` on the first key of the row you changed.
It now looks like this:
shot

You can copy and paste this as a custom layout in the app to test it.

RZHSSNZDH:
Yes you are right. I try to fix it.

Issue #438, opened 2023-08-19
Add keyboard combinations like Ctrl+A/Z/X/C/V for russian layout


vort3:
On any modern PC with Russian keyboard layout, letter «я» is the same key as the letter «z» in English. The list goes like this:

- A — Ф
- Z — Я
- X — Ч
- C — С (note that this is cyrillic «С», not English «C»)
- V — М

Is it possible to make all Ctrl key combinations work with keys in Russian layout as if the layout was English, so pressing Ctrl + Ч would be the cut hotkey, same as Ctrl + X in English?
Currently I have to switch to English layout every time I write something in Russian and want to copy/paste/cut/undo/select all, which is very annoying.

Maybe if it's an easy change to make myself I can help with that, but please point me in the right direction.

Julow:
This can be implemented by adding more cases to this function: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L387

A related issue is open: https://github.com/Julow/Unexpected-Keyboard/issues/109

Julow:
The solution to this issue is to add a `<modmap>` in all the layouts that need it, as discussed in https://github.com/Julow/Unexpected-Keyboard/issues/721. Could you provide the mappings ?

Julow:
Here's how it's done for Serbian: https://github.com/Julow/Unexpected-Keyboard/pull/727

Julow:
It's now possible to implement this without having latin letters appear on the layout (700ec23). Here's how it's implemented in the Serbian layout: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/layouts/cyrl_lynyertz_sr.xml#L11-L37
@bokidori Do you know the mappings for the Russian layout ?

Issue #109, opened 2022-03-21
[feature request] switching from russian to qwerty on ctrl/alt/meta


slashdevslashurandom:
I really like this keyboard app. It is easy enough to use and provides me with all the keys needed for apps like Termux.

Having said that, one annoying feature I have noticed is that, when using the Russian keyboard layout, the Ctrl/Alt/Meta keys are kind of useless, since the combinations still expect a Latin key to go along with them.

I think that, when a non-Latin layout is being used, whenever any of the aforementioned keys are active, the rest of the keyboard should temporarily switch to a QWERTY layout.

Julow:
Hi! Your suggestion is interesting but I hoped to come up with a more general solution to switching between different layouts.

My last attempt is to add an button for switching to a "programming" layout and back (aa78229).
I believe this is not a satisfying solution but I'd like to release soon, so I'll include it so you have at least something. It'll be disabled by default, you'll have to enable it in the keyboard's settings.

I delay fixing this issue to the next-next release.

Julow:
This issue will be solved with per-layout custom Ctrl mappings and not by switching layouts. I'll close this in favor of https://github.com/Julow/Unexpected-Keyboard/issues/438 and https://github.com/Julow/Unexpected-Keyboard/issues/721

PR #718, opened 2024-07-25
Add Serbian Cyrillic layout

bokidori:
Adds Serbian Cyrillic layout. Modmap of Fn key allows for desktop-like easier access to actions such as selectAll, copy, paste, etc. This is because, unlike desktop systems, on mobile ctrl does not combine with cyrillic keys to perform actions.

### DIFF
--- a/check_layout.output
+++ b/check_layout.output
@@ -29,6 +29,9 @@ Layout includes some ASCII punctuation but not all, missing: $, &, *, ., /, <, >
 0 warnings
 # cyrl_jcuken_uk
 0 warnings
+# cyrl_lynyertz_sr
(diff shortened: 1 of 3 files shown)


Julow (approved):
Thanks !

I'm not too fan of adding editing key mappings in layouts. I think the ctrl shortcut problem could be solved in the keyboard instead, see https://github.com/Julow/Unexpected-Keyboard/issues/109 https://github.com/Julow/Unexpected-Keyboard/issues/438

bokidori:
Agreed! Now that ctrl is supported, I will update the layout and do another PR tomorrow. Thank you

Issue #437, opened 2023-08-19
Acute accent with Russian vowels


vort3:
This might be a bug or just me totally misunderstanding how keyboards work.

I have a Russian ЙЦУКЕН layout. I went to the options and enabled the ◌́ symbol, which I believe to be an «Acute accent» symbol. When I hold (enable) this key, I see some of my letters have this ( ́  ) symbol, which is what we call a «stress mark» in Russian. You probably know that already but symbols like á are not real Russian letters, they are not a part of an alphabet: they're just a letter «a» with a stress mark on top.

The problem I have is that I can only letters «к» and «г» with a stress mark. However, in Russian we use stress mark to show which vowels are stressed in certain words, so only vowels can be stressed (у, е, а, о, и). Some vowels are always stressed so they can't have a stress mark, and consonants can't be «stressed» at all.

As far as I know there are two ways to make «stressed letters»:
1. There are single symbols for stressed letters. Those only exist for languages where there are real letters with acute symbol on top of them in alphabet.
2. «Combined» accent symbols, which allow to create combinations of any base letter + any accent symbol, even if they don't exist in any language.

The way I do that on my PC is I enter the base letter (for example, t) then any accent symbol (for example, ̂ ) and create letters like t̂, so, the second method. The way I do Russian «stressed vowels» is the same, so а́, е́, и́ and so on.

What does your ◌́ symbol on keyboard do, how do I combine it with letters? Or does it use the first method and replace letters on keyboard with their stressed variants? If yes, then for Russian layout it's wrong, because letters «к» and «г» don't have «accented» variants. Vowels do. But in Unexpected keyboard this doesn't work properly.

Can I please have some explanation if I'm doing something wrong?

Spike-from-NH:
This is correct. Moreover, if a user has a Russian keyboard and tries to accent the Russian "a" (which is a different Unicode from the English/Spanish "a", though the glyph is identical), the result should be the Russian "а́", not the Spanish one.

Julow:
The symbols modified by the acute accents are defined here: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L503
Symbols not defined here go through the `KeyCharacterMap.getDeadChar` function, which is incomplete. This problem was reported for some diacritics in the past: https://github.com/Julow/Unexpected-Keyboard/issues/375
Russian stressed letters should be added. PR welcome.

[...]

PR #719, opened 2024-07-25
Update accent_circonflexe.json

bokidori:
Allow combining circumflex \u0302 for Cyrillic vowels, enabling users to express themselves accurately.

ја сам са̂м = I am alone
ја сам сам = I am am

да = yes
да̂ = give

код = at
ко̂д = code

Similar to #437 for aigo.


Julow (approved):
Thanks!

PR #720, opened 2024-07-25
Add Serbian Latin QWERTY layout

bokidori:
Adds Serbian Latin QWERTY layout. Serbian language uses both Latin and Cyrillic as its official scripts, and this variant should cover anyone from Serbo-Croatian speaking group that uses Latin script.

### DIFF
--- a/srcs/layouts/latn_qwerty_sr.xml
+++ b/srcs/layouts/latn_qwerty_sr.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard name="QWERTY (Srpski, latinica)" script="latin">
+  <modmap>
+    <fn a="a" b="â" />
+    <fn a="o" b="ô" />
+    <fn a="e" b="æ" />
+    <fn a="€" b="œ" />
+    <fn a="cursor_left" b="home" />
+    <fn a="cursor_right" b="end" />
+  </modmap>
+  <row>
+    <key key0="q" ne="1" se="loc esc"/>
+    <key key0="w" nw="~" ne="2" sw="\@"/>
+    <key key0="e" ne="3" sw="\#" se="€"/>
+    <key key0="r" ne="4" sw="$" se="loc £"/>
+    <key key0="t" ne="5" sw="%"/>
+    <key key0="y" ne="6" sw="^"/>
+    <key key0="u" ne="7" sw="&amp;"/>
+    <key key0="i" ne="8" sw="*"/>
+    <key key0="o" ne="9" sw="(" se=")"/>
+    <key key0="p" ne="0" sw="[" se="]"/>
+  </row>
+  <row>
+    <key key0="a" nw="loc tab" ne="loc selectAll"/>
+    <key key0="s" nw="loc §" ne="š"/>
+    <key key0="d" ne="đ"/>
+    <key key0="f"/>
+    <key key0="g"/>
+    <key key0="h" ne="loc accent_circonflexe" sw="{" se="}"/>
+    <key key0="j" ne="-" sw="_"/>
+    <key key0="k" ne="=" sw="+"/>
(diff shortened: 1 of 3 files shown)


srcs/layouts/latn_qwerty_sr.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard name="QWERTY (Srpski, latinica)" script="latin">
+  <modmap>
+    <fn a="a" b="â" />
+    <fn a="o" b="ô" />
+    <fn a="e" b="æ" />
+    <fn a="€" b="œ" />
+    <fn a="cursor_left" b="home" />
+    <fn a="cursor_right" b="end" />
```
Julow, line 9: Like in https://github.com/Julow/Unexpected-Keyboard/pull/718, these two mappings are your personal preference and they make these keys work differently than in other layouts. It's fine for now but might cause problems in the future as it's a rare occurrence.

Why do you need two home and end keys on your layout ?

Julow (approved):

srcs/layouts/latn_qwerty_sr.xml
bokidori: tl;dr: home on spacebar easier to access; end on spacebar added for left-right consistency; keys on (compose) button of course kept as this is standard for the app.

longer:
Yes, these are indeed just a personal preference, as they sped up some actions for me personally. It is not that I need 2 of each, it is just that the left (home) on the spacebar I find easier to access, and on the (compose) button they are standard for this app so of course I wouldn't touch them.

With Latin keyboard, I'm often writing or editing texts, so Shift+Fn+swipe-left-on-spacebar is a very, very quick way to select a line of text and then either delete, cut, or do what I needed with it. Finger moves less than 2cm, vs ~5cm.

Similar with Cyrillic: I'm bad when composing personal messages and they often come out a bit off at first. Then I'm jumping back to change how I'm addressing a person, or to change it to plural form, etc. Fn+swipe-left-on-spacebar is a ~1cm move that gets me to the start, versus moving a finger from left side of the screen (Fn) to right side (Compose) ~ 4cm.

I'm not advocating this should be standard for any other layouts, but for me it speeds up some actions. If it is fine for now, I would indeed prefer to keep it this way. Although I'm open to change - the layouts are primarily aimed for people from my speaking regions, not for me personally.

Julow: I think that's a good use case. I'm also considering removing the regular arrow keys in exchange for the "sliding" ones but I didn't manage to make them behave intuitively enough yet.
Could you add that as a standard for the whole app ? That happens in [KeyModifier](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L196), the cursor moving keys are of [kind `Cursor_move`](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L348).

bokidori: >Could you add that as a standard for the whole app ? That happens in [KeyModifier](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyModifier.java#L196), the cursor moving keys are of [kind Cursor_move](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L348).

Sorry for the delay, I haven't done any real coding since 2016, and even then - never in Java. Github also being new ground, and life tasks demanding attention didn't help speed things up. That said, this was fun! 😄

I added Fn to cursor_left and cursor_right, turning them into home and end keys respectively. PR coming.

Julow: That PR was perfect, thanks very much!

Issue #707, opened 2024-07-19
the ˇ key moved


adrian4096:
Hi,
the custom added keys should be moved as rarely as possible, as I've developed muscle memory for their locations.
If the answer is to start depending more on the compose key, please fix ignoring the first capital, so "<compose>Vs" should produce "Š" (<compose>vS). Not necesarilly always, just for first capital is enough.

Julow:
I agree that custom keys should move as rarely as possible. Their position was never guaranteed and was stable until now purely by chance. Adding new keys to the keyboard or changing the layouts would cause them to move, and both of them happen out of your control.
This is what I tried to fix in 66cf64357484b227acbb58f9e687a987ad82a1b0, which made some custom added keys more stable but it's still not perfect and doesn't apply to arbitrary custom keys.
This could be improved further but will never be perfect as the app and the layouts continue to evolve.

I just added a way to easily add more compose sequences in https://github.com/Julow/Unexpected-Keyboard/pull/713. Could you add all the other caron accent sequences into `srcs/compose/compose/extra.json` ?

adrian4096:
To be percise: would it be possible, to delay the auto caps? Would it be possible, to set the caps flag on (due to this being at the start of a line), click compose, the caps being delayed for the next keypress? This wouldn't work if any sequence requires first a captial letter and then the lowercase, then a more complicated description of which key consumes caps would be needed.

adrian4096:
Just skimmed over `en_US_UTF_8_Compose.pre` this might get complicated

Julow:
The file in which to add compose sequences is [extra.json](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/compose/extra.json).

You are right, auto-caps should apply to the result of the sequence, not during the sequence.

Julow:
If you provide me with the list of sequences to add, I can add them to the app.

adrian4096:
Sorry for the delay, real life stuff got in the way, ill submut a pr first thing tommorow

Julow:
Pressing compose should now clear the automated Shift (2309792). I think we can close this issue.

PR #725, opened 2024-07-27
Added supported key for avoiding the pain of auto-caps

adrian4096:
I think only these 3 are needed for most of balkan. Since Đ is made from <minus> and D, we needn't worry about capital `-`, and same for Ć being made from `'` and C.
This starts to remove the concerns in #707


Julow (approved):
Thanks!

PR #727, opened 2024-07-27
Update cyrl_lynyertz_sr.xml

bokidori:
Include comprehensive ctrl modmap, enabling the use of ctrl modifier with Serbian Cyrillic keys. Remove clunky fn modmap previously used for just a few actions such as cut/copy/paste/etc.

(I don't know why diff is showing also the closing &lt;/keyboard&gt; tag as modified; I just used github's "in place" web editor to edit the modmap)

### DIFF
--- a/srcs/layouts/cyrl_lynyertz_sr.xml
+++ b/srcs/layouts/cyrl_lynyertz_sr.xml
@@ -6,14 +6,34 @@
     <fn a="и" b="и̂" />
     <fn a="о" b="о̂" />
     <fn a="у" b="у̂" />
-    <fn a="ж" b="selectAll" />
-    <fn a="&lt;" b="undo" />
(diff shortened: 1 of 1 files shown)


srcs/layouts/cyrl_lynyertz_sr.xml
```diff
@@ -53,4 +73,4 @@
     <key key0="ђ" ne=":" sw="."/>
     <key width="1.5" key0="backspace" ne="delete"/>
   </row>
-</keyboard>
\ No newline at end of file
+</keyboard>
```
Julow, line 76: This is due to the line ending in a newline character (which is the standard), while the previous version didn't.

Julow (approved):
Perfect, thanks!

PR #728, opened 2024-07-28
Update KeyModifier.java, Fn for Cursor_move

bokidori:
Fn modifier switches cursor_left to home key, cursor_right to end key.


Julow (approved):
Thanks!

PR #730, opened 2024-07-31
Modmap: Reorganize

Spike-from-NH:
Put new example in Examples section; put all overview material at start; use "modmap" for the modmap tag but "mapping" for the various subtags; set out each subtag separately; stop illustrating example mappings as complete modmap.

### DIFF
--- a/doc/Custom-layouts.md
+++ b/doc/Custom-layouts.md
@@ -104,32 +104,39 @@ Built-in strings that assign a special function to a key are described in [this 
 In a layout, a key value can also start with the `loc` prefix. These are place-holders; the tap or swipe does nothing unless enabled through the "Add keys to keyboard" option in the Settings menu, or implicitly enabled by the language the device is set to use. For example, `ne="loc accent_aigu"` says that a northeast swipe produces the acute accent combinatorial key—if enabled.
 
 ## Modmap
-The `<modmap>`...`</modmap>` pair encloses custom mappings for modifier keys. The modmap is placed inside the `<keyboard>`...`</keyboard>` pair, but outside any row. A layout can have at most one modmap.
+The `<modmap>`...`</modmap>` pair encloses custom mappings for modifier keys. The modmap is placed inside the `<keyboard>`...`</keyboard>` pair, but outside any row. A layout can have at most one modmap. It can contain any number of mappings. Each mapping has an `a` property and a `b` property and directs Unexpected to map the `a` string to the `b` string. Valid values for these strings are listed in [Possible key values](Possible-key-values.md).
 
-A `<modmap>` can contain any number of elements of the form:
+Unexpected supports the following mappings:
+
 ```xml
   <shift a="before" b="after" />
 ```
 This means that when the Shift modifier is on, the key `before` is changed into `after`.
-Valid values for `before` and `after` are listed in [Possible key values](Possible-key-values.md).
-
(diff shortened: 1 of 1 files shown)


doc/Custom-layouts.md
```diff
@@ -104,32 +104,39 @@ Built-in strings that assign a special function to a key are described in [this
 In a layout, a key value can also start with the `loc` prefix. These are place-holders; the tap or swipe does nothing unless enabled through the "Add keys to keyboard" option in the Settings menu, or implicitly enabled by the language the device is set to use. For example, `ne="loc accent_aigu"` says that a northeast swipe produces the acute accent combinatorial key—if enabled.
 
 ## Modmap
-The `<modmap>`...`</modmap>` pair encloses custom mappings for modifier keys. The modmap is placed inside the `<keyboard>`...`</keyboard>` pair, but outside any row. A layout can have at most one modmap.
+The `<modmap>`...`</modmap>` pair encloses custom mappings for modifier keys. The modmap is placed inside the `<keyboard>`...`</keyboard>` pair, but outside any row. A layout can have at most one modmap. It can contain any number of mappings. Each mapping has an `a` property and a `b` property and directs Unexpected to map the `a` string to the `b` string. Valid values for these strings are listed in [Possible key values](Possible-key-values.md).
```
Julow, line 107: I don't like to call a and b "strings". I'd prefer "key".

doc/Custom-layouts.md
```diff
@@ -104,32 +104,39 @@ Built-in strings that assign a special function to a key are described in [this
...
+
+Unexpected supports the following mappings:
```
Julow, line 109: I don't like to name the keyboard here. Same on the paragraph above.
```suggestion
The following mappings are supported:
```

Julow (commented):
Nice examples! I have a couple comments.

Spike-from-NH:
Accepted in full, but with reservations: ① "mappings are supported" is a passive, a vague way of speaking that hides the actor. I named "Unexpected" to use the active voice. If this product name is awkward in prose, "UK" is too; "Great Britain did it?" ② A "string" is what the user types, even if you don't store them as strings. Documentation is aimed at the user.

Spike-from-NH:
Separately, consider the paragraph just before **Modmap/Examples** (on precedence of Shift vs. Fn mappings.

1. They should be mentioned as the mappings are presented. ("This means that when the Shift modifier is on _(including when the user makes the clockwise-circle gesture)_,")
2. I think the case that gives rise to this paragraph is when both Shift and Fn modifiers are on—not if there are multiple mappings.
3. This paragraph is saying that Shift is handled to the exclusion of Fn, right?
4. "and if that did not modify the key"—Do you really test whether the Shift mapping is a no-op? Or merely if one is defined?
5. Do the rules change now that there are three modifiers for which the user can write mappings?

Julow (approved):
I think you are trying to specify too much. The clockwise and round trip gestures are unrelated to custom layouts and if there were documented in an other document (that does not exist), they would be in terms that make this unambiguous.

Mentioning this here is a nice touch but it's counter productive to add more.

> Do you really test whether the Shift mapping is a no-op?

Yes but I don't want that to be explained in this document.

I think this PR is perfect in its current state.

Julow:
Sorry for the slow reply, I'll merge now.

Spike-from-NH:
Thanks! I do try to write complete descriptions, and it is always valid to reply that you want the option of changing the implementation without being bound by them.

Issue #705, opened 2024-07-18
keystroke


LVMuftu:
"Hello, I have a question. How can I write two parentheses characters and move the cursor between them with a single keystroke? Thank you."

Julow:
Hi, that's currently not possible but would be a great addition to the app. Related issues: https://github.com/Julow/Unexpected-Keyboard/issues/568 https://github.com/Julow/Unexpected-Keyboard/issues/576

LVMuftu:
Thanks a lot

Y-T-G:
I wanted to do the same thing

Issue #576, opened 2024-02-26
Combination key like (Select all) + (Paste) in one


k1915361:
Is it possible to make a custom combination key such as (Select all) + (Paste)? I tried adding `(Select all)` or `Select all` but it pastes the literal string instead of activating the key. Any response is appreciated, thanks.

Spike-from-NH:
See the feature request in #568. If this feature were provided, you could have a key or gesture generate `Ctrl` `A` `Ctrl` `V` and give it a suitable legend.

I'm afraid the magic word is in CamelCase: `key4="selectAll"`. But I think this is an exact match, and appending `paste`, you would again see the characters rather than achieve a select-all-and-paste. However, you can easily do it with two gestures.

Perhaps @Julow could tweak the algorithm so that, after matching on `selectAll` etc., it continues to parse any remaining characters of the string.

Julow:
All of this is not possible right now but would be a good addition.

The problem is not the parsing but the data representation.

The representation of keys is currently minimalist (on purpose): 2 final fields for an object size of 3 words: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L130-L133
This has several advantages: `KeyValue`s are used as keys in hashtbls, same keys are structurally equal, memory usage is very low.
I have not heard a suggestion yet but this is absolutely possible to implement while keeping these advantages.

This could indeed be considered a duplicate of https://github.com/Julow/Unexpected-Keyboard/issues/568 but I'd leave the two issues open as the use case is slightly different.

AnabasisXu:
Hi @Spike-from-NH , could you kindly describe how to setup a custom key for Alt+x? Having that would be a game changer for using Emacs on Android.

Spike-from-NH:
I don't know the answer. There is a `<modmap>` tag for redefining Shift and Fn, but not yet Alt.

But maybe you have it backward. Instead of saying, I need this combination to do something, you should be saying, I need to generate these codes—then assign a key or swipe to do so.

Julow:
It's not possible to add a "Alt+x" key on the keyboard currently. Here's a related issue: https://github.com/Julow/Unexpected-Keyboard/issues/568

Julow:
https://github.com/Julow/Unexpected-Keyboard/pull/878 is merged 🎉

You can now add these keys in the settings or in your custom layouts:

- `Vp:selectAll,paste`
- `Ax:alt,x`

[...]

Issue #568, opened 2024-02-21
How to add ctrl+b (tmux prefix) in custom layout?


qw457812:
Hello, 
Thanks for the great keyboard, how can I add combination keys (modifiers + regular key) in custom layout like `<key key0="ctrl+b"/>` (and display as "tmux")?

Spike-from-NH:
That's a reasonable request. The quoted string can include multiple keystrokes (without `+`) but I don't know of a way to specify modified characters. As for the legend, @Julow has said he wants to keep layout separate from appearance, suggesting that if he provides this feature, it would not be with new syntax in the layout file.

Julow:
This is a good idea! A key is currently represented as an object with two fields: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L130-L133, which makes a 3 words object.
Adding anything to this object will add a word to it. This is not a bad deal but we must make sure not to add too much and it might requires some wider changes.

> As for the legend, @Julow has said he wants to keep layout separate from appearance, suggesting that if he provides this feature, it would not be with new syntax in the layout file.

You are mixing appearance of symbols (which is a property of the symbol) and appearance of the keyboard (which is defined in a theme that is separate from layouts).

Spike-from-NH:
Pleased to be corrected! The KeyValue.java piece seems to be an existing variable that holds the keycap legend; all you need is XML syntax to specify a custom one.

I asked for custom legends in an off-topic comment in #483. (What OP @ghost requested there is already doable; also, now, the Compose feature.)

Julow:
https://github.com/Julow/Unexpected-Keyboard/pull/878 is merged 🎉

This can be written as `C-b:ctrl,b`.

The letters before the `:` character is the symbol that will be visible on the keyboard. It can be any [unicode symbol](https://symbl.cc/).

PR #733, opened 2024-08-13
custom key labels in xml layout

alotbsol555:
allow for defining optional custom key labels inside the <key> element of xml layouts using key0_label and key1_label...key8_label or their synonyms nw_label...s_label.

currently, this is mostly useful for keys with multiple characters, since the automatic label might be too long on key0 or not unique on key1...key8. the custom labels might also be useful as a workaround on key1...key8, where automatic labels are cut off after 3 characters instead of 3 visible symbols, which can lead to strange results with emojis or other unicode combinations. in the future, custom labels might be neccessary, when key combinations/sequences or a "goto layout" button should be implemented.

the labels do not interfere with the functionality of the keys. for example, the custom label is not changed by modifiers, even if the underlying key still is. on key0 custom labels longer than 1 character are drawn using the reduced font size in the same way as automatic labels. on key1...key8 custom labels are not truncated after 3 characters. you can define the label in the way it fits your layout.

### DIFF
--- a/srcs/juloo.keyboard2/Keyboard2View.java
+++ b/srcs/juloo.keyboard2/Keyboard2View.java
@@ -331,11 +331,11 @@ public class Keyboard2View extends View
         boolean isKeyDown = _pointers.isKeyDown(k);
         drawKeyFrame(canvas, x, y, keyW, keyH, isKeyDown);
         if (k.keys[0] != null)
-          drawLabel(canvas, k.keys[0], keyW / 2f + x, y, keyH, isKeyDown);
+          drawLabel(canvas, k.keys[0], k.keyslabels[0], keyW / 2f + x, y, keyH, isKeyDown);
         for (int i = 1; i < 9; i++)
         {
           if (k.keys[i] != null)
-            drawSubLabel(canvas, k.keys[i], x, y, keyW, keyH, i, isKeyDown);
+            drawSubLabel(canvas, k.keys[i], k.keyslabels[i], x, y, keyW, keyH, i, isKeyDown);
         }
         drawIndication(canvas, k, x, y, keyW, keyH);
         x += _keyWidth * k.width;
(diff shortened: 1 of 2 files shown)


Julow (commented):
That's a very nice feature! But I think the approach is not ideal.
The symbol is more a property of the key than a property of the layout. For example, it would be useful to specify a custom label in a `<modmap>` or in a custom key in the settings.

`KeyValue` holds a symbol, but that's also the outputted string in case the key outputs more than one character. Perhaps we'd need to make the `KeyValue` class bigger ?
Other issues require the `KeyValue` class to hold both a custom symbol and something more complex than currently: https://github.com/Julow/Unexpected-Keyboard/issues/576 https://github.com/Julow/Unexpected-Keyboard/issues/568 https://github.com/Julow/Unexpected-Keyboard/issues/705
I have the feeling that custom labels is an important part of these issues.

What do you think ?

Julow:
Defining keys elsewhere with a name is also not ideal for the reasons you mentionned. The option I have in mind do not require naming keys.

`KeyValue` has to be changed as this is what is at the same time in layouts, is the output of a modmap and can be added in the settings.
For example, with this PR, the label do not change if the key that is behind it is changed by a modifier.
The issues I linked to also requires this so it's possible to come up with a solution that makes everything easier to implement.

Here's the solution I suggest:

A `KeyValue` should carry its label (it already does in some cases) and it should be possible to specify the label of a key in layouts and other places.
Key attributes need to be made more expressive, for example:

- `:` is the `:` character
- `str:foobar:foo` is the key outputting `foobar` with label `foo`
- `str:'foo:bar':'f:b'` is the key outputting `foo:bar` with label `f:b`

This should implement labels but is also a good base for the other features (which might require the syntax to change).

> custom key sequences (eg. "select all, copy" or "(, ), arrow_left" and key combinations (ctrl+c or ctrl+end) are technically equal? in the sense that combinations are just sequences that start with modifiers

Not really. When a modifier is activated, the other keys on the keyboard are changed but they still are `KeyValue`. That class can represent a lot of different things but do not allow currently to represent sequences.
Modifiers like `ctrl` also toggle a state in the keyboard that is not propagated through the `KeyValue` class. It's not possible to have a `ctrl+c` key on the keyboard without also having the `ctrl` modifier activated currently.

Julow:
Thanks for your PR, it's always useful to try things out.

Julow:
Here's an other way of defining string keys with a custom symbol: https://github.com/Julow/Unexpected-Keyboard/pull/774
It's defined like this: `key1=":str symbol='sym':'output'"`.

PR #739, opened 2024-08-28
Add newline and tab as whitespace

Spike-from-NH:


Spike-from-NH:
Well, that didn't work.

Julow:
Thanks for looking at this. I've implemented the fix for enter in c886a89.

Spike-from-NH:
Thanks for correcting this!

PR #743, opened 2024-09-02
Add Japanese translation and layout

cknight828:
Add Japanese translation and  qwerty layout based on OADG-109A, which is most popular layout in Japan.


Julow (approved):
Huge thanks :) And sorry for the slow reply.

What's blocking writing the kana layout ?

Issue #744, opened 2024-09-03
[feature request] modify keyboard character (font) size


58673:
How to change the keyboard icon/character/symbol size? The defaults are hard to discern -- even with normal eyesight I can hardly see some characters in the corners.

58673:
> In the Settings (swipe the gear logo), in the Style section, try the setting "Label size". 

Great.

I suggest changing the default value.

Spike-from-NH:
As shown, I agree the default is small. But changing the default requires me to change my 1.30x on all devices to retain my preference. No; please maintain backward-compatibility!

Julow:
Both approach are good :) As you have made the change, could you open a PR ?

bokidori:
I opened a PR with the mentioned first approach - just increasing the default label size which does not affect user set preferences. Upon testing it today much longer than I'd like, I don't believe it solves the issue of label size becoming too small when keyboard height is reduced.

Could @Spike-from-NH or @58673 please mention their keyboard height and label size values used? Did labels become too small because keyboard height was reduced from default 35%?

For me:
Device 1, label size 1.3, keyboard height 27%.
Device 2, label size 1.2, keyboard height 31%.

Spike-from-NH:
Oh, I see—You don't want to change what 1.00x means, only what value comes up initially. Then my objection is wrong.

I have keyboard height Landscape(40%,4-row layout), Portrait(30%,5-row layout), Labels 1.30x.

PR #747, opened 2024-09-10
Increase default label size

bokidori:
PR as per #744

This particular implementation increases default label size by 15%, while existing user set preferences remain unaffected. However, upon testing various default values (1.4, 1.3, 1.25, 1.2, 1.15, 1.1) against different keyboard heights (ranging 27%-35% for portrait mode), I came to the conclusion this might not be the right approach to resolve the issue raised there.

Why? Because the reason I believe we were increasing label size is because we were also reducing the height of the keyboard. With default keyboard height of 35%, label size of 1 is just fine. Only as we reduce the height, label becomes smaller and less intelligible.

However, modification to how the label size scales with regard to keyboard height would require a different approach, with more testing...


Julow (approved):
Sorry for the slow reply. Thanks :)

I also think that's not the right approach but this improves the current state and is a simple change. 1.15x seems good in my testing as well. Let's merge!

PR #748, opened 2024-09-10
fix bone layout

### LINKED ISSUE #741: Neo Bone missing German characters
As the title says, the "latn_bone" layout doesn't feature dedicated keys for the "ä", "ö", "ü" letters necessary for writing in German, even though the original "bone" layout includes them. They can be typed using the diacritics, but for the implementation of a layout specifically designed for the german language, this is quite cumbersome. I understand this was probably done as a space saving measure, but the standard german layout accomodates for these characters too.

quantenzitrone:
- fix jdk version in gradle issue with `shell.nix`
- fix alignment with tabs mess in bone layout file
- make like actual bone layout, instead of trying to fit it into a 10 key
  wide keyboard
  - fixes missing `ü` `ä` `ö`
  - fix missing `$`
  - moves `q` and `ß` where they belong
  - remove all the diacritic keys (they can be added through the
    settings as extra keys)
  - kept the number row extra keys integration into top row
  - kept the idea of compressing `,` and `.` into the swipe actions of the
    bottom row
  - kept the number keys as `key4` in the layer 4 positions (instead of
    moving them to a number row, which I also considered)

Screenshot_Kvaesitso_20240911-013153

 closes #741

### DIFF
--- a/check_layout.output
+++ b/check_layout.output
@@ -64,7 +64,7 @@ Layout includes some ASCII punctuation but not all, missing: (, ), <, >, [, ], {
 # latn_bepo_fr
 0 warnings
 # latn_bone
-Layout includes some ASCII punctuation but not all, missing: $
+Layout doesn't define some important keys, missing: loc esc, loc tab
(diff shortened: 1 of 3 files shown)


Julow (approved):
Many thanks for maintaining the layout :) And thanks for fixing the shell.

Sorry for the slow reply!

PR #756, opened 2024-09-17
Keyboard metadata: Point to code for built-in bottom row and number row for use as starting points

Spike-from-NH:


Spike-from-NH:
It would be more consistent if the number row, like the bottom row, were disabled via custom XML rather than by the Settings menu. Are there many people who want a keyboard without the number row, short of writing a complete custom layout?

Julow (approved):
Thanks :) I took the liberty to change the numbered list symbols to a more conventional list syntax. I also removed the last sentence about the number row option as I think was making the doc too long.

The number row is different than the bottom row as it's not enabled by default. Meaning you can unset the option in the settings add a custom one in your layout, which can't be done for the bottom row. The bottom row is an important part of a layout while the number row is not, the keyboard is not usable without.

I also expect it to be used rarely, I'd love to hear from number row users!

Issue #749, opened 2024-09-11
Change the place of چ in Persian Keyboard Layout


Danger-Mkh:
Replace the 'چ' character to place it was.

Now :
https://github.com/Julow/Unexpected-Keyboard/blob/a8e331c3222dc7f84b823e1a53e3e6fe22839210/srcs/layouts/arab_pc_ir.xml#L14

Bbefore:
https://github.com/Julow/Unexpected-Keyboard/blob/deb611a0d2c7e75a1c276238064219fc54b4417d/srcs/layouts/arab_pc_ir.xml#L39

The change is annoying and it doesn't add anything more

Spike-from-NH:
In other words, چ used to be on its own key but has been moved to a swipe of the ج key. Are these letters related? Is the first one relatively rare, that would justify this move? Is there a bigger problem than an annoying change? Is it a solution to copy-paste the old XML as a custom layout on your device?

Danger-Mkh:
It's **separate letter** like others, so it should be on it's own key.

Yes, copy-paste old XML would solve the problem but the general problem will exists.

Both 'ج' and 'چ' used as normal letter and it's not other format of a letter like 'ا' and 'آ'.
'ا' and 'آ' is the same letter but used in different places.

For example 'آ' used as first letter of word and in middle of word letter 'ا' will be used.

Danger-Mkh:
@Spike-from-NH  any suggestions?!

Julow:
This was changed in https://github.com/Julow/unexpected-Keyboard/pull/711 by @RZHSSNZDH
@RZHSSNZDH What was the idea being removing this key ?

Spike-from-NH:
@Danger-Mkh, you convinced me. @Julow asks the important question. There is an edit entitled "Change the place of چ in persian" but it doesn't say why.

RZHSSNZDH:
> This was changed in #711 by @RZHSSNZDH @RZHSSNZDH What was the idea being removing this key ?

The backspace key had a small size so I removed چ from the button row, and I could not find a place for it in other rows.

RZHSSNZDH:
> > This was changed in #711 by @RZHSSNZDH @RZHSSNZDH What was the idea being removing this key ?
> 
> The backspace key had a small size so I removed چ from the button row, and I could not find a place for it in other rows.

I do it on my own because I didn't think if there is a person who uses this keyboard in persian layout 😅
[...]

PR #757, opened 2024-09-19
Change the place of چ  and ژ in Persian Keyboard Layout

Danger-Mkh:
#749


Julow (approved):
Thanks!

@RZHSSNZDH Do you approve ?

RZHSSNZDH:
> Thanks!
> 
> @RZHSSNZDH Do you approve ?

It is fine, but I think switching the places of `.` and `ژ` will be better.
Thank you @Danger-Mkh!

Danger-Mkh:
> It is fine, but I think switching the places of `.` and `ژ` will be better.

I think the place of `.` is good because it remains consist.
@RZHSSNZDH

Danger-Mkh:
I think we're ready to merge.

What do you think @Julow ?

Julow:
Thanks !

PR #760, opened 2024-09-21
layouts/latn_bone.xml: improve

quantenzitrone:
new syntax: n/s/e/w/ne/nw/se/sw instead of key{1..8}

Some keys were missing keys according to the tests, this adds them similarly to latn_neo2.
Only `switch_backward` didn't fit (because of the 0 at the bottom of the spacebar).

I also added a `\` in front of the `@`, not sure if that is necessary, but it is part of the escape character documentation.


Julow (approved):
Thanks!

`switch_backward` and `switch_forward` are extremely important keys as they allow between layouts.
The bottom row is similar to the default one except for some keys added in this PR. To avoid maintaining this in the future and missing new features, what do you think of removing the custom bottom row in this layout ? The cost is to move `0` elsewhere.

quantenzitrone:
I've always been switching keyboard with `change_method` in the upper right of the `Fn` key.

quantenzitrone:
Oh I realize its for switching the internal layout.
Gonna fix this later.
At least `switch_forward` should be on the space bar.

quantenzitrone:
Ok now only `switch_backward` is missing.
Which may be annoying if you have more than 2 different keyboard layouts to switch between.
But switching is possible.

I could put the `0` on `se` like the other numbers, but I think it should remain on space.
I think I had problems with corner keys on the spacebar due to the length, but I could try it.

quantenzitrone:
I have an idea.
Is it be possible to combine accent_circonflex and superscript with a modmap?

quantenzitrone:
Do we have these diacritical marks?
https://en.wikipedia.org/wiki/Rough_breathing
https://en.wikipedia.org/wiki/Smooth_breathing
I have no idea why neo people put these on the bone layout. They are probably not needed, especially not on the latin layer.

Julow:
> I could put the 0 on se like the other numbers, but I think it should remain on space.

That means having to maintain the layout more often and having inconsistent placement for important keys.

>  Is it be possible to combine accent_circonflex and superscript with a modmap?

That's currently only possible by modifying the app to add this combinaison as a new diacritic.

> Do we have these diacritical marks?

We don't have Rough and Smooth breathing I believe but they could be added if necessary.

quantenzitrone:
Nah I think rough and smooth breathing are not necessary.
Not even computer neo layouts have a use for rough breathing, according to the documentation the key is completely useless.

quantenzitrone:
Would you accept if I add neo specific modifiers (diacritics)?
They could be used for neo2 and bone as well as other neo variants that may be added in the future: neoqwert{z,y}, mine, noted.
While they are not super needed, they are a neat feature of neo layouts.
https://www.neo-layout.org/Benutzerhandbuch/Tote-Tasten-und-Compose/#erzeugbare-zeichen-mit-den-toten-tasten (docs in german)

Julow:
Adding neo specific diacritics and compose sequences to the app is fine! It's possible to add a new compose key that is either a pre-applied sequence or an entirely different set of sequences.

quantenzitrone:
If you want, you can merge this like this.
I'm currently working on improving (completing) the current diacritics modifiers, so I can combine some of them in the way neo does it for the neo specific modifiers.
I opened another pr #770 because it got out of scope of this pr.

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+<keyboard name="Bone" script="latin">
+	<modmap>
+		<!-- shift on number row:
+			1234567890 -> °§ℓ»«$€„“”
+		-->
+		<shift a="1" b="°"/>
+		<shift a="2" b="§"/>
+		<shift a="3" b="ℓ"/>
+		<shift a="4" b="»"/>
+		<shift a="5" b="«"/>
```
Julow, line 12: These are already `Fn+<` and `Fn+>`

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+		<shift a="6" b="$"/>
+		<shift a="7" b="€"/>
+		<shift a="8" b="„"/>
+		<shift a="9" b="“"/>
+		<shift a="0" b="”"/>
```
Julow, line 17: These are already `Fn+"`, `Fn+(` and `Fn+)` respectively.

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+			accent_cedille/accent_ogonek
+			accent_ring
+			accent_trema
+			accent_rough_breathing (not implemented yet)
+			accent_macron/accent_macron_below (not implemented yet)
+	-->
+	<row>
+		<!--left side-->
+		<key width="0.5" key0="loc accent_circonflexe" nw="superscript"/>
+		<key key0="1" ne="¹" se="₁" nw="loc accent_caron" n="subscript"/>
```
Julow, line 50: `₁`, `₂`, `₃` and `₀` are already accessible with `subscript`, which is also present on the layout.
Same for the superscripts.

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+		<key key0="2" ne="²" se="₂" nw="↻"/>
+		<key key0="3" ne="³" se="₃" nw="loc accent_dot_above"/>
+		<key key0="4" ne="›" se="♀" nw="loc accent_hook_above" n="loc accent_horn"/>
+		<key key0="5" ne="‹" se="♂" nw="loc accent_dot_below"/>
+		<!--right side-->
+		<key key0="6" nw="¢" sw="⚥" ne="loc accent_macron"/>
```
Julow, line 56: `¢` is already `Fn+c` and `¥` is `Fn+y`

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+			accent_macron/accent_macron_below (not implemented yet)
+	-->
+	<row>
+		<!--left side-->
+		<key width="0.5" key0="loc accent_circonflexe" nw="superscript"/>
+		<key key0="1" ne="¹" se="₁" nw="loc accent_caron" n="subscript"/>
+		<key key0="2" ne="²" se="₂" nw="↻"/>
+		<key key0="3" ne="³" se="₃" nw="loc accent_dot_above"/>
+		<key key0="4" ne="›" se="♀" nw="loc accent_hook_above" n="loc accent_horn"/>
+		<key key0="5" ne="‹" se="♂" nw="loc accent_dot_below"/>
```
Julow, line 54: `‹` and `›` are already on the keyboard as `Fn+{` and `Fn+}`

srcs/layouts/latn_bone.xml
```diff
@@ -1,72 +1,131 @@
...
+			accent_rough_breathing (not implemented yet)
+			accent_macron/accent_macron_below (not implemented yet)
+	-->
+	<row>
+		<!--left side-->
+		<key width="0.5" key0="loc accent_circonflexe" nw="superscript"/>
+		<key key0="1" ne="¹" se="₁" nw="loc accent_caron" n="subscript"/>
+		<key key0="2" ne="²" se="₂" nw="↻"/>
+		<key key0="3" ne="³" se="₃" nw="loc accent_dot_above"/>
+		<key key0="4" ne="›" se="♀" nw="loc accent_hook_above" n="loc accent_horn"/>
```
Julow, line 53: Do you use symbols like `♀`, `♂`, `⚥` and `↻` so often that you need them on the main layout ? I'd suggest creating compose sequences for these.

Julow (commented):
The layout is extremely busy and the extra row makes it hard to read. Are you sure you need all these symbols on the main layer ?
Many of these are already on the `Fn` layer or in compose sequences. For even more symbols, I'd suggest you to make a second custom layout dedicated to rarely used symbols.

srcs/layouts/latn_bone.xml
quantenzitrone: yeah the fn mapping conflicts with neo layouts, 
i was considering removing it or completely mod-remapping it for some other purpose

srcs/layouts/latn_bone.xml
quantenzitrone: not really but i just wanted to make it more close to the original
i don't think it hurts to include them
Furthermore, neo layouts use ↻ in compose sequences for rotating characters. e.g. Compose ↻ r becomes ɹ

srcs/layouts/latn_bone.xml
quantenzitrone: i think they were put there because ¹²³₁₂₂₀ are used more frequently that all the other numbers

srcs/layouts/latn_bone.xml
Julow: Présumably, these symbols are on an other layer in the neo layout too ? I'm fine with remapping the entire Fn layer but the main layout seems crowded.

srcs/layouts/latn_bone.xml
Julow: Including these symbols in the main layout has two effects: The layout is crowded and hard to read. These symbols can't be typed in an other layout.
That's why I suggested adding a compose sequence for them.

The sequences for reversed characters look fun and could be included too.

Julow (approved):
Thanks! I'll hit merge now. Please don't hesitate to make more change later.

quantenzitrone:
Sorry for not doing anything on this pr for so long.
I switched to using florisboard and neo_qwerty :)

PR #770, opened 2024-09-27
compose/accent_*.json: improvements (more accents)

quantenzitrone:
i'm not finished with going through all accents, adding everything I can find


Julow (approved):
That's extremely useful, thanks a lot!
I'd merge as soon as you mark the PR as ready.

quantenzitrone:
does
```json
{
  "l": "ḷ",
  "l\u0304": "ḹ"
}
```
not work?

The generator script tells me there is a sequence collide and drops one of them.
There unfortunately is no l with only macron _(this almost sounds political lol)_.

quantenzitrone:
I finished going through all the accents and searching through unicode for symbols, I'm currently testing and tweaking.

quantenzitrone:
What do you think about `ǖ` vs `ṻ`?
`ǖ` is implemented by my font, so I think it is more common, so used this as `¯ + ¨ + u`.
Other than that, there is no reason `¯ + ¨ + u` is not `ṻ`.

quantenzitrone:
Can you do the java implementing of `accent_grave_double`?
I could probably figure it out too, but you already know your codebase.

Julow:
> {
>   "l": "ḷ",
>   "l\u0304": "ḹ"
> }

"l" with a macron doesn't exist in unicode it seems (but "ḹ" does), that's two unicode characters and cannot be part of a compose sequence.
The second sequence means "compose + , + l + \u0304 = ḹ" where "\u0304" is a combining accent that we don't want on the keyboard.

> What do you think about ǖ vs ṻ?

Both could be available as `¯ + ¨ + u` and `¨ + ¯ + u`.

> Can you do the java implementing of accent_grave_double?

Can you move that to an other PR so this one stays small and get merged faster ?
Happy to do the implementation but I lack time currently.

quantenzitrone:
> Both could be available as `¯ + ¨ + u` and `¨ + ¯ + u`.

From my observation, in the current implementation the accents have an order and no matter which one is activated first the result stays the same. For example, `` ` `` + `´` + `a` = `´` + `` ` `` + `a` = `á`, because acute takes precedence over grave and `á` + `` ` `` is not defined so stays at `á`. This is also the reason we don't have to implement shift for uppercase letters. The Java built-in shift does all the work and gets applied after everything else. Even if we implemented the cases differently, it wouldn't have any effect.

quantenzitrone:
Alright, I cleaned up my commits and removed the double grave accent.

Julow:
> From my observation, in the current implementation the accents have an order and no matter which one is activated first the result stays the same.

That's correct! Sorry for the confusion, I was thinking of the compose key.

ṻ is less common ? I think it's fine if it's not easy to type at the moment. My reasoning is that, if someone needs it, they'll also need other things and they should open an issue.
It should be typeable with the new combining diacritic keys added in 8a6dc34.

Julow:
That's an awesome change :) Thanks a lot !

Issue #541, opened 2024-02-03
Central kurdish layout design bugs


anaskaejdar:
Firstly, there is no period, nor colon.

Secondly, the key for ـہ types a keshideh (ـ) in front of the letter automatically, which seems like it'd pretty much never be the desired behaviour.

Thirdly, dotless-qaf (ٯ) is on the keyboard by default. I don't know what is the purpose of this letter, but maybe the intent was dotless-feh (ڡ) instead, as they look similar, and dotless-feh would actually make sense: it is a relatively recent optional addition to the kurdish sorani alphabet. But if dotless-feh is included, dotless-beh (ٮ) should be included also for the same reason.

Fourthly, alef with standard hamzeh above (أ) is on the keyboard, but this letter is arabic and not kurdish. If such a thing is included, it should rather be the persian long initial alef (آ), as other letters seem to have been included for reasonable compatibility with persian (ط ص ض ظ ث ذ)

Julow:
The Kurdish (کوردی) layout was made by @GoRaN909 in https://github.com/Julow/Unexpected-Keyboard/pull/353, https://github.com/Julow/Unexpected-Keyboard/pull/407

@GoRaN909 Are you using this layout often ? Is it possible that you are both using slightly different dialects ?

Adding a new layout might be a good solution, if they can be given different names.

anaskaejdar:
IF a new, separate layout is the best route, then look here, I designed a new one. Based on standard persian keyboard, and modified for compatibility with kurdish.
![Notes_240206_130829_1](https://github.com/Julow/Unexpected-Keyboard/assets/32847759/075ed94c-61d6-4cbe-8c77-eea3a316d235)

I don't know how to code this. I will try to figure it out.

Julow:
Thanks for all the information !

The kurdish layout is defined like this: [arab_pc_ckb](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/layouts/arab_pc_ckb.xml). You might be able to change it with the "edit in place" button.
I'd be glad if you could improve it :)

> I just like the kurdish keyboard way way better for typing persian than I like the persian keyboard itself.

The app can automatically add symbols to the layout depending on the language you have installed.
[...]

PR #773, opened 2024-09-28
New and Improved Kurdish layouts

anaskaejdar:
This is the two layouts I provided in issue #541
One is just the existing one, but I fixed a few problems. The other is a new one I designed.


anaskaejdar:
oh oops. Period and Colon are still missing from the qwerty one. I'mma put them on the م button

Julow (approved):
Sorry for the slow reply. This is perfect, thanks!

PR #779, opened 2024-10-03
compose/accent_grave_double.json: add double grave accent

quantenzitrone:
TODO:
- [ ] java-side implementation
- [ ] docs


Julow:
The double grave accent seems to be quite rare. So rare that's there's no glyph for it in the Roboto font.
For what purpose do you need it ?

quantenzitrone:
it is possible to type on neo layouts, so i wanted to add it for completeness

Julow:
What's missing is a glyph for the dead key. Could you make it by editing, for example, the [grave accent glyph](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/special_font/54.svg) ?

quantenzitrone:
Like this?
is `63.svg` the right filename? I looked at the other files names but couldn't find a pattern.
The accents are a bit further left than if their bottoms were fully centered, due to them being slanted they still look a bit further right.

quantenzitrone:
i had them even further left at first, but i forgot to push my changes so now they are just a tiny left.

Julow (approved):
Perfect! I've done the Java side and pushed to your branch. Let's merge.

>  The accents are a bit further left than if their bottoms were fully centered, due to them being slanted they still look a bit further right.

This seems consistent with how the accented letters look.

PR #780, opened 2024-10-04
Add Belgian layout

ptitgnu:
Adds Belgian AZERTY layout used in Belgium for Dutch, French and German languages.


srcs/layouts/latn_azerty_be.xml
```diff
@@ -0,0 +1,38 @@
...
+    <key key0="q" key2="loc tab"/>
+    <key key0="s" key2="ß"/>
+    <key key0="d" key2="é"/>
+    <key key0="f"/>
+    <key key0="g"/>
+    <key key0="h" key1="[" key2="]"/>
+    <key key0="j" key1="è" key2="accent_trema" key3="accent_circonflexe"/>
+    <key key0="k" key1="€" key2="*" key3="$"/>
+    <key key0="l" key1="loc accent_caron" key2="%" key3="ù" key4="accent_aigu"/>
+    <key key0="m" key1="loc accent_breve" key2="£" key3="µ" key4="accent_grave"/>
```
Julow, line 26: The keyboard contains àèù, is `accent_grave` useful ? For occasionally writing other accented letters, isn't the compose key enough ?
It's also possible to make this key appear only if the user has a language that needs it installed on their device.

Same for `accent_aigu` and `accent_cedille`, éç are already on the keyboard.

srcs/layouts/latn_azerty_be.xml
```diff
@@ -0,0 +1,38 @@
...
+    <key key0="e" key2="3" key3="&quot;" key4="\#"/>
+    <key key0="r" key2="4" key3="'"/>
+    <key key0="t" key2="5" key3="(" key4=")"/>
+    <key key0="y" key2="6" key3="§" key4="^"/>
+    <key key0="u" key2="7" key3="{" key4="}"/>
+    <key key0="i" key2="8" key3="!" key4="ç"/>
+    <key key0="o" key1="loc accent_ogonek" key2="9" key3="à" key4="loc accent_cedille"/>
+    <key key0="p" key2="0" key3="-" key1="_"/>
+  </row>
+  <row>
```
Julow, line 16: `~` is missing on the keyboard. It's useful to type with the compose key and might be used in passwords.
`accent_tilde` is not a substitute because it doesn't work with the compose key.

srcs/layouts/latn_azerty_be.xml
Julow, line 26: £ and `accent_grave` are very close to the edge of the screen and might be hard to type. What do you think of moving more of the side characters to the right ?

Julow (commented):
Awesome :) I've left some comments.

Julow (approved):
I've pushed the modifications I suggested. Let's merge!

PR #800, opened 2024-11-10
Update KeyValue.java - add "scrollLock" - appropriate location.

DocJr90:
Added "scrollLock" under "menu"; equates to KeyEvent.KEYCODE_SCROLL_LOCK


Julow (approved):
Thanks !
I've renamed it `scroll_lock` for consistent casing and added it to the extra keys option and the doc.

PR #809, opened 2024-12-03
Georgian layouts added

Kxeo:
Decided to make some layouts for Georgian, because it definitely doesn't seem like this app has any.

### DIFF
--- a/check_layout.output
+++ b/check_layout.output
@@ -14,8 +14,7 @@ Layout includes some ASCII punctuation but not all, missing: ", %, ', +, ,, ;, <
 # arab_pc_ckb_fa
 Layout includes some ASCII punctuation but not all, missing: ", #, $, %, &, ', ,, /, ;, ?, @, \, ^, `, |, ~
 Layout doesn't define some important keys, missing: loc esc, loc tab
-These keys are now added automatically, unexpected: f11_placeholder, f12_placeholder
-3 warnings
+2 warnings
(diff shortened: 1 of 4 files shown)


Kxeo:
Whoops. I forgot layouts.xml. Well it was not so much as forgetting than not knowing how to edit it. I'm on a tablet rn.

Julow (approved):
Thanks!

Don't worry about `layouts.xml`, it's generated automatically. I pushed to your branch.

The layouts contain every ASCII punctuation symbols. Are they useful ? If they are, only `` ` `` is missing.

I'll merge if you confirm it's as intended.

Kxeo:
> The layouts contain every ASCII punctuation symbols. Are they useful ?
Well, definitely, if you consider Armenian useful. I based both of them off the EN-US QWERTY layout.
>  If they are, only `` ` `` is missing.
Screenshot_20241204-155146
Whoops. That shoulsn't be up there. Gonna have to fix that fast.
`<key key0="ნ" key1="ჼ" key2="`" key3=":" key4=";"/>`
There. Now all is right in the world.

Julow:
> Well, definitely, if you consider Armenian useful.

Sorry, my question was, are ASCII punctuation symbols like `[]{}$~`, etc.. useful ? I guess yes as you added the missing `` ` ``.

I've pushed your change, I'll merge.

PR #814, opened 2024-12-05
Czech diacritic QWERTZ layout

tenextractor:
Layout similar to https://f-droid.org/en/packages/cz.vitSkalicky.klavesnice/ or iPhone Czech keyboard


Julow (approved):
Thanks!

PR #816, opened 2024-12-05
Welsh layout

tenextractor:


tenextractor:
image

srcs/layouts/latn_qwerty_cy.xml
```diff
@@ -0,0 +1,39 @@
...
+  </row>
+  <row>
+    <key key0="shift" key2="loc capslock"/>
+    <key key0="z"/>
+    <key key0="x" key2="loc †"/>
+    <key key0="c" key2="&lt;" key3="."/>
+    <key key0="v" key2="&gt;" key3=","/>
+    <key key0="b" key2="\?" key3="/"/>
+    <key key0="n" key2=":" key3=";"/>
+    <key key0="m" key2="&quot;" key3="'"/>
```
Julow, line 35: `'` (key3 here) is also in the center of the next key.

Julow: Shift and the delete key on the 3rd row are quite small. Couldn't the `'` be a swipe ?

tenextractor: ' is quite commonly used in Welsh, the shift and backspace are still as big as other keys.

Julow: Alright, then `'` should be removed from the corner of the `m` key. Making Shift and Backspace the same size as the other keys is fine but as they are more frequently used, having them bigger improves typing.

tenextractor: Ok, I will remove it from the `m` key.

tenextractor: Removed

srcs/layouts/latn_qwerty_cy.xml
```diff
@@ -0,0 +1,39 @@
...
+    <key key0="a" key1="loc tab" key2="`"/>
+    <key key0="s" key2="loc §" key3="loc ß"/>
+    <key key0="d"/>
+    <key key0="f"/>
+    <key key0="g" key2="-" key3="_"/>
+    <key key0="h" key2="=" key3="+"/>
+    <key key0="j" key4="}" key3="{"/>
+    <key key0="k" key3="[" key4="]"/>
+    <key key0="l" key2="|" key3="\\"/>
+    <key key0="accent_circonflexe" key2="accent_trema" key3="accent_grave" />
```
Julow, line 25: You have the dead key and the accented letters at the same time on the layout. This spot could be used for an other dead key or for '.

tenextractor: It's the same logic as the Hawaiian layout, the circumflex is the most common diacritic in Welsh and it just gives multiple options to the user to type it.

Julow (approved):
In this case, let's merge!

PR #817, opened 2024-12-05
Hawaiian layout

tenextractor:
image

### DIFF
--- a/srcs/layouts/latn_qwerty_haw.xml
+++ b/srcs/layouts/latn_qwerty_haw.xml
@@ -3,17 +3,17 @@
   <row>
     <key key0="q" key2="1" key4="loc esc"/>
     <key key0="w" key1="~" key2="2" key3="\@"/>
-    <key key0="e" key1="!" key2="3" key3="\#" key4="loc €"/>
+    <key key0="e" key1="!" key2="3" key3="\#" key4="ē"/>
     <key key0="r" key2="4" key3="$"/>
     <key key0="t" key2="5" key3="%"/>
     <key key0="y" key2="6" key3="^"/>
-    <key key0="u" key2="7" key3="&amp;"/>
-    <key key0="i" key2="8" key3="*"/>
-    <key key0="o" key2="9" key3="(" key4=")"/>
-    <key key0="p" key2="0"/>
+    <key key0="u" key2="7" key3="&amp;" key4="ū"/>
+    <key key0="i" key2="8" key3="*" key4="ī"/>
+    <key key0="o" key2="9" key3="(" key4="ō"/>
+    <key key0="p" key2="0" key3=")"/>
   </row>
   <row>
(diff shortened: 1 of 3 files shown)


srcs/layouts/latn_qwerty_haw.xml
```diff
@@ -0,0 +1,39 @@
...
+  <row>
+    <key key0="shift" key2="loc capslock"/>
+    <key key0="z"/>
+    <key key0="x" key2="loc †"/>
+    <key key0="c" key2="&lt;" key3="."/>
+    <key key0="v" key2="&gt;" key3=","/>
+    <key key0="b" key2="\?" key3="/"/>
+    <key key0="n" key2=":" key3=";"/>
+    <key key0="m" key2="&quot;" key3="'"/>
+    <key key0="accent_macron"/>
```
Julow, line 36: Shift and backspace on this row are quite small. Can't this be a swipe ?

tenextractor: It's copied from iPhone Hawaiian layout, ʻ (ʻokina) is a letter in the Hawaiian alphabet, and they're still the same size as other keys so it's okay.

srcs/layouts/latn_qwerty_haw.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard name="QWERTY (Hawaiian)" script="latin">
+  <row>
+    <key key0="q" key2="1" key4="loc esc"/>
+    <key key0="w" key1="~" key2="2" key3="\@"/>
+    <key key0="e" key1="!" key2="3" key3="\#" key4="ē"/>
```
Julow, line 6: I don't understand now we have `āēīū` on the layout but also the corresponding dead key. I'd prefer if only one solution is used.
I consider that swiping is faster than using a dead key. Dead keys are better if there are too many accented letters or if they are used less often. For Hawaiian I would suggest having `āēīū` as swipes.

tenextractor: The key4 slot is empty anyway right? Is there a downside to leaving both options available so that the user can choose?
Normally phone keyboards have extra options on long press, and even if a layout has dead keys, accented letters are still available as long press options. This keyboard has swiping instead, so I added them as swipe instead of long press.

Julow: The downsides of having too many options on layouts is that they get fuller and less readable. The keyboard also adds more keys to the layouts automatically depending on languages and options and that needs some empty space.

tenextractor: There is still a lot of empty space in other places, like on the a s d f z x keys

Julow: If you are fine with the layout, then let's merge.

PR #818, opened 2024-12-05
Irish layout

tenextractor:
image

### DIFF
--- a/srcs/layouts/latn_qwerty_ga.xml
+++ b/srcs/layouts/latn_qwerty_ga.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<keyboard name="QWERTY (Irish)" script="latin">
+  <row>
+    <key key0="q" key2="1" key4="loc esc"/>
+    <key key0="w" key1="~" key2="2" key3="\@"/>
+    <key key0="e" key1="!" key2="3" key3="\#" key4="é"/>
+    <key key0="r" key2="4" key3="$"/>
+    <key key0="t" key2="5" key3="%"/>
+    <key key0="y" key2="6" key3="^"/>
+    <key key0="u" key2="7" key3="&amp;" key4="ú"/>
+    <key key0="i" key2="8" key3="*" key4="í"/>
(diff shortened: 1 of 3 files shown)


srcs/layouts/latn_qwerty_ga.xml
```diff
@@ -0,0 +1,38 @@
...
+    <key key0="a" key1="loc tab" key2="`" key4="á"/>
+    <key key0="s" key2="loc §" key3="loc ß"/>
+    <key key0="d"/>
+    <key key0="f"/>
+    <key key0="g" key2="-" key3="_"/>
+    <key key0="h" key2="=" key3="+"/>
+    <key key0="j" key4="}" key3="{"/>
+    <key key0="k" key3="[" key4="]"/>
+    <key key0="l" key2="|" key3="\\"/>
+    <key key0="accent_aigu" key2="accent_dot_above" key3="€" key4="£"/>
```
Julow, line 25: `áéíóú` are on the layout. Is this key useful ?

tenextractor: The acute accent is very common in Irish, and it's only taking space that would be otherwise be empty.

Julow: Having both the accented letters and the accent dead key on the layout seems odd.
If that key is removed, that leaves space to make the Shift and Backspace keys bigger. These keys are more used and having them 1.5x bigger is useful. See the QWERTY (US) layout that have this.

tenextractor: Are you talking about the Welsh layout? Because removing the acute key here does not allow for bigger shift and backspace.

Julow: Right! I miss read the layout. Still, are you fine with having the dead key and the accented letters at the same time ?

tenextractor: Yes it's not a problem

Julow: Perfect, then let's merge!

PR #824, opened 2024-12-14
Lithuanian QWERTY layout

tenextractor:
Based on Samsung keyboard
image


srcs/layouts/latn_qwerty_lt.xml
```diff
@@ -0,0 +1,49 @@
...
+  <row>
+    <key shift="0.5" key0="ą"/>
+    <key key0="č"/>
+    <key key0="ę"/>
+    <key key0="ė"/>
+    <key key0="į"/>
+    <key key0="š"/>
+    <key key0="ų"/>
+    <key key0="ū"/>
+    <key key0="ž"/>
```
Julow, line 12: Not having such rows is one of the purpose of this keyboard. All these letters should be swipes on other keys.

tenextractor: It's a copy of the traditional Lithuanian layout, used on PC, Samsung, AnySoftKeyboard, and some other keyboards.

Julow: In this app, symbols in the corner of keys are more easily accessed than on other layouts. It's not useful to copy layouts directly from other apps, instead they should be adapted to make use of the app's features.

tenextractor:
The most common letters now have separate keys, the rest are swipes
image

srcs/layouts/latn_qwerty_lt.xml
```diff
@@ -0,0 +1,40 @@
...
+  <row>
+    <key key0="shift" key2="loc capslock"/>
+    <key key0="z" key4="ž"/>
+    <key key0="x" key2="loc †"/>
+    <key key0="c" key2="&lt;" key3="." key4="č"/>
+    <key key0="v" key2="&gt;" key3=","/>
+    <key key0="b" key2="\?" key3="/"/>
+    <key key0="n" key2=":" key3=";"/>
+    <key key0="m" key2="&quot;" key3="'"/>
+    <key key0="ų"/>
```
Julow, line 36: Is ų a frequently used letter ? It could be moved as a swipe to make the shift and delete keys bigger, which is a good thing.

Julow:
What would be the extra_keys for Lithuanian ? See my comments in this review: https://github.com/Julow/Unexpected-Keyboard/pull/828#pullrequestreview-2524915689

srcs/layouts/latn_qwerty_lt.xml
tenextractor: Yes, let's have separate keys for Ė and Ų.

tenextractor:
The extra_keys would be ą|č|ȩ|ė|į|š|ų|ū|ž

Julow (approved):
Awesome, let's merge!

There were already extra keys for Lithuanian: `accent_ogonek:ą:ę:į:ų@s|accent_caron:č:š:ž@f|accent_dot_above:ė@s|accent_macron:ū@o|€`
This means that `accent_ogonek` should be added on the `s` key if `ąęįų` are not already on the layout, same for the other keys. Similarly for the others. This is useful information for the future as I plan to have options to control how dead-keys and extra keys work.

PR #826, opened 2024-12-14
Estonian layout

tenextractor:
Standard Estonian layout
image


Julow (commented):
The layout is 11 columns wide. Are you sure none of the accented letters can't be in the corner of an other key to make the keys bigger ?

tenextractor:
The accented letters are not just variations, they are full letters of the alphabet, they need to be on the top level, like in the German layout. Also this is the standard Estonian layout.

Julow (approved):
If it's a standard layout that people expect, then let's merge.

But I want to emphasis that the swipes are intended to be fast and practical and that layouts should be revised to take advantage of them.

PR #827, opened 2024-12-14
Azerbaijani layout

tenextractor:
image


Julow (commented):
Similarly to https://github.com/Julow/Unexpected-Keyboard/pull/824, the top row makes the keyboard less practical. I also think that keys on the top row are harder to type than if these letters were swipes on other keys. What do you think ?

MintArchit:
Am 19. Dezember 2024 22:55:26 UTC schrieb Jules Aguillon ***@***.***>:

tenextractor:
The top row letters are too frequent, making them swipe letters is not practical.

Julow:
Why do you think swipes are not practical ? I claimed the opposite in my previous comment so can you expand ?

tenextractor:
Ə is the 3rd most common letter in Azerbaijani, ı is the 10th most common letter, and ö ü frequently occur together. Swipes are good for relatively low frequency characters like punctuation or accented letters in French, but it is not practical to use them for very common letters in the language. It's like having an English layout with the letters T and H as swipes.

Julow:
You are not convincing me that swipes are less efficient than a 5th row. The 5th row requires lifting the hand off the phone and balancing it between two hands, or it requires holding the phone in an unsafe position. 

I've looked at some Azerbaijani keyboard on the play store and found that they all use a 4 row layout with more column. For example, this one: https://play.google.com/store/apps/details?id=com.azerbaijan.keyboard.azerbaijan.language.keyboard.app
There are 12 columns layouts, which is huge but preferable to an extra row. Perhaps moving some letters as swipes could remove a column ?
The PC keyboard also uses the standard number of rows.

tenextractor:
Edited it based on PC layout:
image

Julow (approved):
Awesome! Here's the extra keys I came up with: `accent_trema:ü:ö@w|accent_cedille:ç:ş@s|ğ@g|ı@k|ə@l`

PR #207, opened 2022-09-26
Basic greek layout.

lpv11:
Made it for myself, it's pretty basic, just changed what I could see real quick but it works. Also no idea how to add 'tones' (έ,ί,ό) but really for those you use word suggestions. Also for dummies like me you might want to mention in contributing readme that you need to edit Config.xml for layouts :P
Take it or leave it I don't mind.

PR #830, opened 2024-12-16
Added Greek accent symbol button

cuhsy:
Added Greek accent symbol button; a necessary key that should always be on the keyboard, as every word over 2 syllables contains one

### DIFF
--- a/srcs/layouts/grek_qwerty.xml
+++ b/srcs/layouts/grek_qwerty.xml
@@ -8,7 +8,7 @@
     <key key0="τ" key2="5" key3="%"/>
     <key key0="υ" key2="6" key3="^"/>
     <key key0="θ" key2="7" key3="&amp;"/>
-    <key key0="ι" key1="accent_trema" key2="8" key3="*"/>
+    <key key0="ι"  key2="8" key3="*"/>
     <key key0="ο" key2="9" key3="(" key4=")"/>
     <key key0="π" key2="0"/>
   </row>
@@ -19,9 +19,10 @@
     <key key0="φ"/>
     <key key0="γ" key2="-" key3="_"/>
     <key key0="η" key2="=" key3="+"/>
-    <key key0="ξ" key2="accent_aigu" key3="{" key4="}"/>
-    <key key0="κ" key2="accent_grave" key3="[" key4="]"/>
+    <key key0="ξ" key3="{" key4="}"/>
+    <key key0="κ" key2="" key3="[" key4="]"/>
     <key key0="λ" key2="|" key3="\\"/>
+    <key key0="accent_aigu" key1="accent_trema" key2="accent_grave"/>
   </row>
   <row>
     <key width="1.5" key0="shift" key2="loc capslock"/>
--- a/srcs/layouts/grek_qwerty.xml
+++ b/srcs/layouts/grek_qwerty.xml
@@ -20,7 +20,7 @@
     <key key0="γ" key2="-" key3="_"/>
     <key key0="η" key2="=" key3="+"/>
     <key key0="ξ" key3="{" key4="}"/>
-    <key key0="κ" key2="" key3="[" key4="]"/>
+    <key key0="κ" key3="[" key4="]"/>
     <key key0="λ" key2="|" key3="\\"/>
     <key key0="accent_aigu" key1="accent_trema" key3="accent_grave"/>
   </row>
--- a/srcs/layouts/grek_qwerty.xml
+++ b/srcs/layouts/grek_qwerty.xml
@@ -8,7 +8,7 @@
     <key key0="τ" key2="5" key3="%"/>
     <key key0="υ" key2="6" key3="^"/>
     <key key0="θ" key2="7" key3="&amp;"/>
-    <key key0="ι"  key2="8" key3="*"/>
+    <key key0="ι" key2="8" key3="*"/>
     <key key0="ο" key2="9" key3="(" key4=")"/>
     <key key0="π" key2="0"/>
   </row>
--- a/srcs/layouts/grek_qwerty.xml
+++ b/srcs/layouts/grek_qwerty.xml
@@ -13,7 +13,7 @@
     <key key0="π" key2="0"/>
   </row>
   <row>
-    <key shift="0.5" key0="α" key1="loc tab" key2="`"/>
+    <key key0="α" key1="loc tab" key2="`"/>
     <key key0="σ"/>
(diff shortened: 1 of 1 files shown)


srcs/layouts/grek_qwerty.xml
```diff
@@ -19,9 +19,10 @@
     <key key0="φ"/>
     <key key0="γ" key2="-" key3="_"/>
     <key key0="η" key2="=" key3="+"/>
-    <key key0="ξ" key2="accent_aigu" key3="{" key4="}"/>
-    <key key0="κ" key2="accent_grave" key3="[" key4="]"/>
+    <key key0="ξ" key3="{" key4="}"/>
+    <key key0="κ" key2="" key3="[" key4="]"/>
     <key key0="λ" key2="|" key3="\\"/>
+    <key key0="accent_aigu" key1="accent_trema" key2="accent_grave"/>
```
Julow, line 25: As you added a key, you need to remove the `shift="0.5"` on the same raw, as that was used to center the row.

Julow, line 25: `key2` on the edge of the screen might be hard to type on some devices. Could it be on `key3` (bottom left) ?

Julow (commented):
Thanks!

@lpv11 What do you think of this change as you added the layout in https://github.com/Julow/Unexpected-Keyboard/pull/207

lpv11:
I'm afraid I have no idea which symbol you're talking about. However reinstalling your app firstly I noticed you added the toning symbols(great!), although one more than necessary. This one(ί) is probably the correct one and this one (ὶ) is probably unnecessary in modern greek. I'm not entirely sure that's the correct one but Google uses that one, I'd suspect they've done their research. Then there's still this one (ΐ) which is rather rare but used in words such us food (φαΐ), but I'd suggest you add it if you find the time. 
I'm still not sure about the symbol you guys are talking about though. Thanks for implementing and improving my change though =)

cuhsy:
@lpv11 you're right. The acute accent ( ί ) is in fact the correct one, as the grave accent ( ὶ ), along with the circumflex accent ( ῑ ) have been left behind since 1982 and are no longer being used in the official modern Greek language. I just thought they could be useful to keep, and didn't remove them, since they were in the keyboard in the first place.

Also, @Julow thank you for the feedback, I have implemented your suggested changes. Please let me know if you have any other ideas.

cuhsy:
> Then there's still this one (ΐ) which is rather rare but used in words such us food (φαΐ), but I'd suggest you add it if you find the time.

@lpv11 This is "accent_trema", which was already in the keyboard. I have already put it on key1 of the accent key, I believe. Do you have a list of all the accent names so I can verify?

lpv11:
@cushy I do not, hopefully Julow does.

lpv11:
@cuhsy I do not, hopefully Julow does.

Julow:
The list of the (weirdly named) diacritics is here: https://github.com/Julow/Unexpected-Keyboard/blob/master/doc/Possible-key-values.md#other-modifiers-and-diacritics

Julow (approved):
Looks awesome :)
shot

bill88t:
Greek here, I just updated. **NO.**
It breaks the alignment of the 2nd collumn. The tone was already available above the letter ξ.
It's easier and faster than a dedicated key.

The issue though is using english & greek with the 2nd collumn being out of place.
It makes for a horrible experience which I am already certain is not something you can pget used to.

cuhsy:
> Greek here, I just updated. **NO.** It breaks the alignment of the 2nd collumn. The tone was already available above the letter ξ. It's easier and faster than a dedicated key.
> 
> The issue though is using english & greek with the 2nd row being out of place. It makes for a horrible experience which I am already certain is not something you can get used to.

Honestly, it sounds like you're just attached to the old keyboard layout, not because it was actually easier or faster. As someone who uses this keyboard extensively, I can tell you that constantly swiping ξ to the top right for every word that needs a tone (which is most of the Greek words) is exhausting. This new layout mirrors how physical keyboards handle the tone key, so it's a lot more intuitive once you give it a chance.

As for the alignment of the second row, many other layouts, like the Russian, handle similar inconsistencies just fine. English isn't the standard for keyboard design, and Greek doesn’t have to follow it. And trust me, I get how annoying a bad layout can be. The tone above the ξ bothered me so much that, even as someone who’s not much of a coder, I spent hours figuring out how to tweak it. Now, it feels perfect, and I’m already adapting to it. I suggest giving it a bit more time. You might be surprised how quickly it grows on you.

bill88t:
The whole point of unexpected is swipes, and you just gave up on learning it's way.

When I installed unexpected, I did consider making the same PR as you, as it's quite bold having the tone there.
Except I gave the whole layout a try, seeing promise, purposefully trying to accept it as is.

Then once I memorized the swipes (which honestly took a while, yea) I realized just how fast I'd become.
While you're untrained, it's a negative, but then once you learn it, it's a positive.

I only used keyboards with dedicated tone keys before unexpected. I can't stand holding keys.
By the way, at the 1st row, it's much better than this. But, swipes are still better, and by a wide margin once you learn em.

Well, custom layouts are a thing, I'll stick to the old one.

cuhsy:
> The whole point of unexpected is swipes, and you just gave up on learning it's way.
> 
> When I installed unexpected, I did consider making the same PR as you, as it's quite bold having the tone there. Except I gave the whole layout a try, seeing promise, purposefully trying to accept it as is.
> 
> Then once I memorized the swipes (which honestly took a while, yea) I realized just how fast I'd become. While you're untrained, it's a negative, but then once you learn it, it's a positive.
> 
> I only used keyboards with dedicated tone keys before unexpected. I can't stand holding keys. By the way, at the 1st row, it's much better than this, but swipes are still better, and by a wide margin once you learn em.
> 
> Well, custom layouts are a thing, I'll stick to the old one.

I see where you're coming from, but I’d argue that swiping works best for less frequently used symbols only. The tone, however, deserves a dedicated spot because it's arguably more common than all the letters themselves.

For instance, imagine if the letter 'ς' were something you had to swipe for. You wouldn't support this either, right? That is because you have been used to it being positioned conveniently on the top left after the ';' , which, by the way, is a symbol that gets its own key despite being used far less often than the tone.

That’s why having the tone key on the right of the second row (just like on physical Greek keyboards) makes perfect sense. It’s about practicality and efficiency, not just tradition. I still encourage you to consider using the new layout, and I do believe it is all a matter of getting used to it.

PR #833, opened 2024-12-17
Added tamil layout

ms-jagadeeshan:
Added a tamil layout. The layout looks better with 0.8 scale. Is there any way to it automatically when selecting keyboard?
One more question, is there any way to add transliteration layout?
Thank you very much for creating this project, i have been using unexpected keyboard from its first release.


ms-jagadeeshan:
okay, this layout have all dependent and independent forms.

Julow (approved):
Awesome, thanks! The layout looks very nice.

Julow:
> The layout looks better with 0.8 scale. Is there any way to it automatically when selecting keyboard?

You mean the size of the symbols ? There's several ways but I think doing that globally in `KeyValue` is the one that will require the less amount of maintenance. I've implemented it and pushed to your branch, what do you think ?

> One more question, is there any way to add transliteration layout?

Like for the numpad for example ? No this is not possible currently. What would be yur use case ?

ms-jagadeeshan:
> You mean the size of the symbols ? There's several ways but I think doing that globally in `KeyValue` is the one that will require the less amount of maintenance. I've implemented it and pushed to your branch, what do you think ?

yeah i meant the symbol size, now the letters look better

> Like for the numpad for example ? No this is not possible currently. What would be yur use case ?

Transliteration layout i mean, generating tamil text from english , for example a becomes அ , aa becomes ஆ , ka becomes க etc

Julow:
Transliteration is not possible currently. Would you be interested in implementing it ?

Julow (approved):
Thanks for this nice contribution, let's merge!

I pushed to your branch to resolve the conflicts and some other changes.

ms-jagadeeshan:
Yeah, I am interested to implement, i will gothrough how other keyboards implement and you could also guide me with the approach and resources.

PR #837, opened 2024-12-18
Hindi Phonetic Keyboard Layout addition

npnpatidar:
Hindi Phonetic keyboard is added. It has all characters needed for modern Hindi as well as Sanskrit.


check_layout.output
```diff
@@ -59,6 +59,9 @@ Layout doesn't define some important keys, missing: loc esc, loc tab
 Duplicate keys: ।
 Layout includes some ASCII punctuation but not all, missing: ", $, ', ^, _, `, |
 2 warnings
+# deva_phonetic_in
+Duplicate keys: ट
```
Julow, line 63: You have `ट` at two places at the same time. Is this intended ?

srcs/layouts/deva_phonetic_in.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<keyboard name="हिन्दी फोनेटिक - Hindi Phonetic" script="devanagari">
+  <modmap>
+    <shift a="अ" b="आ" />
```
Julow, line 4: I added these to `shift.json` in bd2170e. You should be able to merge the master branch and remove most of them except for these three: ड़ क़ ज्ञ (that can't work in shift.json for now)

check_layout.output
npnpatidar: Yes. Since it is a phonetic sound of T, people may expect it to be there but there was another space available so I put it at both places  to make it more accessible.

srcs/layouts/deva_phonetic_in.xml
npnpatidar: I have removed them. I think everything looks good. Most of the those shifted to `shift.json` would be useful for other future layout  if any but some may be odd. Existing Hindi layouts don't use `<modmap>` so no need to worry I think.

npnpatidar (commented):
yep everything works as expected.

Julow (approved):
Awesome, thanks!

Issue #842, opened 2024-12-22
UI renders behind system windows on Android 15


cillyvms:
Android 15 introduces [edge-to-edge enforcement](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) for apps targeting it.
The latest Unexpected Keyboard update changed the target SDK to Android 15, but there is no code handling insets which causes the UI of activities and keyboard to render behind system bars, making it quite unusable.

| ![Screenshot_1](https://github.com/user-attachments/assets/78c9b4c3-7493-4f60-9ebb-2e48237e46f2) | ![Screenshot_2](https://github.com/user-attachments/assets/fe08860c-1940-4700-a807-8e768795a96e) |
| - | - |
| ![Screenshot_3](https://github.com/user-attachments/assets/fcd536af-d56a-47f7-ab5f-fad1ae418855) | ![Screenshot_4](https://github.com/user-attachments/assets/0fca178c-9a6e-4f66-b4d0-7cb08dccefa1) |

This can be solved by retrieving `WindowInsets` from the system and applying padding to the lists and the keyboard view accordingly.

cillyvms:
Fix in c13e7608fa473a3efb816e066a548a45d5b2726a works for gesture navigation and 3-button navigation modes, however sometimes android decides to show buttons in gesture navigation mode, which take up a similar amount of space as 3-button nav, but aren't included in the padding.
Main and settings activity UI still renders behind the action bar and status bar.

<img width=400 src="https://github.com/user-attachments/assets/9705912f-7dd5-4379-af47-b9e2d30e4b68">

Ripple of the layout change button is above the keys.

cillyvms:
The fix breaks on Android 14, on version 1.30.1 the navbar inset is applied twice.
<img width=400 src="https://github.com/user-attachments/assets/89fd8282-948b-40c5-9177-5caf5e4def36">

Julow:
Thanks for opening this issue so fast with screenshots. I released the change in a rush and got it wrong (3162a5a).
I'm still trying to fix this issue as I had problems setting up the emulators correctly. It seems that the [code that queries the insets](https://github.com/Julow/Unexpected-Keyboard/blob/3162a5abe66846a18387f351518b75a465763a62/srcs/juloo.keyboard2/Keyboard2View.java#L265-L270) can't detect the extra button on the navigation bar. Do you have an idea?

For the launcher and settings activities, I believe drawing behind the navigation is the expected behavior ?

[...]

PR #848, opened 2024-12-24
Proper support for Android 15 edge-to-edge

Julow:
Fix https://github.com/Julow/Unexpected-Keyboard/issues/842

The keyboard background now extends under the system bars and display cutout but the keys do not.
This does not yet handle the gesture navigation bar changing height when the IME dismiss and switch buttons are added.


illtown:
I'm sorry to bumb this, but have you checked emoji keyboard? It's still renders behind system bar. Also the system bar has transparent background unlike other keyboards.

PR #853, opened 2024-12-27
[Tamil Layout] Shift and fn changes

ms-jagadeeshan:
Reduced pollution in shift.json


Julow:
Why is this pollution ? Are these shift mappings only ever useful for one layout ?

ms-jagadeeshan:
It's better to put those mappings in fn(as those are not alternative form of letter, instead those are representation of numbesr 100, 1000, and symbol 'om' which look similar to the letters)

In existing layout, every letters covered without use of shift(so these special characters could be by both fn-global and shift-local), but other layout might need shift for other letter mapping(it will have only fn-global for special characters)

1 mapping which i left in shift is for Rupees symbol(it will be used in every layout)
Have shifted other 4 mapping from shift to fn(those are normal letters mapped with special symbols)

Julow (approved):
Thanks for the explanation and thanks for continuing to maintain the layout!

PR #858, opened 2024-12-29
Update kann_kannada.xml

srikanban:
Added zwnj & zwj
1. zwnj used for writing non Kannada words or nouns like Leo Tolstoy as ಲಿಯೊ ಟಾಲ್‌ಸ್ಟಾಯ್ (also can be written as ಲಿಯೊ ಟಾಲ್ಸ್ಟಾಯ್)
2. zwj used for writing ಅರ್ಕ as ಅರ‍್ಕ

Removed unecessary symbols not present in Kannada language:
1. ∪ is a set symbol, a similar symbol is used in Carnatic music, unfortunately not present in Unicode
2. ॰ used only in Devanagari
3. • bullet, not a character or alphabet in Kannada (Can't remember why I added it, maybe because I use it personally in non markdown texts)

### DIFF
--- a/res/values/layouts.xml
+++ b/res/values/layouts.xml
@@ -111,7 +111,7 @@
     <item>두벌식 (Korean)</item>
     <item>Hebrew 1</item>
     <item>Hebrew 2</item>
-    <item>ಕನ್ನಡ</item>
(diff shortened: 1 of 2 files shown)


Julow (approved):
Awesome, thanks for maintaining the layout !

PR #868, opened 2025-01-06
Revert Greek accent symbol button (#830)

bill88t:
Just the reversion of a1dd0b8f2bc1d47dd29b5fe5860907717012963c.

For more info view https://github.com/Julow/Unexpected-Keyboard/pull/830#issuecomment-2574071141

### DIFF
--- a/srcs/layouts/grek_qwerty.xml
+++ b/srcs/layouts/grek_qwerty.xml
@@ -8,21 +8,20 @@
     <key key0="τ" key2="5" key3="%"/>
     <key key0="υ" key2="6" key3="^"/>
     <key key0="θ" key2="7" key3="&amp;"/>
-    <key key0="ι" key2="8" key3="*"/>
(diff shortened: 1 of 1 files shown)


cuhsy:
Please consider my response: https://github.com/Julow/Unexpected-Keyboard/pull/830#issuecomment-2581143441

Julow:
Thanks for for discussing this change. I'm leaning in favor of the dedicated accent key so I'll close this, though that doesn't seem ideal yet. The tone dead-keys all in one place seem more intuitive for new users and makes the acute tone easier to type but it's still two strokes for typing frequent letters.
What do you think of placing all the toned letters as swipes on the layout ? For example, north for acute, south for grave, west for circumflex or ç.

cuhsy:
I personally don't like this idea for the reasons I already stated in #830 The tone key is just used too often, probably more than any other letter. It was too hard swiping it for every word, and the new tone button was the perfect solution for me.

PR #876, opened 2025-01-15
[Discussion needed] Adds functionality for Macro Complex Keys / Key Chords

lokeshlkr:
Inspired from: https://github.com/Julow/Unexpected-Keyboard/commit/b4f72ab8a98b32bc9515fdf266e90243c0cb1515

## Important
1. Appart from new code, mapping logic from `KeyModifier.turn_into_keyevent` is moved into a new function `KeyModifier.getKeyCodeFromChar` 

## Highlight
3. In new code, please check `Macro.compareTo()` method as i did not fully understand its use.
4. Also in `KeyEventHandler.send_macro()` please see if the wait is needed, if needed is it optimal. 
5. As well as the way to handle the String case, initially i planned to parse string into individual char events (to make macros simpler, like `:macro:'ctrl,ac'` instead of `ctrl,a,ctrl,c`) but then normal strings will get affected so removed it.

## EXTRA - [Not Implemented Yet]
1. This is the simple iteration, I would like to change the syntax for macro if everything else looks fine to you:
One Example i have in mind is like: `":macro:'ctrl,a,c|delete|ctrl,v,shift,v'"`
The chords are separated by pipe `|` and then within one chord the state of meta keys stay active, decoding above like:
```python
Input = 'ctrl,a,c|delete|ctrl,v,shift,v'
Chords: [
    # turn on ctrl mod; 
    # send a; 
    # send c; 
    # turn off all mods;
    ['ctrl','a','c'], 

    # send delete;
    # turn off all mods;
    ['delete'],

    # turn on ctrl mod; 
    # send v;     <-- this goes as ctrl+v 
    # turn on shift as well; 
    # send v;     <-- this goes as ctrl+shift+v 
    # turn off all mods;
    ['ctrl','v','shift','v'] 
]
```
We can discuss the utility vs complexity of this method.


srcs/juloo.keyboard2/KeyEventHandler.java
```diff
@@ -190,6 +191,37 @@ void send_key_down_up(int keyCode)
...
+        case String:
+        case Keyevent:
+          key_up(key,active_mods);
+          active_mods = _mods.as_copy();
+          break;
+        case Modifier:
+          active_mods = active_mods.with_extra_mod(key);
+        default:break;
+      }
+      wait(20);
```
Julow, line 222: What is the reason for the wait ?

Julow (commented):
I've very happy to see contributions to the code :) I unfortunately made a similar implementation in this branch: https://github.com/Julow/Unexpected-Keyboard/tree/macro
I couldn't open a PR yet due to a lack of time, sorry about that. One thing I prefer in my implementation is that it doesn't interpret keys differently in macros and uses `KeyModifier.modify` instead.

I am not happy with the syntax yet. I think it's too verbose and should be made shorter, something like `symbol:ctrl,a,ctrl,x,'string'`. Could you help with that ?

srcs/juloo.keyboard2/KeyEventHandler.java
lokeshlkr: I was thinking the system might bug out if its a complex macro and happens within few frames. So thought of adding the wait. No other specific use.

lokeshlkr:
> I unfortunately made a similar implementation

That's alright, I'll just wait for your implementation to be merged.  You can cancel this merge request.

> I am not happy with the syntax yet.

Yeah for syntax, I too think it is complex. Also I was thinking of adding the `layers` somehow for which the XML parsing needs to be modified. 
And if we go that route we can make a `tag` for each complex key. Where this tag will be put, for that i am thinking a new top level scope maybe something called `<variables>`? 

```xml
<keyboard>
    <variables>
        ...
        <macro id="myCopyMacro" symbol="copy" flags="dim,small">
            <chord mods="ctrl" keys="a,c"/>
        </macro>
        <macro id="myPasteMacro" symbol="paste" flags="dim,small">
            <chord mods="ctrl" keys="a"/>
            <chord mods="ctrl,shift" keys="v"/>
        </macro>
        ...
        <str id="myCustomString" symbol="lkr" payload="[LKR](my_github_url)" flags="dim,small"/>
        ...
        ...

    </variables>
</keyboard>
```

and using this would look like:
```xml
...
<key key0=":macro:myCopyMacro" s=":macro:myPasteMacro" n=":str:myCustomString"/>
...
```
This way the _layout_ part of the keyboard remains clutter free. And it makes easier to add more complex keys later.

Now, for writing the macro inline which is ofcourse shorter, your suggestions works best. this now becomes a general key payload, like

**if** -> payload starts with `:` means its a complex key with a variable defined for its action, like what i showed above, 
**else if** -> payload starts normally but has an un-escaped `:` in middle, that means its an in-line complex key, left part being symbol, right part being actual payload.

lokeshlkr:
# [Important]

In branch: https://github.com/Julow/Unexpected-Keyboard/tree/macro

You should add on to the existing mods when running the macro, that is more useful. Just ignoring the existing mods is not ideal. I have tried that in my approach and then changed to adding on to existing mods.

Example case:
I have a macro that does `ctrl,left` and another `ctrl,right` these move the cursor left and right one word at a time. now if we respect the existing mods, i can press `shift` and the macro keys, which will alter the selection one word at a time. Otherwise it only moves the cursor even if i manually press shift.

srcs/juloo.keyboard2/KeyEventHandler.java
Julow: I haven't found any problem with not using a wait for now.

lokeshlkr:
Regarding the mods thing:
1. the way i was handling them (not sure about your approach), existing mods don't "modify" the macro key. like modmap says to change `left` to `end` on pressing `fn`. But the macro key is `(),left`, so its immune to fn modification. I have tried it myself and is working without issues. `fn` just modifies specifically whats in modmap.
2. I tried following combo as well
> ctrl,a,ctrl,c being turned into ctrl,á,ctrl,ć 
and it is working as expected, it does not type áć. 
3. And in my own experience i prefer keeping the existing mods as is and adding on the macro mods, that increases the utility of the mod. And i feel if someone presses a mod and then runs the macro, they would want that mod not be ignored as that is the "expected" behavior.
4. And this way we don't have to write modmap for basic stuff.

All that being said, you have talked to more users than me, so you can keep the behavior which makes more sense to you. I am anyway using my custom build so its all cool.

Conclusion being, we agree that this syntax can be taken forward: `key0="123:ctrl,1,2,3"`. 
And keeping or ignoring existing mods is up to you.
For the `layers` conversation i'll keep it to myself for now and will raise an issue later.

Julow:
> Regarding the mods thing:

That's something nice in your implementation, as you are not calling `KeyModifier.modify`, modifiers like `fn` and dead keys are not used by macros.
That's interesting but that makes the macros too different from other keys, which is not ideal for maintenance.

The correct behavior for modifiers is to modify keys that are on the keyboard, not keys that are inside a macro. Modifiers can modify macros and that's the behavior that creates less surprises and integrates better with other features of the app (like custom modmaps and the way modifiers work in general).

> Conclusion being, we agree that this syntax can be taken forward: key0="123:ctrl,1,2,3".

Nothing final yet. That's just the syntax I would look into implementing first. The final syntax might be different because of other contraints: key-event keys inside macros, "dim" and "small" flags, extensibility for future needs.

Thanks you very much for your contribution in time and code :) I'm very sorry about how that went and I hope you are not discouraged from contributing again.
Would you want to implement the new syntax ?

Julow:
I've opened https://github.com/Julow/Unexpected-Keyboard/pull/878 with my version. I'll close this. Thanks again!

lokeshlkr:
> I'm very sorry about how that went

Its Ok, You were already working on that. 

> Would you want to implement the new syntax ?

I can take it if no one else is taking it up.
However, I'm working on a demo implementation of the `layers`. Will continue on that and if i am able to do it will discuss in a new PR.

PR #880, opened 2025-01-18
misc changes to deva alt layout

lrvideckis:
- remove keypad key in favor of circular swipes. This allows a larger backspace key
- move around some symbols in a way which I've found more convenient:
   - ऽ key is now a tap because i use it frequently
   - ् key moved to center to be convenient to tap with both thumbs as i use this frequently
   - । key is now a swipe (previously it was a tap); i use this symbol very seldom

### DIFF
--- a/srcs/layouts/deva_alt.xml
+++ b/srcs/layouts/deva_alt.xml
@@ -1,29 +1,29 @@
 <?xml version="1.0" encoding="utf-8"?>
 <keyboard name="देवनागरी (हिंदी)-2" script="devanagari">
   <row>
-    <key shift="0.35" width="0.9" key0="क" key1="ख" key2="घ" key3="ङ"  key4="ग"/>
-    <key              width="0.9" key0="च" key1="छ" key2="झ" key3="ञ"  key4="ज"/>
-    <key              width="0.9" key0="ट" key1="ठ" key2="ढ" key3="ण"  key4="ड" anticircle="७" indication="७"/>
-    <key              width="0.9" key0="त" key1="थ" key2="ध" key3="न"  key4="द" anticircle="८" indication="८"/>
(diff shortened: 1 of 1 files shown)


lrvideckis:
IMG_20250118_145009

srcs/layouts/deva_alt.xml
```diff
@@ -3,28 +3,27 @@
   <row>
     <key shift="0.35" width="0.9" key0="क" key1="ख" key2="घ" key3="ङ"  key4="ग"/>
     <key              width="0.9" key0="च" key1="छ" key2="झ" key3="ञ"  key4="ज"/>
-    <key              width="0.9" key0="ट" key1="ठ" key2="ढ" key3="ण"  key4="ड"/>
-    <key              width="0.9" key0="त" key1="थ" key2="ध" key3="न"  key4="द"/>
-    <key              width="0.9" key0="प" key1="फ" key2="भ" key3="म"  key4="ब"/>
+    <key              width="0.9" key0="ट" key1="ठ" key2="ढ" key3="ण"  key4="ड" anticircle="७"/>
```
Julow, line 6: The anticircle might be hard to discover. Perhaps you can add `indication="७"` on every keys with a gesture to make them visible to users ?

Julow (commented):
Thanks for maintaining the layout!

lrvideckis:
one more change: I removed the `*` swipe. I originally added this because when I text my indian friends, and I make a typo, I can quickly correct myself by fixing the typo and typing `*`

But since I added `*`, whatsapp added the feature to edit messages. So I just do this now, and I no longer use `*`

lrvideckis:
IMG_20250126_181824

Julow (approved):
Awesome, thanks!

Issue #890, opened 2025-02-03
Layouts with embedded number row & full layout (& more)


dzaima:
For a layout for [BQN](https://mlochbaum.github.io/BQN/keymap.html) an embedded number row is essentially required to squeeze in everything necessary. I've ended up [implementing a layout attribute](https://github.com/dzaima/Unexpected-Keyboard/commit/9feea92f0bcbec5a9d8cdd1f6979b5307eda9ee6) for such, so that the "Show number row" checkbox, if checked, doesn't end up with two number rows.

Would there be interest in a PR of the attribute?

Additionally, I've made a full [layout](https://github.com/dzaima/Unexpected-Keyboard/commit/71413343d2cd5a2b820ebcca1d5eaa4686993e9f) intended for landscape (embedded number row, 60%-of-104-key-like); would there be interest for a PR of that?

And also of [the layouts for BQN](https://github.com/dzaima/Unexpected-Keyboard/commit/cc4157f61290e6d95ad4cae844c3f4e23a1ca525) (a compact one for portrait that's further from the standard layout, and one based on the full layout for landscape; both have the embedded number row)? (and perhaps [APL](https://aplwiki.com/wiki/File:Keyboard_Mnemonics.png) when I get around to it) These are programming languages rather than human languages, and somewhat niche at that (though, for what it's worth, linux xkb does come with APL & BQN layouts). It comes with the minor complication of a need to `\uFE0E` some characters to have them not render as emoji.

Some unrelated things:

- I've made a ["Dark for transparent" theme](https://github.com/dzaima/Unexpected-Keyboard/commit/ca6e98e3052a9d4268b4599348d7393a38bfcedf) (black background, white keys), with which with very low key opacity (I'm using 8%) one can get a transparent dark keyboard with distinctly-lighter keys that remains as such on both light and dark backgrounds.
- Would be nice to allow to customize the slide speed; I've tuned mine down [even more](https://github.com/dzaima/Unexpected-Keyboard/commit/1133247429ea557a742acd6702bd3dfe2ed02170).
- Might make sense to add the traditional `!@#$%^&*()` to the built-in number row, providing consistent locations across layouts.
- There's no standard way to exit the compose key on accidentally tapping it; tapping again to exit would be nice.

Julow:
> I've ended up implementing a layout attribute for such, so that the "Show number row" checkbox, if checked, doesn't end up with two number rows.

That would be an awesome addition, yes please open a PR.

[...]

PR #891, opened 2025-02-03
Add embedded_number_row keyboard attribute

dzaima:
#890; BQN layout coming afterward as it depends on this.


Julow (approved):
Perfect, thanks!

Jaoheah:
The change is annoying for me. Maybe it could be changed so that the "Show Number Row" checkbox is instead like the "Show Numpad" option where you can choose if you want this new number row layout or the old one?

PR #892, opened 2025-02-03
Add swipe actions on number bar

dzaima:
From #890; a potential question would be whether to put `()` on the same char on different directions, or keep separately on 9/0; both are done by some layouts.

### DIFF
--- a/res/xml/number_row.xml
+++ b/res/xml/number_row.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <row height="0.75">
-    <key key0="1"/>
-    <key key0="2"/>
-    <key key0="3"/>
-    <key key0="4"/>
-    <key key0="5"/>
-    <key key0="6"/>
-    <key key0="7"/>
-    <key key0="8"/>
-    <key key0="9"/>
-    <key key0="0"/>
+    <key key0="1" se="!"/>
+    <key key0="2" se="@"/>
+    <key key0="3" se="#"/>
+    <key key0="4" se="$"/>
+    <key key0="5" se="%"/>
+    <key key0="6" sw="^"/>
+    <key key0="7" sw="&amp;"/>
+    <key key0="8" sw="*"/>
+    <key key0="9" sw="("/>
+    <key key0="0" sw=")"/>
 </row>


res/xml/number_row.xml
```diff
@@ -1,13 +1,13 @@
...
-    <key key0="2"/>
-    <key key0="3"/>
-    <key key0="4"/>
-    <key key0="5"/>
-    <key key0="6"/>
-    <key key0="7"/>
-    <key key0="8"/>
-    <key key0="9"/>
-    <key key0="0"/>
+    <key key0="1" sw="!"/>
```
Julow, line 3: The `sw` direction might be hard to hit on this key as it's close to the edge of the screen. The `se` direction doesn't go toward the edge of the screen and feels more natural for the thumb.
What do you think of using the `se` direction for the symbols on keys 1 to 5 ?

res/xml/number_row.xml
```diff
@@ -1,13 +1,13 @@
...
+    <key key0="2" sw="@"/>
+    <key key0="3" sw="#"/>
+    <key key0="4" sw="$"/>
+    <key key0="5" sw="%"/>
+    <key key0="6" sw="^"/>
+    <key key0="7" sw="&amp;"/>
+    <key key0="8" sw="*"/>
+    <key key0="9" sw="("/>
+    <key key0="0" sw=")"/>
```
Julow, line 12: I think the parentheses on separate keys is fine. These keys are shorter and they shouldn't be made more crammed.

Julow (commented):
This is a good way to make the keyboard more readable :) This might surprise some users at first but I hope this shouldn't be a problem as these are infrequent symbols, except for parentheses.

res/xml/number_row.xml
dzaima: I'm actually using the `s` direction in my current embedded-number-row layouts (the compact BQN layout also squeezing `` `~|\ `` on downwards swipes in its number row, in addition to the BQN upwards swipes). So I'm fine with any change here.

dzaima: Switched 1-5 to `se`; works out nicely with the BQN layout ([current look of it](https://github.com/dzaima/Unexpected-Keyboard/commit/4891cd69963ebbd1cd36762a4d4abfbef2721fac)).

Julow: Thanks for the change! I think `!()` are quite easy to type, which is what matters the most. The BQN layout is definitely an advanced layout, perhaps seeking consistency with the builtin number row is not needed ?

With the `s` direction, I would be afraid that the symbols are too close together and look messy.

Julow (approved):
That looks very nice :) I'll merge now but comments are welcome from anyone using the number row.

Julow:
Fixed in https://github.com/Julow/Unexpected-Keyboard/pull/964

PR #899, opened 2025-02-12
APL and BQN layouts

dzaima:
#890; somewhat based on existing layouts for each language but of course a ton of changes to fit into the compact size. ASCII is almost all downwards swipes, and largely consistent across the two layouts. Also am separating ASCII and non-ASCII with whitespace padding in the XML, can remove if preferred. No complaints from `./gradlew checkKeyboardLayouts`.


srcs/layouts/latn_qwerty_apl.xml
```diff
@@ -0,0 +1,50 @@
...
+  <row>
+    <key c="shift" e="loc capslock"/>
+    <key c="z" nw="⊂" ne="⊆" sw="⍧" se="⍝"/>
+    <key c="x" nw="⊃" ne="⊇"/>
+    <key c="c" nw="∩" ne="⍛"                  sw="." se=":"/>
+    <key c="v" nw="∪" ne="√"/>
+    <key c="b" nw="⊥" ne="⍎" sw="⌿" se="⍀"/>
+    <key c="n" nw="⊤" ne="⍕"                  sw="/" se="\\"/>
+    <key c="m" nw="≡" ne="≢" sw="∥" se="⍠"/>
+    <key width="1.5" c="backspace" e="delete"/>
```
Julow, line 48: `delete` at the east position is interesting but doesn't work visually. I plan to make the backspace/delete key better in the future and it would be easier if you don't make it different from the other layouts.

srcs/layouts/latn_qwerty_apl.xml
```diff
@@ -0,0 +1,50 @@
...
+    <key c="f"  s="…"/>
+    <key c="g" nw="∇" ne="⍢"                  s="_"/>
+    <key c="h" nw="∆" ne="⍙"/>
+    <key c="j" nw="∘" ne="⍤" w="⊣" e="⊢"/>
+    <key c="k"        ne="⌸"/>
+    <key c="l" nw="⎕" ne="⌷" se="⍪"/>
+    <key c=";"  w="⌺"  n="⋄"                  s="," width="0.75"/>
+  </row>
+  <row>
+    <key c="shift" e="loc capslock"/>
```
Julow, line 40: I think you meant to make this key wider:
```suggestion
    <key width="1.5" c="shift" e="loc capslock"/>
```

srcs/layouts/latn_qwerty_apl.xml
```diff
@@ -0,0 +1,50 @@
...
+    <key c="a" nw="⍺" ne="⍶"                  sw="loc tab" shift="0.25"/>
+    <key c="s" nw="⌈" ne="ᑈ" se="«"/>
+    <key c="d" nw="⌊" ne="ᐵ" se="»"/>
+    <key c="f"  s="…"/>
+    <key c="g" nw="∇" ne="⍢"                  s="_"/>
+    <key c="h" nw="∆" ne="⍙"/>
+    <key c="j" nw="∘" ne="⍤" w="⊣" e="⊢"/>
+    <key c="k"        ne="⌸"/>
+    <key c="l" nw="⎕" ne="⌷" se="⍪"/>
+    <key c=";"  w="⌺"  n="⋄"                  s="," width="0.75"/>
```
Julow, line 37: The symbol at the west position overlaps with the symbol at the center. Why not make the layout a grid ?

Julow:
There is a lot of symbols on these layouts. Are they all equally needed ? If some symbols are logical variation of an other, using the Fn and Shift layers could help make the layout more intuitive and readable. The compose key can also be extended for rarely used symbols.

Here's how they look on my device:
shot

dzaima:
For BQN, `𝕨𝕩𝕤𝕗𝕘` vs `𝕎𝕏𝕊𝔽𝔾` are in fact just a case difference, so it could make sense to move those to shift if that's possible.

Other than that, though, things get much less trivial, much from most glyphs having at least two meanings, each of which would have its own pairing (and like `⊏` & `⊑` are both very important related glyphs, whereas `⊐` is rarer, and `⊒` is perhaps the most rare glyph, but having those be present non-uniformly would be visually and learning-wise weird), and most of the semantic pairings are equal-level rather than one over another. Maybe the number bar 1-6 could be on fn (e.g. `¨` primary, `⚇` secondary on 2), and perhaps `⇐` could be removed for the already-present `fn+←`. Technically `→` isn't even used at all and we currently doesn't have plans for it; `⍳` isn't used by anything either, but it does have a plan (for complex number stuff).

A big problem with having much stuff on layers is that you'd lose the ability to visually search for the key though.

The APL layout could be trimmed a good bit more, as a consequence of there being a good number of APLs with slightly different character sets and me trying to squeeze most of the somewhat-common glyphs on the main layer, so there's a good amount of rarer stuff (`ᑈᐵ⇐∥⍧⌶⍫⍠⍞⍙⍡⍶⍹`, maybe `⍎⍕∞…⍢⌹⌺`, from a first scan; `⍱⍲` are just negations of the usual boolean `∨∧`) and I'm already relying on some fn & compose sequences for some rare glyphs.

(I didn't make the full non-"compact" layout for nothing :) )

srcs/layouts/latn_qwerty_apl.xml
dzaima: I did intentionally do offsets as I did because otherwise I typo'd everything from being used to a staggered layout :)

yeah might make sense to follow your usual rules for a built-in layout.

srcs/layouts/latn_qwerty_apl.xml
dzaima: putting `⌺` on fn+`⋄` works out well as an implicit fix for the west issue.

I assume gridifying would include keeping the shift key 1-wide as-is instead of 1.5 as in the other suggestion? I guess that could work fine (max 0.5 key-width offset from your basic layouts, and 0.25 compared to traditional staggered).

Or maybe I could try harder getting rid of the semicolon char.. (semicolon itself isn't particularly important in APL nor BQN, but housing `⋄` on it is very nice, and generally gives much-needed space for glyphs)

dzaima:
> I don't understand why ⊏ and ⊐ should be on the same level if they are not related. Are you thinking of these like brackets ? I think that they are not similar if one is much rarer than the other.

The semantic relations are `⊑`↔`⊏`, `⊏`↔`⊐`, and `⊐`↔`⊒` (but `⊑` vs `⊒` are a ship-of-theseus away); these are not at all like brackets though.  
Big annoyance is that `⊑` is more common than `⊏` but `⊐` is more common than `⊒`, and having the underbarful version be the primary for one but not the other would be very funky. I guess having fn+`⊏` → `⊐` and fn+`⊑` → `⊒` would be reasonable; though then there's the other reason of me just following the standard BQN layout for `⊏⊐⊑⊒` (and really everywhere feasible or reasonable); and I'd expect most BQN/APL users would benefit a good bit from as many shared mappings between PC and phone, I'd imagine learning two separate layouts would be quite an annoyance for those not using these langs daily (which is most people).

I'll think about some changes more.

srcs/layouts/latn_qwerty_apl.xml
Julow: > I assume gridifying would include keeping the shift key 1-wide as-is instead of 1.5 as in the other suggestion? I guess that could work fine (max 0.5 key-width offset from your basic layouts, and 0.25 compared to traditional staggered).

I didn't consider the third row when I said grid but if a very frequent symbol could be promoted to its own key, why not :)

Julow:
> though then there's the other reason of me just following the standard BQN layout for ⊏⊐⊑⊒ (and really everywhere feasible or reasonable); and I'd expect most BQN/APL users would benefit a good bit from as many shared mappings between PC and phone

That's an important concern! In my experience, crammed layouts turn people away because they look ugly but also because they are hard to get into. That's possibly different for BQN/APL users.

dzaima:
> The Shift mappins are defined here: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/shift.json\n\nFeel free to add as many as required. You'll need to run gradle test to update the generated files.

Unfortunately this appears broken for `𝕨𝕤𝕗𝕘𝕩` as those require a UTF-16 surrogate pair to represent, but all/much of the compose sequence logic appears to work with Java `char`s.

Julow:
> Unfortunately this appears broken for 𝕨𝕤𝕗𝕘𝕩 as those require a UTF-16 surrogate pair to represent, but all/much of the compose sequence logic appears to work with Java chars.

This limitation cannot be easily lifted unfortunately. You might try with considering individual surrogates to be part of the sequence.

dzaima:
Unfortunately neither of these in shift.json work:
```js
  "\uD835": {
    "\uDD68": "𝕎", // 𝕨 → 𝕎
    "\uDD69": "𝕏", // 𝕩 → 𝕏
    "\uDD57": "𝔽", // 𝕗 → 𝔽
    "\uDD58": "𝔾", // 𝕘 → 𝔾
    "\uDD64": "𝕊"  // 𝕤 → 𝕊
  },


  "\uDD68": "𝕎", // 𝕨 → 𝕎
  "\uDD69": "𝕏", // 𝕩 → 𝕏
  "\uDD57": "𝔽", // 𝕗 → 𝔽
  "\uDD58": "𝔾", // 𝕘 → 𝔾
  "\uDD64": "𝕊", // 𝕤 → 𝕊
```
So I'm keeping the uppercase doublestruck chars as main-level swipes for now. Pushed making the layout a grid, and moving some of the APL chars into fn+ sequences or requiring compose. I think they should be pretty good as-is, unless there's something else to try for the surrogate pair issue.

Julow:
That's right! My suggestion was wrong since compose sequences were not applied on to multi-16bit-chars keys. This should now work with ca25cc5

dzaima:
That works, thanks! Rebased & removed uppercase `𝕎𝕏𝕊𝔽𝔾` from the layout.

dzaima:
Oh, the merged macro thing means that e.g. `nw="↕&#xFE0E;:↕"` & co now work for the emoji-killing that I'm currently doing in Java; I suppose I should move to that now

dzaima:
Yep, don't have anything else to change.

Julow:
Perfect, let's mrege.

PR #896, opened 2025-02-08
Change the appareance of some keys

Julow:
This changes the colors of the bottom row keys and of the shift and delete keys. The intention is to make the keyboard feel less complicated. I hope this can make the keyboard more readable by making some keys easier to ignore.

Feedback welcome.

Here's how it looks:
shot

PR #904, opened 2025-02-17
Better material themes

0skar2:
I tried to fix material themes to fit gboard better, please let me know whether i messed up something :pray:

# Dark mode:
https://github.com/user-attachments/assets/37aa5744-052c-4383-bd2f-98d0bfcaa993

# Light mode:

https://github.com/user-attachments/assets/72606bb7-f635-4f63-83f0-659994d283cd

Also note that it'd be great to also color the enter, ctrl, shift and backspace with slight accent, but i think it isnt implemented yet


Julow (approved):
That's excellent! Sorry for the slow reply, I somehow missed your very welcome PR.

I'm changing the color of some keys like Shift, Backspace and the bottom row in https://github.com/Julow/Unexpected-Keyboard/pull/896 Your feedback is welcome :)

PR #907, opened 2025-02-23
Clerical edits to "Possible values.md"

Spike-from-NH:
Also, regarding escapes: ➊ The new syntax adds `:` and `,` as characters with special meanings and it would be nice to be able to escape them too. ➋ The section needs an Intro: What escaping is and where you can do so. Perhaps also demote the section so it's the tail end of the general syntax rules. Then the next major section starts the list of keywords previewed above.

### DIFF
--- a/doc/Possible-key-values.md
+++ b/doc/Possible-key-values.md
@@ -52,17 +52,22 @@ Value    | Escape code for
 `&quot;` | `"`
 
 ## Modifiers
-System modifiers are sent to the app, which is free to do whatever they want in response.
-The other modifiers only exist within the keyboard.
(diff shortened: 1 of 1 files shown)


Julow (approved):
Awesome, thanks!

Julow:
> The new syntax adds : and , as characters with special meanings and it would be nice to be able to escape them too.

`:` and `,` can be escaped using quotes: `c="macro:a,',',b"`. It's not possible to escape `:` in the symbol part for simplicity.

> The section needs an Intro: What escaping is and where you can do so.

The doc currently doesn't mention the word "escaping" and I'm happy with that.

PR #912, opened 2025-02-28
doc: Massage section on "Escape codes"

Spike-from-NH:
Other queries:

- In your example of a key that sends an email address, `@` isn't quoted at all. If items in the first table aren't mandatory, the doc should say so, or remove the items altogether.
- Possible helpful sentence: "If you revise a custom layout but Unexpected Keyboard gives a pop-up reporting a syntax error, look for characters you added that could be escaped."

### DIFF
--- a/doc/Possible-key-values.md
+++ b/doc/Possible-key-values.md
@@ -25,8 +25,8 @@ Key values can be any of the following:
   + `⏯:keyevent:85` A play/pause key (which has no effect in most apps).
   + `my@:'my.email@domain.com'` A key that sends an arbitrary string
 
-- A macro, `symbol:key_def1,key_def2,...`.
-  This results in a key that behaves as if the sequence of `key_def` had been pressed in order.
+- A macro, `legend:key_def1,key_def2,...`.
+  This results in a key with legend `legend` that behaves as if the sequence of `key_def` had been pressed in order.
 
   Examples:
   + `CA:ctrl,a,ctrl,c` A key with legend CA that sends the sequence `ctrl+a`, `ctrl+c`.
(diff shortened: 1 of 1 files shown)


doc/Possible-key-values.md
```diff
@@ -32,24 +32,29 @@ Key values can be any of the following:
...
 :---- | :------
-`\?`  | `?`
-`\#`  | `#`
-`\@`  | `@`
-`\n`  | Literal newline character. This is different from `enter` and `action` in certain apps.
-`\t`  | Literal tab character. This is different from `tab` in certain apps.
-`\\`  | `\`
+`?` | `\?`
+`#` | `\#`
+`@` | `\@`
```
Julow, line 43: These make sense when writing layout files for inclusion in the app. For example, `@` don't need to be escaped in custom layouts, extra keys settings or when it's not the first character of the key in the app's xml files.
They are separate from the others, perhaps this could be moved into a new paragraph and not a table as it's less important.

doc/Possible-key-values.md
```diff
@@ -32,24 +32,29 @@ Key values can be any of the following:
...
-`&amp;`  | `&`
-`&lt;`   | `<`
-`&gt;`   | `>`
-`&quot;` | `"`
+`&` | `&amp;`
+`<` | `&lt;`
+`>` | `&gt;`
+`"` | `&quot;`
+
+For simplicity, there is no escape for `,` or `:`.
```
Julow, line 57: This makes sense for the "key with symbol" syntax, not in general.

doc/Possible-key-values.md
```diff
@@ -32,24 +32,29 @@ Key values can be any of the following:
...
-`\\`  | `\`
+`?` | `\?`
+`#` | `\#`
+`@` | `\@`
+A literal newline character, which is different from `enter` and `action` in certain apps. | `\n`
+A literal tab character, which is different from `tab` in certain apps. | `\t`
+`\` | `\\`
 
-XML escape codes also work, including:
+Certain other characters have special effects in the XML language. To have these characters taken literally, you can escape them in the usual way for XML:
```
Julow, line 48: That's also the case for `\n`, `\t` and `\\` in a sense. What do you think of merging the two tables ?

doc/Possible-key-values.md
Spike-from-NH: I don't understand which lines your comment applies to. I've done another edit to address your other two comments.

Spike-from-NH: @Julow, I understood my confusion after the conversation in #915 and my fourth Commit should resolve this comment too. Let me know if it doesn't; or pull and post-edit.

Julow (approved):
Sorry for the slow reply. This is perfect!

Issue #887, opened 2025-01-26
UX: Shift + space bar swipe for selecting text feels unexpected and unintuitive if you don't know how it works and do it by accident


abstractionmage:
I never knew that the Unexpecred Keyboard allows you to select text by pressing shift and sliding left or right across the space bar until I watched the demo video on your GitHub readme today, and while that's a missed opportunity in its own right it also caused me some problems when I activated that feature by accident:

When I write a longer text messages like this UX issue report I usually start by writing my thoughts down top to bottom and then go over the message a few time to add more details, improve my wording, fix typos, connect sentences in meaningful ways or even rewrite entire passages. While I'm doing that, it sometimes happens that I press the Shift key to start a new sentence, only to realize that I actually want to edit a different part of my message. I swipe right on the space bar in order to move my cursor to the new location, but quickly realize that I am actually making a selection. I get totally spooked because I have no idea why that is happening and I instinctively and without thinking press on the space bar to exit the selection mode, which deletes my selection irreversibly. 

I had exactly that happen to me a fair number of times already. A rare event given that I've used Unexpected Keyboard exclusively ever since I discovered it a few years ago, but when it happens it's always pretty annoying. I would say that, depending on luck and my phone usage, this happens somewhere between once a month and once a year.

While I wanted to report this UX issue, I can't say I have any good ideas about solving it. 
- Maybe an in-app tutorial could make people aware of this feature?
- Moving the key that engages the selection mode away from the shift key might help against users entering the space bar selection mode by accident. Perhaps you could move that function to the Fn key or the modifier key between the space bar and the enter key (I still have no idea what that key does, btw.)

Spike-from-NH:
I've done that too. Moreover, when the keyboard is shifted (for any reason, including auto-cap), merely scrolling the text window makes a gigantic selection. I've written elsewhere that mistakenly swiping southeast from A often selected all—usually replaced everything with my next keystroke long before I knew what I did. This problem was solved by moving SelectAll to the west swipe.

You will quickly, if painfully, learn not to press Space to "cancel the mode." You can learn all the other quirks too. How to help new users learn quickly and not painfully is a valid question.

The key between Space and Enter is the arrow keys, if you swipe it.

scornork:
[...]

PR #913, opened 2025-03-01
Selection mode

Julow:
This adds the "selection mode", which is activated when text is selected in the text box. The selection mode is exited when the selection is cleared.

While the selection mode is activated, some keys behave differently:

- Space and Esc remove the selection without changing the text.
  The space bar is otherwise easy to type by accident during a selection and causes the selected text to be deleted.
  This fixes https://github.com/Julow/Unexpected-Keyboard/issues/887

- The left side of the space bar slider moves the left position of the selection.
  To shrink the selection from the left side, the slider must be activated by sliding to the left, extending the selection temporarilly, then by sliding to the right.

- The right side of the slider affects the right position if the selection.


abstractionmage:
Author of [#913](https://github.com/Julow/Unexpected-Keyboard/pull/913) here trying out your debug build for the first time and I have to say I really like your changes to the selection mode. It might not fully solve the UX issue of users getting into this mode by accident, but it certainly helps getting out of it without deleting your text by accident. 

I also find it really intuitive how you can now expand and shrink the left and right sides of the selection individually by sliding the space bar in certain ways.

Thanks a lot for your great work! I think I'll keep using your debug build for now until this PR gets merged.

Julow:
Thanks for trying the debug build :) I'll merge this now.

PR #920, opened 2025-03-09
Possible-key-values.md — More concise organization

Spike-from-NH:
See discussion at the end of #912.

- Avoid repetition of the list inside the list.
- New subsection **Legends** rather than adding a legend to each definition.
- Spell out more in discussion of KeyEvent.
- New subsection **Examples** at end to avoid using syntax before defining it.

### DIFF
--- a/doc/Possible-key-values.md
+++ b/doc/Possible-key-values.md
@@ -9,32 +9,21 @@ Key values appear in the following places:
 
 Key values can be any of the following:
 
-- The name of a special key. A complete list of valid special keys follows.
+- The name of a special key. A complete list of valid special keys appears later in this document.
 
-- An arbitrary sequence of characters not containing `:`.
-  This results in a key that writes the specified characters.
+- Any other sequence of characters not containing `:` results in a key that sends those characters.
 
-- The syntax `legend:key_def`.
-  `legend` is the visible legend on the keyboard. It cannot contain `:`.
-  `key_def` can be:
-  + The name of a special key, as listed below.
-  + `'string'` An arbitrary string that can contain `:`. `'` can be added to the string as `` \' ``.
-  + `keyevent:keycode` An Android keycode. They are listed as `KEYCODE_...` in [KeyEvent](https://developer.android.com/reference/android/view/KeyEvent#summary).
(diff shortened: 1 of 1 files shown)


doc/Possible-key-values.md
```diff
@@ -9,32 +9,21 @@ Key values appear in the following places:
...
-  This results in a key that writes the specified characters.
+- Any other sequence of characters not containing `:` results in a key that sends those characters.
 
-- The syntax `legend:key_def`.
-  `legend` is the visible legend on the keyboard. It cannot contain `:`.
-  `key_def` can be:
-  + The name of a special key, as listed below.
-  + `'string'` An arbitrary string that can contain `:`. `'` can be added to the string as `` \' ``.
-  + `keyevent:keycode` An Android keycode. They are listed as `KEYCODE_...` in [KeyEvent](https://developer.android.com/reference/android/view/KeyEvent#summary).
+- `keyevent:keycode` An Android keycode: one of the numbers listed with symbol `KEYCODE_...` in the Android [KeyEvent](https://developer.android.com/reference/android/view/KeyEvent#summary) document.
```
Julow, line 16: Keys with a legend are really not the same as keys without a legend. `keyevent:keycode` cannot be without a legend.

Julow (commented):
The previous version was right.

doc/Possible-key-values.md
```diff
@@ -48,7 +37,18 @@ A literal tab character, which is different from `tab` in certain apps. | `\t`
 
 The characters `?`, `#`, and `@` do not need to be escaped when writing custom layouts. Internally, they can be escaped by prepending backslash (by typing `\?`, `\#`, and `\@`).
 
-The characters `,` and `:` can be escaped in a key value, using single quotes. For example, this macro defines a key with legend `http` that sends a string containing `:`: `<key c="http:home,'https://'" />` For simplicity, `,` and `:` cannot be escaped in the key legend.
+The characters `,` and `:` must be escaped in a key value, using single quotes.
+
+**When defining a legend**, there are no escape codes.
+
+### Examples
```
Julow, line 44: I don't think adding more examples is useful. At least a few selected examples should be closer to the definition of what is a key, as it was before.

I wouldn't be against a large "examples" section, that could give customization ideas. But not part of what defines a key.

Julow (commented):
I'll close as the previous version was clearly better.

PR #949, opened 2025-03-20
Swapping the default placement of ! and | on the Colemak layout.

Jaoheah:
I swapped the placement of ! and | on the Colemak layout to make it more reasonable. It makes more sense to group | with the slashes since it's usually grouped with \ on a physical keyboard, and it makes more sense to have ! next to ? as they are both punctuation marks.


Julow (approved):
Thanks :)

PR #958, opened 2025-03-24
Possible-key-values.md—Note syntax change for slider

Spike-from-NH:
The non-backward-compatible change to the XML should be mentioned as users must revise their layouts.


Julow (approved):
Thanks!

Spike-from-NH:
Thank you for the merge! What prompted this is that Slider became "unusable" in 1.31 and I was ready to raise an issue till remembering the syntax change and editing my layouts.

Julow:
Documenting this makes total sense, thanks!

PR #962, opened 2025-03-26
Add macron dead key to QWERTY (Japanese)

HaleyHalcyon:
Even though a Latin-script keyboard is borderline useless for Japanese, which uses kanji, hiragana, and katakana, I’ve added a dead key for the macron, used in rōmaji transcription for long vowels.


Julow (approved):
Thanks!

PR #965, opened 2025-03-28
Sinhala phonetic layout

bergentroll:
# Rationale

Sinhala, or Sinhalese script is native to Sinhalese which make up the majority on Sri Lanka, about 16 million of people.

Sinhala script is an abugida descended from Brahmi script, like plenty of Indic scripts. It used nowadays to write Sinhalese language, sacred Buddhists texts on Pali and also some Sanskrit texts.

Sinhala script contains 56 letters, has combined signs and some ligatures. Also typeface is incompact, which leads to complexity of placement.

Some lack of Free and Open Source solutions to type Sinhala script exists. Vanilla AOSP keyboard does not provide it. Only few apps can be found on F-Droid by "sinhala" query, including:

  - [Sithakuru](https://codeberg.org/sinhala/sithakuru.git). 
  - [Fcitx5](https://github.com/fcitx5-android/fcitx5-android) Android port.
  - AOSP based [Indic keyboard](https://gitlab.com/indicproject/indic-keyboard).
  
The last two only provide typing of Sinhala letters with Latin onscreen keyboard as yet.

# The `sinhala_phonetic.xml`

XKB provides the "Sinhala (phonetic)" which is especially convenient for those, who are familiar with a QWERTY layout (which is true for many Sinhalese, because English is one of official languages of Sri Lanka). It actively utilizes 3rd and 4th levels, especially for vowel signs.

This PR provides the `sinhala_phonetic.xml` layout which is strongly based on XKB's one and on `latn_qwerty_us.xml`. Level 2 is mapped to "ne" direction and levels 3, 4 are on `shift` combinations, while all additional symbols of `latn_qwerty_us.xml` are presented. Some extraneous related characters had been added.

Screenshot: label size 1.04x

# Generator

To make experiments on the layout more simple the `gen_sinhala_layout.py` script had been created. So it is included in PR as proposal. It may be helpful for creating another layouts based on existing.

# Questions

If the layout may be generously merged:
- Should I delete the `gen_sinhala_layout.py`?
- May be I should squash commits?
- There is an entry in `res/xml/method.xml`. Not sure why `isAsciiCapable="true"`? But seems all layouts have `true`.

___
P. S.: Unexpected Keyboard is one love, a firm piece of FOSS, and also fine original UX. Thanks!


Julow (commented):
Thanks for this very high quality PR :)

I love how the layout is constructed. I'll definitely take inspiration from it in the future.

> Also typeface is incompact, which leads to complexity of placement.

You can add something similar to what was done with Tamil: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L758-L770
This makes the font 25% smaller for the given letters. Perhaps you need something better thought ?

> while all additional symbols of latn_qwerty_us.xml are presented.

Is it useful to have all the ASCII punctuations like `$%^&*{}[]`, etc.. on the layout ?
If these symbols are used rarely, people can easily add QWERTY as a secondary layout and switch to it when needed.

> Should I delete the gen_sinhala_layout.py?

Let's keep the script around as it's full of good ideas. I might move all the scripts to a subdirectory later.

> May be I should squash commits?

The PR will be squashed automatically on merge.

> Not sure why isAsciiCapable="true"? But seems all layouts have true.

This should possibly be set to `false` but because all other layouts have `true` here, let's keep it `true`. This is used by Android for automatically switching between layouts but that's not a useful behavior. I plan to remove `method.xml` in the future so no need to dig into this now.

bergentroll:
Thank you.

> You can add something similar to what was done with Tamil: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L758-L770
> This makes the font 25% smaller for the given letters.

Looks good as for me.

Screenshot: -25% letters size

The first approach was to add every character as for Tamil layout.
<details> 
<summary>KeyValue.java switch/case patch</summary> 

```diff
diff --git a/srcs/juloo.keyboard2/KeyValue.java b/srcs/juloo.keyboard2/KeyValue.java
index ecfdd94..5913577 100644
--- a/srcs/juloo.keyboard2/KeyValue.java
+++ b/srcs/juloo.keyboard2/KeyValue.java
@@ -768,6 +768,37 @@ public final class KeyValue implements Comparable<KeyValue>
       case "௹": case "ள": case "ஹ": case "௰": case "ௐ": case "௱":
       case "௲": case "௳":
         return makeStringKey(name, FLAG_SMALLER_FONT);
+      /* Sinhala letters to reduced size */
+      case "අ": case "ආ": case "ඇ": case "ඈ": case "ඉ":
+      case "ඊ": case "උ": case "ඌ": case "ඍ": case "ඎ":
+      case "ඏ": case "ඐ": case "එ": case "ඒ": case "ඓ":
+      case "ඔ": case "ඕ": case "ඖ": case "ක": case "ඛ":
+      case "ග": case "ඝ": case "ඞ": case "ඟ": case "ච":
+      case "ඡ": case "ජ": case "ඣ": case "ඤ": case "ඥ":
+      case "ඦ": case "ට": case "ඨ": case "ඩ": case "ඪ":
+      case "ණ": case "ඬ": case "ත": case "ථ": case "ද":
+      case "ධ": case "න": case "ඳ": case "ප": case "ඵ":
+      case "බ": case "භ": case "ම": case "ඹ": case "ය":
+      case "ර": case "ල": case "ව": case "ශ": case "ෂ":
+      case "ස": case "හ": case "ළ": case "ෆ":
+      /* Astrological numbers */
+      case "෦": case "෧": case "෨": case "෩": case "෪":
+      case "෫": case "෬": case "෭": case "෮": case "෯":
+      case "ෲ": case "ෳ":
+      /* Diacritics */
+      case "\u0d81": case "\u0d82": case "\u0d83": case "\u0dca":
+      case "\u0dcf": case "\u0dd0": case "\u0dd1": case "\u0dd2":
+      case "\u0dd3": case "\u0dd4": case "\u0dd6": case "\u0dd8":
+      case "\u0dd9": case "\u0dda": case "\u0ddb": case "\u0ddc":
+      case "\u0ddd": case "\u0dde": case "\u0ddf":
+      /* Archaic digits */
+      case "𑇡": case "𑇢": case "𑇣": case "𑇤": case "𑇥":
+      case "𑇦": case "𑇧": case "𑇨": case "𑇩": case "𑇪":
+      case "𑇫": case "𑇬": case "𑇭": case "𑇮": case "𑇯":
+      case "𑇰": case "𑇱": case "𑇲": case "𑇳": case "𑇴":
+      /* Exta */
+      case "෴": case "₨":  // Rupee is not exclusively Sinhala sign
+        return makeStringKey(name, FLAG_SMALLER_FONT);
 
       /* Internal keys */
       case "selection_mode": return makeInternalModifier(Modifier.SELECTION_MODE);
```
</details>

But it looks very big.  The current approach is to check, is a symbol from corresponding Unicode blocks. Since there are only few numeric comparisions, it looks short, but there is not sure about performance. If it is OK, it is also seems applicable to fold some other checks.

The other thing I am not sure about is a rupee sign "₨", which is included to the predicate, but actually is not exclusively Sinhalese.

> Is it useful to have all the ASCII punctuations like $%^&*{}[], etc.. on the layout ?

True, true. I thought on this proposal for some time. But think may be it is better to keep. For now it is enough place for these. And this characters may be useful on typing some numeric data.

As an idea it may be "a light" and "an extended" variants of the layout in future. There is also 20 more ancient numeric signs, which is really rare today. But they may be placed e. g. as "fn" combinations.

Julow:
Sorry for the slow reply!

Adding a case for every letter/character is the more maintainable way. The numeric way is shorter in the code but less performant and doesn't fit with the other cases. Regularity will help make change to the code and if even more of these cases are added, I'd opt for a generated table, like for the compose key, which works on characters and not on ranges.
Do you still have the bigger version ?

> The other thing I am not sure about is a rupee sign "₨", which is included to the predicate, but actually is not exclusively Sinhalese.

This might impact other users but I think it's better to have the ₨ sign be smaller for everyone.

> True, true. I thought on this proposal for some time. But think may be it is better to keep. For now it is enough place for these. And this characters may be useful on typing some numeric data.

Good for me!

> As an idea it may be "a light" and "an extended" variants of the layout in future.

If this is needed, yes! Let's not create it if no one asked for it.

> There is also 20 more ancient numeric signs, which is really rare today. But they may be placed e. g. as "fn" combinations.

Adding these would be nice! You can add Fn combinations in [this file](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/fn.json).
If a character is too rare to be useful as a Fn combination, you can add it as a compose sequence in [this file](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/compose/extra.json).
After modifying these files, you'll need to run a command, see [CONTRIBUTING.md](https://github.com/Julow/Unexpected-Keyboard/blob/master/CONTRIBUTING.md#adding-symbols-to-shift-fn-compose-and-other-modifiers).

bergentroll:
> Sorry for the slow reply!

Sure, no problem. I also was busy with routines and reply not so fast.

> Adding a case for every letter/character is the more maintainable way.

Done. `if` based commit had been removed.

> Adding these would be nice! You can add Fn combinations in [this file](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/fn.json).
> If a character is too rare to be useful as a Fn combination, you can add it as a compose sequence in [this file](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/compose/extra.json).

Such a nice abilities in Unexpected Keyboard! Therefore I decided to put the archaic digits on the modmap `fn` of the layout to not interfere with other possible layouts in future. (Hope it is OK...)

Now the ancient digits `1-10` is on the upper row keys with corresponding `1-9, 0` digits. Digits `12-19, 100` is on the next row, and the 1000 is at `NE` of `100` key. So it looks intuitive.

Screenshot: label size 1.04x

Julow:
> Such a nice abilities in Unexpected Keyboard! Therefore I decided to put the archaic digits on the modmap fn of the layout to not interfere with other possible layouts in future. (Hope it is OK...)

Totally OK !

> Now the ancient digits 1-10 is on the upper row keys with corresponding 1-9, 0 digits. Digits 20, 30 ... 100 is on the next row, and the 1000 is at NE of 100 key. So it looks intuitive.

If needed, you can add a [placeholder](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L84) key to the layout. These keys don't show up on the keyboard but can be transformed to an other symbol with the modmap.
You'd need to add it to the `Placeholder` type and to give it a name [here](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L732).

bergentroll:
> If needed, you can add a [placeholder](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyValue.java#L84) key to the layout.

Thank you. Seems no need for the moment. Big Sinhalese symbols may have thin nuances which reads better on central position.

It was a thought to hide somehow the bottom row when `fn` is pressed to not to be inconsistent with numbers. Nevertheless it is seems a redundant complication.

Julow (approved):
This is awesome, thanks!

Don't hesitate to make more changes in the future :)

PR #969, opened 2025-03-31
some housekeeping

HaleyHalcyon:
* Make French AZERTY more similar to AZERTY Amélioré
* Make ezh composable
* Make ĉĝĥĵŝ composable with the ^ after the letter

### DIFF
--- a/srcs/compose/compose/en_US_UTF_8_Compose.pre
+++ b/srcs/compose/compose/en_US_UTF_8_Compose.pre
@@ -599,8 +599,10 @@ XCOMM Latin Extended-A
 <Multi_key> <c> <apostrophe>		: "ć"	U0107 # LATIN SMALL LETTER C WITH ACUTE
 <dead_circumflex> <C>			: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
 <Multi_key> <asciicircum> <C>		: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
+<Multi_key> <C> <asciicircum>		: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
 <dead_circumflex> <c>			: "ĉ"	U0109 # LATIN SMALL LETTER C WITH CIRCUMFLEX
 <Multi_key> <asciicircum> <c>		: "ĉ"	U0109 # LATIN SMALL LETTER C WITH CIRCUMFLEX
+<Multi_key> <c> <asciicircum>		: "ĉ"	U0109 # LATIN SMALL LETTER C WITH CIRCUMFLEX
 <dead_abovedot> <C>			: "Ċ"	U010A # LATIN CAPITAL LETTER C WITH DOT ABOVE
 <Multi_key> <period> <C>		: "Ċ"	U010A # LATIN CAPITAL LETTER C WITH DOT ABOVE
 <Multi_key> <C> <period>		: "Ċ"	U010A # LATIN CAPITAL LETTER C WITH DOT ABOVE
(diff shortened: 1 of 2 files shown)


srcs/compose/compose/en_US_UTF_8_Compose.pre
```diff
@@ -599,8 +599,10 @@ XCOMM Latin Extended-A
 <Multi_key> <c> <apostrophe>		: "ć"	U0107 # LATIN SMALL LETTER C WITH ACUTE
 <dead_circumflex> <C>			: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
 <Multi_key> <asciicircum> <C>		: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
+<Multi_key> <C> <asciicircum>		: "Ĉ"	U0108 # LATIN CAPITAL LETTER C WITH CIRCUMFLEX
```
Julow, line 602: This file is copied from an other project (from X11) and is not meant to be modified.
You should add the new compose sequences to this file instead: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/compose/compose/extra.json

srcs/layouts/latn_azerty_fr.xml
```diff
@@ -1,39 +1,39 @@
...
-    <key key0="k" key1="è" key2="€" key3="$"/>
-    <key key0="l" key2="%"/>
-    <key key0="m" key3="*"/>
+    <key key0="q" key2="loc tab" key4="loc esc"/>
+    <key key0="s" key1="\@" key2="\#" key3="loc ß" key4="§"/>
+    <key key0="d" key1="`" key2="€" key3="^" key4="$"/>
+    <key key0="f" key3="~" key4="¤"/>
+    <key key0="g" key3="{" key4="}"/>
+    <key key0="h"/>
+    <key key0="j" key1="ù" key3="“" key4="”"/>
```
Julow, line 24: Symbols like ÷×… and quotes are available with the Fn key. I think that should be sufficient ?

srcs/layouts/latn_azerty_fr.xml
```diff
@@ -1,39 +1,39 @@
...
-    <key key0="x" key1="loc †"/>
+    <key key0="w" key3="&lt;" key4=">"/>
+    <key key0="x" key1="loc †" key2="÷" key4="×"/>
     <key key0="c" key1="loc accent_cedille" key2="ç" key3="," key4="\?"/>
-    <key key0="v" key3=";" key4="."/>
-    <key key0="b" key3=":" key4="/"/>
-    <key key0="n" key1="loc accent_tilde" key2="§" key4="!"/>
+    <key key0="v" key2="…" key3=";" key4="."/>
+    <key key0="b" key1="°" key2="\\" key3=":" key4="/"/>
+    <key key0="n" key1="loc accent_tilde" key2="EFIS:nnbsp" key3="=" key4="!"/>
```
Julow, line 36: `nnbsp` can be added in the settings. Isn't that sufficient as this symbol is used quite rarely ?

srcs/layouts/latn_azerty_fr.xml
```diff
@@ -1,39 +1,39 @@
...
-    <key key0="r" key2="4" key3="'"/>
-    <key key0="t" key2="5" key3="(" key4=")"/>
-    <key key0="y" key2="6" key3="-" key4="|"/>
-    <key key0="u" key2="7" key3="à" key4="`"/>
-    <key key0="i" key2="8" key3="_" key4="\\"/>
-    <key key0="o" key2="9" key3="\@"/>
-    <key key0="p" key2="0"/>
+    <key key0="a" key2="1" key4="à"/>
+    <key key0="z" key1="accent_aigu" key2="2" key3="é" key4="½"/>
+    <key key0="e" key1="accent_grave" key2="3" key3="è" key4="æ"/>
```
Julow, line 8: Do you need `accent_grave` if you also have èàù ?

srcs/layouts/latn_azerty_fr.xml
```diff
@@ -1,39 +1,39 @@
...
+    <key key0="r" key1="accent_circonflexe" key2="4" key3="ê" key4="&amp;"/>
+    <key key0="t" key1="accent_trema" key2="5" key3="(" key4=")"/>
+    <key key0="y" key1="accent_macron" key2="6" key3="[" key4="]"/>
```
Julow, line 11: If you ever need `accent_macron` I think you could use the compose key ?

Julow (commented):
I think the new layout is too different from the previous one and users might not appreciate the change.
I also find the layout a bit too dense. Perhaps some keys are not needed ?

If other people like this new layout, perhaps you can create a new AZERTY variant instead of changing the existing one ?

srcs/layouts/latn_azerty_fr.xml
HaleyHalcyon: NNBSP is actually the most correct space to put before big punctuation in French, but technical reasons (it not being available in old AZERTY) caused regular NBSP or ASCII space to be used instead.

HaleyHalcyon:
I’ll remove the extra diacritic deadkeys. I added them because Ameliorized AZERTY is also meant to enable multilingual typing, but as you point out, the Compose key can deal with that.

Julow:
I'll close. Thanks a lot for your work and I hope to see more of it.

Issue #997, opened 2025-05-22
[Feature Request] Simple mod to allow full scaling of the key pad past "safe" limits


Walters-Marissa:
This is quite possibly the best keyboard I've ever seen. theres only one thing keeping it from true LG wing greatness. Please allow the key scaling to be pushed to 100% of the screen. You can put it behind a warning or check box perhaps, but please remove the artificial limit as it keeps it from being comfortable to use on the second screen. It already swaps pretty well with my modded lg wing's keyboard switcher, and for the first time, I can do all kinds of excel operations. several options to extend the keys to 100% also work in other example keyboards such as heliboard, but they don't have the features and usability of this one. The 50% of portrait is not enough.

Julow:
I'm not sure I understand correctly. Do you mean an option to remove all the margin on the left and right sides of the keyboard ?

Spike-from-NH:
@Julow, he's asking that, in Settings > Keyboard height, the slider not have a maximum of 50%.

Walters-Marissa:
> [@Julow](https://github.com/Julow), he's asking that, in Settings > Keyboard height, the slider not have a maximum of 50%.

This. It should allow the full height.

Julow:
I've done that in https://github.com/Julow/Unexpected-Keyboard/pull/1005
Can you confirm that it works and looks good ? You can find a debug build at the end of this page: https://github.com/Julow/Unexpected-Keyboard/actions/runs/15334657276?pr=1005

Spike-from-NH:
> Only an option to add another row of functions or macros

Have you tried [custom layouts](https://github.com/Julow/Unexpected-Keyboard/blob/master/doc/Custom-layouts.md)? You can have all the rows you want!

Julow:
Thanks for trying ! https://github.com/Julow/Unexpected-Keyboard/pull/1005 is merged.

> There is also the autocomplete or function row of android keyboards, but i think that was not included intentionally and to be honest i am not one of the ones who uses it often other than as spell check.

This is a very often requested feature and I hope to make it this year.

> More importantly lg wings lack of any support except for their proprietary keyboard means that swapping and clicking spaces reframes the keyboard into the wrong screen again.

Also, when you unfold the screen, the keyboard becomes unusable ?
The keyboard could perhaps support the LG wing directly but it's impossible to do without having the phone. Do you think you could do the work ? There's also the slight possibility that it's impossible.
It would be nice to continue this discussion in an other issue.

[...]

PR #1005, opened 2025-05-29
Settings: Remove the upper limit in keyboard height

Julow:
Fix https://github.com/Julow/Unexpected-Keyboard/issues/997

This can be used on phones that have several screens as reported in https://github.com/Julow/Unexpected-Keyboard/issues/997
Some work is done to avoid rendering glitches when the option is set to a very high value.


Julow:
Validated in https://github.com/Julow/Unexpected-Keyboard/issues/997

Issue #808, opened 2024-11-27
Custom layout for number / FN / PIN layouts


matejdro:
From what I can see, I can only define custom layout for the main keyboard?

Any chance we could also customise other layouts? Thanks.

Spike-from-NH:
See also #806 on customizing the emoji layout. At present, you cannot. It is a little more complicated than it looks; for the main keyboard, there are a bunch of built-in layouts or maybe several Custom Layouts, **plus** remembering which of them was last selected in Portrait and which in Landscape. Plus the question whether a change of main keyboard layout wants to also change the other keyboards.

PR #1007, opened 2025-06-04
Add option to also disable number entry layout

matejdro:
PR adds another setting next to the PIN entry setting that also disables automatic switch to the number entry layout whenever user focuses a number field.

For me, this is a workaround to the #808. For my use case, numpad layout is missing some of the keys that I use regularly (such as copy paste). And since it cannot be customized and because I have a number row on the main layout anyway, just using that layout for number entry is not that much worse than using the dedicated numpad layout.

### DIFF
--- a/srcs/juloo.keyboard2/Config.java
+++ b/srcs/juloo.keyboard2/Config.java
@@ -12,7 +12,6 @@ import java.util.Map;
 import juloo.keyboard2.prefs.CustomExtraKeysPreference;
 import juloo.keyboard2.prefs.ExtraKeysPreference;
 import juloo.keyboard2.prefs.LayoutsPreference;
-import juloo.keyboard2.NumberLayout;
 
 public final class Config
 {
--- a/srcs/juloo.keyboard2/Keyboard2.java
+++ b/srcs/juloo.keyboard2/Keyboard2.java
@@ -26,7 +26,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import juloo.keyboard2.prefs.LayoutsPreference;
-import juloo.keyboard2.NumberLayout;
 
 public class Keyboard2 extends InputMethodService
   implements SharedPreferences.OnSharedPreferenceChangeListener


res/xml/settings.xml
```diff
@@ -26,6 +26,7 @@
     <CheckBoxPreference android:key="vibrate_custom" android:title="@string/pref_vibrate_custom" android:defaultValue="false"/>
     <juloo.keyboard2.prefs.IntSlideBarPreference android:dependency="vibrate_custom" android:key="vibrate_duration" android:title="@string/pref_vibrate_duration_title" android:summary="%sms" android:defaultValue="20" min="0" max="100"/>
     <CheckBoxPreference android:key="pin_entry_enabled" android:title="@string/pref_pin_entry_enabled_title" android:summary="@string/pref_pin_entry_enabled_summary" android:defaultValue="true"/>
+    <CheckBoxPreference android:key="number_entry_enabled" android:title="@string/pref_number_entry_enabled_title" android:summary="@string/pref_number_entry_enabled_summary" android:defaultValue="true"/>
```
Julow, line 29: Having two checkboxes for encoding three possible states feels weird. Can you change them into a `ListPreference` ?
Also, the titles of the options are confusing. The list pref could be something like "Layout when typing numbers, dates and phone numbers", "PIN entry", "number pane", "Use the main layout".
This also makes room for a "Custom" case.

Julow (commented):
Awesome !

res/xml/settings.xml
matejdro: Good sugestion. Updated the code.

srcs/juloo.keyboard2/Config.java
```diff
@@ -12,6 +12,7 @@
 import juloo.keyboard2.prefs.CustomExtraKeysPreference;
 import juloo.keyboard2.prefs.ExtraKeysPreference;
 import juloo.keyboard2.prefs.LayoutsPreference;
+import juloo.keyboard2.NumberLayout;
```
Julow, line 15: The import is not needed, it's in the same namespace
.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -26,6 +26,7 @@
 import java.util.List;
 import java.util.Set;
 import juloo.keyboard2.prefs.LayoutsPreference;
+import juloo.keyboard2.NumberLayout;
```
Julow, line 29: Import also not needed.

srcs/juloo.keyboard2/Config.java
```diff
@@ -169,7 +170,7 @@ public void refresh(Resources res, Boolean foldableUnfolded)
     switch_input_immediate = _prefs.getBoolean("switch_input_immediate", false);
     extra_keys_param = ExtraKeysPreference.get_extra_keys(_prefs);
     extra_keys_custom = CustomExtraKeysPreference.get(_prefs);
-    pin_entry_enabled = _prefs.getBoolean("pin_entry_enabled", true);
+    selected_number_layout = NumberLayout.valueOf(_prefs.getString("number_entry_layout",  "pin").toUpperCase());
```
Julow, line 173: I didn't know the `valueOf` method !

Julow (approved):
That's totally awesome! And thanks for thinking about the migrate function.

Julow:
This is awesome, thanks a lot!

matejdro:
Oops, sorry, forgot about this PR. Thanks for looking into it!

PR #1009, opened 2025-06-11

JeelsBoobz:


srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -124,6 +125,46 @@ public void onCreate()
     _foldStateTracker.setChangedCallback(() -> { refresh_config(); });
   }
 
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    int inputClass = _lastInputType & InputType.TYPE_MASK_CLASS;
+    InputConnection conn = getCurrentInputConnection();
+    if (conn == null || !isInputViewShown() || inputClass != InputType.TYPE_CLASS_TEXT) {
```
Julow, line 132: Do you think the `_lastInputType` thing could be avoided if this code was moved to `Keyboard2View` ?

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -124,6 +125,46 @@ public void onCreate()
...
+        }
+      }
+      return true;
+    }
+    return super.onKeyDown(keyCode, event);
+  }
+
+  @Override
+  public boolean onKeyUp(int keyCode, KeyEvent event) {
+    if (keyCode == KeyEvent.KEYCODE_VOLUME_UP || keyCode == KeyEvent.KEYCODE_VOLUME_DOWN)
```
Julow, line 163: This doesn't do the same `isInputViewShown` and `_lastInputType` checks as `onKeyDown`.

Julow (commented):
That's a really good idea! But it should be guarded behind an option and be disabled by default. People will definitely not expect this.

JeelsBoobz:
> That's a really good idea! But it should be guarded behind an option and be disabled by default. People will definitely not expect this.

I think you are right, you should use options to enable/disable. You can make it and maybe better code.

Julow:
I don't have enough time to do work on this feature. Anyone else would want to do it ?

Issue #954, opened 2025-03-22
selection mode improvement


alotbsol555:
hi, thanks for adding selection mode, I think it is a brillant idea and very intuitive to use!

however, there is one issue: if you move one end of the selection beyond the other end of the selection, the behaviour becomes erratic. the handles to size the selection via android disappear and sometimes single letters outside the initial selection are selected and selection mode stays active. sometimes selection mode is exited. maybe selection mode should be exited, if the selection shrinks to zero length. or the selection should just increase consistently beyond the other end of the selection.

hopefully this video helps:

https://github.com/user-attachments/assets/7e213690-0032-4e10-87bf-fa01ceb4f41d

Julow:
I've made the behavior more consistent in https://github.com/Julow/Unexpected-Keyboard/pull/1012 but I'm not sure it's intuitive. Can you tell me what you think ?

I think a more intuitive behavior would be to inverse the left and right ends of the selection. For example, if the right end hits the left end, the motion would then act on the left end, the right end stays where the collision happened. This is harder to implement as we must make sure not to keep the inversed controls for the next stroke.

Julow:
After some more testing, I find the behavior I proposed above to be unintuitive and decided to implement the better behavior in https://github.com/Julow/Unexpected-Keyboard/pull/1013

Could you try it ? The debug build is at the end of this page: https://github.com/Julow/Unexpected-Keyboard/actions/runs/15667950603?pr=1013

alotbsol555:
hi, thanks for looking into this :) the debug build seems to be working as described and I think it is the most intuitive way for the cursor movement.

one minor point: I would suggest to automatically exit selection mode if keys like delete/forward_delete/cut/ctrl+x etc are used, i.e. when the selection is gone or reduced to zero length. currently the space bar still shows the "esc"-label, but the left/right-cursor will of course move normally without a selection.

Julow:
@alotbsol555 Thanks for checking, I'll merge.

I've found a fix for the selection mode exit bug: https://github.com/Julow/Unexpected-Keyboard/pull/1020
Can you check it out too ?

PR #1012, opened 2025-06-15
Avoid the selection being empty in selection mode

Julow:
Fix https://github.com/Julow/Unexpected-Keyboard/issues/954

Emptying the selection with the spacebar slider puts the selection mode in an unexpected state. This checks that this doesn't happen.

For example, moving the right end of the selection (by going right then left in a single stroke) after it reached the left end will push it with every steps to keep the selection of size 1.


Julow:
Closed in favor of https://github.com/Julow/Unexpected-Keyboard/pull/1013

PR #1014, opened 2025-06-15
Fix: delay too fast

JeelsBoobz:


Julow (commented):
Can you describe the bug this is fixing ? Did you test this ?

JeelsBoobz:
> Can you describe the bug this is fixing ? Did you test this ?

Autocapital trigger detection too fast from event detection

Julow:
What is the bug that you observes from the user point of view ?

PR #1016, opened 2025-06-17
Added Macedonian language layout

dacc9:
Added a Macedonian language layout. Compiled and tested. It's working well.


srcs/layouts/cyrl_lynyertdz_mk.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<keyboard name="ЉЊЕРТЅ (Македонски)" script="cyrillic">
+  <row>
+    <key width="1.09" key0="љ" key2="1" key4="esc"/>
```
Julow, line 4: I've added the `scale` attribute to avoid having to write all these widths in https://github.com/Julow/Unexpected-Keyboard/pull/1018
If you rebase, you'll be able to do `<row scale="12">`.

Julow (commented):
`python3 gen_layouts.py` reports:
```
cyrl_lynyertdz_mk: Duplicate keys: ., `
cyrl_lynyertdz_mk: Layout includes some ASCII punctuation but not all, missing: ", ,
```
The second warnings is a bit hard to read, it says that `"` and `,` are missing while the other ASCII punctuations are present.
If that's on purpose, you can keep it like that, thought I think the duplicate keys are not intentional.

srcs/layouts/cyrl_lynyertdz_mk.xml
dacc9: I'm unsure about how the scale attribute works, is the value the number of keys in the row? If so, when set it as 11, for example for the first row, I still get the trailing blank space to the right of the keys.

dacc9: Duplicate characters are now fixed.
Also the new kind of scaling is implemented.
The layout  missing **"** is intentional, because Macedonian uses **„“** for quotes.

Julow (approved):
That's perfect, thanks!

Issue #1011, opened 2025-06-15
Only copy when text is selected


artiekra:
Copy and paste keys are very close together:

![Image](https://github.com/user-attachments/assets/91acb464-49cf-4983-8df4-e323643e916d)

I am not against this layout and i want to use it, but the problem is that its easy to press "copy" on accident instead of "paste".. Even if i dont have any text selected, the copy operation will still execute, basically clearing my clipboard

I suggest that "copy" only works when you have at least one letter selected

This is the problem that annoys me:

https://github.com/user-attachments/assets/cdaeb13f-eb57-410b-9307-28cb7ee74e7d

Spike-from-NH:
See #911. Whether text is selected, and what the Copy command does if nothing is selected, are up to the app. I use a text editor that guards against this mistake, and a custom layout that moves one command to make it harder to type. Unexpected's Clipboard Manager will have your "lost" text.

artiekra:
Sorry, missed that, thanks for the reply!

Julow:
This is not a duplicate. I don't know how to improve the behavior for undo and redo but for copy, I think your suggestion would work.
If someone wants to have a look, this can probably be done in [`handle_editing_key`](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/KeyEventHandler.java#L232)

Julow:
Fixed in https://github.com/Julow/Unexpected-Keyboard/pull/1019 by @alotbsol555

PR #1019, opened 2025-06-18
cut and copy with built in keys only if text is selected

alotbsol555:
cut and copy with built in keys only if text is selected to keep existing clipboard content. does not work for key combinations like ctrl+x or ctrl+c.

possible fix for #1011.


Julow (approved):
This works perfectly :) I just have one remark on the code.

Julow (approved):
That's perfect, thanks!

artiekra:
Appreciate you working on this, thanks! :3

PR #1020, opened 2025-06-18
Fix selection mode not exited when selection cleared

Julow:
The bug was mentioned in https://github.com/Julow/Unexpected-Keyboard/issues/954#issuecomment-2975478282

The selection mode was not exited when the selection was cleared with, for example, ctrl+x.


alotbsol555:
great! this works just fine :)

I used the debug build from here:
https://github.com/Julow/Unexpected-Keyboard/actions/runs/15745487863

Julow:
Thanks for testing :)

PR #1026, opened 2025-07-02
Add a missing comma in /res/values/strings.xml

Jaoheah:
I believe that "Layout when typing numbers, dates and phone numbers" should be "Layout when typing numbers, dates, and phone numbers". 

I was going to include changing "Prev" to "Previous" but I am unsure the context of its use, so it may or may not be a good idea to do that.


Julow (approved):
Thanks!

"Prev" is used for the "action" key that goes on the enter key so it should remain short.

Issue #1027, opened 2025-07-02
[Feature Request] Manual UI Language Selector


Jaoheah:
I would like to see a way to manually select the language used in the app instead of it always being based on the language set in system settings. Maybe it can be an overide in the settings, like how Customize Borders work, you would have to tick a checkbox that says "Manual Language Selection", then a button for a language menu would ungray, the menu could probably function like the theme selector.

The reason I am asking for this is because I have translated everything to Shavian and realised that you would need a way to manually select a language to use it as Android doesn't have a Shavian Locale.  I will also open a draft PR with the translations I have done.

PR #1028, opened 2025-07-02
Shavian Translation

Jaoheah:
Here is everything translated to Shavian, though from what I can tell it wont be able to be used until a manual langauge selector is added as I mentioned in https://github.com/Julow/Unexpected-Keyboard/issues/1027


Jaoheah:
@HaleyHalcyon I translated Unexpected Keyboard into Shavian. I am not sure if you would want to take a look at it or not to see if I made any mistakes, as you have contributed to Unexpected Keybaord before and understand Shavian.

res/values-sh/strings.xml
HaleyHalcyon, line 1: Line 53: either remove the “·” (i.e. treat “PIN” as an acronym like “LASER” and “ICE”), or respell it with an Acroring (⸰𐑐𐑲𐑯 — «𐑐»𐑻𐑕𐑩𐑯𐑩𐑤 «𐑲»𐑛𐑧𐑯𐑑𐑦𐑓𐑦𐑒𐑱𐑖𐑩𐑯 «𐑯»𐑳𐑥𐑚𐑼)

HaleyHalcyon, line 1: Line 72: “Monet” should be “·𐑥𐑴𐑯𐑱” (namer dot, [mo-NAY])

HaleyHalcyon, line 1: Line 75: NOT an error, it really is called “Rosé Pine”

PR #1029, opened 2025-07-02
Use screen width to switch layouts

matejdro:
As discussed on the #982, this changes layout switching logic to detect screen width rather than portrait/landscape/unfolded state.

### DIFF
--- a/srcs/juloo.keyboard2/Config.java
+++ b/srcs/juloo.keyboard2/Config.java
@@ -5,7 +5,6 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.util.DisplayMetrics;
 import android.util.TypedValue;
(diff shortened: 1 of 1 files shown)


Julow (commented):
Awesome! This works perfectly on my device but I'll do a bit more testing before merging.

The hardware bezel should already be taken into account when computing the padding around the keyboard. Do you have problems with the keyboard getting in contact with the bezels ?

matejdro:
No, I personally don't have any issues on my device, but I imagine there might be on another devices? Just want to be sure to not break someone's workflow.

Julow (approved):
Now that the release is done, I'll merge this and I'll test this over the next release cycle.

I think the usefulness of the margin and size settings can only be determined by using foldables. Let's not answer this question now and let's improve the support over time.

Issue #1032, opened 2025-07-02
Different behavior between previous Slider with Left/Right keys and keys with cursor_left and cursor_right in current debug version


Roosader:
Hello! I have one of the current debug builds installed, and it seems that the left/right cursor keys have regressed functionality in some specific contexts versus previous versions that used the slider syntax with left/right keys added. Here's an example video in the Godot editor with a custom layout I made as an example:

https://github.com/user-attachments/assets/d1304cb6-abeb-4f11-96ed-55a41b193a6b

Oddly, the cursor_up/cursor_down keys work just fine, but cursor_left/cursor_right do not work. They do work in normal text boxes and such though, so I'm not sure what the difference is (though it could be an issue with how the Godot editor handles input). I really like the new vertical slider functionality, so I'd prefer if perhaps a workaround of some sort could be implemented, but I can switch back to an older version of the keyboard if necessary for horizontal control (albeit sacrificng the new vertical control)

Thank you!

Julow:
Did this work with a previous version of the keyboard ? I tried with 1.30.3 and 1.31.1 and I get the broken behavior.

Roosader:
> Did this work with a previous version of the keyboard ? I tried with 1.30.3 and 1.31.1 and I get the broken behavior.

Hey there! Thanks for giving this a look! Yes actually, That's what I was trying to show in that video. 

The first demonstration (the transparent keyboard) was using 1.30.3 with w="left" and e="right" set on a key with the slider="true" property. The second demonstration was using a 1.31.1 debug build on a key with w="cursor_left" and e="cursor_right". It seems that the previous method of using "left" and "right" and slider="true" no longer works, which I'm assuming was intentional?

Julow:
Ah yes, I misread your first message. I found a not very subtle workaround: https://github.com/Julow/Unexpected-Keyboard/pull/1033
Does this work for you ? Debug build: https://github.com/Julow/Unexpected-Keyboard/actions/runs/16038466359?pr=1033

Roosader:
> Ah yes, I misread your first message. I found a not very subtle workaround: [#1033](https://github.com/Julow/Unexpected-Keyboard/pull/1033) Does this work for you ? Debug build: https://github.com/Julow/Unexpected-Keyboard/actions/runs/16038466359?pr=1033

[...]

PR #1033, opened 2025-07-03
Workaround Godot editor not implementing setSelection()

Julow:
Issue: https://github.com/Julow/Unexpected-Keyboard/issues/1032


Julow:
Thanks @Roosader for testing!

This is hopefully a temporary work around until the keyboard can detect bad editors.

PR #1034, opened 2025-07-03

JeelsBoobz:
Screenshot_2025-07-03-07-27-22-544_juloo keyboard2

### DIFF
--- a/check_layout.output
+++ b/check_layout.output
@@ -7,7 +7,6 @@ arab_pc_hindu: Layout includes some ASCII punctuation but not all, missing: !, '
 arab_pc_ir: Layout includes some ASCII punctuation but not all, missing: ", %, ', ,, /, ;, <, =, >, ?, [, \, ], `, {, |, }
 beng_national: Layout includes some ASCII punctuation but not all, missing: $
 beng_provat: Layout includes some ASCII punctuation but not all, missing: $, &, *, ., /, <, >, [, \, ], `, {, |, }
-cyrl_jcuken_as: Layout contains unknown keys: combining_acute
 cyrl_lynyertdz_mk: Layout includes some ASCII punctuation but not all, missing: "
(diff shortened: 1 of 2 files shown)


srcs/juloo.keyboard2/Autocapitalisation.java
```diff
@@ -149,7 +149,7 @@ void callback(boolean might_disable)
     _should_disable_shift = might_disable;
     // The callback must be delayed because [getCursorCapsMode] would sometimes
     // be called before the editor finished handling the previous event.
-    _handler.postDelayed(delayed_callback, 1);
+    _handler.postDelayed(delayed_callback, 50);
```
Julow, line 152: What bug is this fixing ? This is an important information that I would like to have as a comment in the code.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -124,6 +124,45 @@ public void onCreate()
     _foldStateTracker.setChangedCallback(() -> { refresh_config(); });
   }
 
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    InputConnection conn = getCurrentInputConnection();
+    if (!_config.move_cursor_with_volume || conn == null || !isInputViewShown()) {
```
Julow, line 130: I think the `conn` check is not necessary. It's not a problem to skip it here as `_keyeventhandler` will do it again and it's not expected to be `null` in normal circumstances.

Julow (commented):
Awesome :)

JeelsBoobz:
> Awesome :)

Can you continue this feature with better code?

Julow:
The reason your PRs are closed automatically is because you force push to the branches that you used to create them. The easiest is to create a different branch for each PR. For example, `git checkout -b volume-key master`. Then you can later go back to `master` and pull other changes (`git checkout master` and `git pull origin master`) and start new PRs.

PR #1039, opened 2025-07-11
update to pull 992

malbajun:


malbajun:
some updating poses for allslavic layout

PR #1040, opened 2025-07-11
replaced combining_acute with combining_aigu

malbajun:
replaced combining_acute with combining_aigu for allslavic


Julow (approved):
Thanks!

PR #1042, opened 2025-07-15
Added Majuscules to the Cyrillic Compose Sequences

anaskaejdar:
The compose sequences lacked majuscules before. With this PR, they will have them, and it will be glorious.

Note of caution, however... I didn't compile the code, as I don't really know how to do that yet, and haven't set up any android dev environment on my computer. Let's check to make sure all is in working order before merging.


Julow:
Thanks for this! I've updated the generated files. Can you try the debug build ? It's at the end of this page: https://github.com/anaskaejdar/Unexpected-Keyboard/actions/runs/16355477910

anaskaejdar:
It works marvellously! Thanks for doing that!

PR #1055, opened 2025-07-20

JeelsBoobz:


Julow (commented):
Sorry for the slow reply! To be fair, I was not happy with the quality of the code and that my remarks in https://github.com/Julow/Unexpected-Keyboard/pull/1034 were not taken into account.

I didn't see anyone else wanting this feature and I don't have the time to do the code, so I'll close this.
Please open again if you make a new implementation.

PR #1058, opened 2025-07-22
Update extra.jsonS

Harry-Ais0:
Ꜩ zTꜨ zTꜨ zT Tz Tz Ꜩ Ꜩ Ꜩ Ꜩ please add the tz (ꜩ Ꜩ) i like that... For the fact theres lots of characters you can make that are found in latin extended's, tz would be a good combination.


Julow (approved):
Thanks!

PR #1062, opened 2025-07-26
Fix alternate keyboard layouts rendering behind navbar on some devices.

cillyvms:
When switching keyboard layouts to emoji or clipboard, for some reason, the insets are not always dispatched, which causes the keyboard to render behind the IME buttons.
In my testing this happens on Android 16 and 15 QPR2, but not on Android 12. Not sure if it's an Android bug or not.

The fix however is simple enough on the app side - just request an inset dispatch every time the IME view changes, so it shouldn't cause any issues on other devices.

| Screenshot_20250727-005409 | Screenshot_20250727-005424 |
|--|--|

### DIFF
--- a/srcs/juloo.keyboard2/Keyboard2.java
+++ b/srcs/juloo.keyboard2/Keyboard2.java
@@ -295,6 +295,7 @@ public class Keyboard2 extends InputMethodService
       ((ViewGroup)parent).removeView(v);
     super.setInputView(v);
     updateSoftInputWindowLayoutParams();
+    v.requestApplyInsets();
   }
 
 


srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -295,6 +295,7 @@ public void setInputView(View v)
       ((ViewGroup)parent).removeView(v);
     super.setInputView(v);
     updateSoftInputWindowLayoutParams();
+    v.requestApplyInsets();
```
Julow, line 298: This requires API 20: https://developer.android.com/reference/android/view/View#requestApplyInsets()
It's done on line 312, for example.

Julow (commented):
Awesome :) Thanks a lot !

srcs/juloo.keyboard2/Keyboard2.java
cillyvms: Hmm, I checked the compatibility, but the minimum API is currently 21. Is there a need for a check here?
https://github.com/Julow/Unexpected-Keyboard/blob/5c7484388fabaf0bfe76265b4304dff345452e36/build.gradle#L17

Julow: You are totally right! It's an old reflex from when the app supported earlier versions.

Julow (approved):
Merging !

PR #1084, opened 2025-08-27
Cobalt & Pine theme

lan700ng:
It was originally a black theme. I added color to it.


lan700ng:
> These themes really look like the Black theme with an accent color. Perhaps you could make a "Monet (Black)" themes that uses the system's colors instead of two themes ?

I've thought about it, but unfortunately, I can't test monet theme because the device is old. That's why I created this.

PR #1093, opened 2025-09-01
Python script encoding fix

Autissima:
Whilst I was trying to add a keyboard layout I encountered issues with the python scripts used to generate the layout lists.

On my setup the open() and xml.elementtree(),write() defaulted to using CP1252, explicitly stating the encoding as UTF-8 in the scripts fixed this.


Julow (approved):
Awesome, thanks !

PR #1096, opened 2025-09-03
Turkish F layout

Autissima:
This adds the Turkish F layout as a layout option to the keyboard. I did take suggestions from someone speaking the language but don't use it myself so I'm not entirely certain what could/should be changed.


srcs/layouts/latin_kbdtuf_tr.xml
```diff
@@ -0,0 +1,47 @@
...
+    <key key0="ü" key1="loc accent_caron" key2="-" key3="_"/>
+    <key key0="t" key2="=" key3="+" key4="loc accent_macron"/>
+    <key key0="k" key3="{" key4="}"/>
+    <key key0="m"/>
+    <key key0="l" key1="loc accent_double_aigu" key3="[" key4="]"/>
+    <key key0="y" key2="accent_aigu" key3="\\" key4="|"/>
+    <key key0="ş"/>
+    <key key0="x" key4="accent_grave"/>
+  </row>
+  <row>
```
Julow, line 35: The third row is not the right size:

shot

You can add `scale="12"` to the row or perhaps you can remove a column by moving some letters to corners ?

Autissima:
As you said moved some letters to corners, also changed some diacritics an iOS keyboard (the person I'm working on with this suggested this)
image

srcs/layouts/latin_kbdtuf_tr.xml
Autissima: As you said moved some letters to corners, also changed some diacritics an iOS keyboard (the person I'm working on with this suggested this)
485618548-9eab4809-7769-4d30-a111-8b172ba858e6

PR #1097, opened 2025-09-03
Upgrade Gradle

JFronny:
While trying to work on a separate improvement, I noticed that the Gradle Wrapper in this seems to be from Gradle 7 based on its checksum.
After I updated that, one thing led to another, and I decided to PR this before working on the PR I actually wanted to do to more cleanly separate these.
First, I updated the Gradle Wrapper and the Gradle Action (as is recommended [in its README](https://github.com/gradle/gradle-build-action)). This also implicitly enables [wrapper validation](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#gradle-wrapper-validation), which should benefit security.
Second, I updated the AGP plugin. The latest Gradle version can absolutely be used with an older AGP, but I decided to just do this while I was at it.
Third, I migrated the build scripts from Groovy to Kotlin, which is now [recommended for Gradle scripts](https://developer.android.com/build/migrate-to-kotlin-dsl), since it enables better syntax highlighting, autocomplete and other IDE support.
I also tried to enable configuration caching, but the `FileOutputStream` in `compileComposeSequences` isn't compatible with that, so I would have had to modify that script.
If you want that, I can certainly try to do so.


build.gradle.kts
```diff
@@ -0,0 +1,170 @@
...
+val checkKeyboardLayouts by tasks.registering(Exec::class) {
+  doFirst { println("\nChecking layouts") }
+  workingDir = projectDir
+  commandLine("python", "check_layout.py")
+}
+
+val compileComposeSequences by tasks.registering(Exec::class) {
+  val out = "srcs/juloo.keyboard2/ComposeKeyData.java"
+  doFirst { println("\nGenerating $out") }
+  val sequences = projectDir.resolve("srcs/compose").listFiles {
```
Julow, line 137: I get these errors and I'm not sure how to resolve them:

```
e: unexpected-keyboard/build.gradle.kts:137:54: Overload resolution ambiguity: 
public open fun listFiles(p0: FileFilter!): Array<(out) File!>! defined in java.io.File
public open fun listFiles(p0: FilenameFilter!): Array<(out) File!>! defined in java.io.File
e: unexpected-keyboard/build.gradle.kts:138:6: Unresolved reference: it
e: unexpected-keyboard/build.gradle.kts:138:34: Unresolved reference: it
e: unexpected-keyboard/build.gradle.kts:139:13: Unresolved reference: it
```

I'm using `8.14.3` and I can't upgrade easily.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
-#Mon Aug 21 18:13:41 CEST 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.0.0-bin.zip
```
Julow, line 3: My distribution (NixOS) distributes gradle `8.14.3` and I'd prefer if we use that instead of gradle `9`. The reason is that I'd wish to reproduce the wrapper update.

Julow (commented):
Thanks!

build.gradle.kts
```diff
@@ -0,0 +1,170 @@
...
+  commandLine("python", "gen_layouts.py")
+}
+
+val checkKeyboardLayouts by tasks.registering(Exec::class) {
+  doFirst { println("\nChecking layouts") }
+  workingDir = projectDir
+  commandLine("python", "check_layout.py")
+}
+
+val compileComposeSequences by tasks.registering(Exec::class) {
```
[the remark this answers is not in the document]
JFronny: I noticed that I can just defer the creation of the stream, which seems to work and doesn't break the configuration cache

gradle/wrapper/gradle-wrapper.properties
JFronny: [This PR](https://github.com/NixOS/nixpkgs/pull/421047) seems to suggest that NixOS recently updated to Gradle 9, so this should be fixed with a system update

Julow: These take a long time to get into the unstable branch, and even more time to get into my system. I'd prefer to avoid compatibility issues for contributors.

build.gradle.kts
JFronny: This is because 8.14.3 ships an outdated version of Kotlin that seemingly doesn't handle this case properly.
Since I have now downgraded Gradle, this has been fixed by providing more type information.

gradle/wrapper/gradle-wrapper.properties
JFronny: I have now downgraded Gradle by invoking the wrapper again.
I had to call `gradlew wrapper` twice until it actually downgraded the jar, but it should now be using consistent versions for the jar, properties, and scripts.

Julow (approved):
Sorry for the slow reply. This is perfect ! Thanks a lot

PR #1110, opened 2025-09-29
Enable inline autofill

JFronny:
Since Android 11, Android allows keyboards [to display autofill suggestions inline](https://developer.android.com/identity/autofill/ime-autofill) instead of trying to show a popup.
Based on some code in [AnySoftKeyboard](https://github.com/AnySoftKeyboard/AnySoftKeyboard), this PR introduces support for that feature here.

To do so, I had to modify the default keyboard layout to include a scrollable view in which the components provided by the framework are displayed. I based this on the layout for the emoji pane, but I must admit that I'm not completely happy with the comparatively large number of additional layout bits the java code now needs to keep track of.
(I could have also tried to reimplement their behavior manually, but that seemed overly complicated).

If you have an idea for a better approach the layout and its handling could be modified to take, please suggest it. It has been quite a while since I wrote any android UI stuff without compose, so I'm almost certain I forgot about something.

I set an `android:background` on the wrapping layout element in an attempt to get keyboard background opacity to work, which seems to have been successful, but I only really tried it in the keyboard app and the browser, both of which have a single-color background, so if you have another app with a more colorful background to try this with, that would be appreciated.

With that said, this implementation of the feature does seem to work quite well in the emulator.


srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -520,4 +545,66 @@ private View inflate_view(int layout)
...
+            new InlinePresentationSpec.Builder(smallestSize, biggestSize)
+                    .setStyle(stylesBundle)
+                    .build();
+
+    List<InlinePresentationSpec> specList = new ArrayList<>();
+    specList.add(spec);
+
+    InlineSuggestionsRequest.Builder builder = new InlineSuggestionsRequest.Builder(specList);
+
+    return builder.setMaxSuggestionCount(InlineSuggestionsRequest.SUGGESTION_COUNT_UNLIMITED)
```
Julow, line 573: This is already the default for `setMaxSuggestionCount` so I think it can be removed.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -520,4 +545,66 @@ private View inflate_view(int layout)
...
+    Bundle stylesBundle = stylesBuilder.build();
+    InlinePresentationSpec spec =
+            new InlinePresentationSpec.Builder(smallestSize, biggestSize)
+                    .setStyle(stylesBundle)
+                    .build();
+
+    List<InlinePresentationSpec> specList = new ArrayList<>();
+    specList.add(spec);
+
+    InlineSuggestionsRequest.Builder builder = new InlineSuggestionsRequest.Builder(specList);
```
Julow, line 571: You can do `Collections.singletonList(spec)` to shorten the list building code.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -520,4 +545,66 @@ private View inflate_view(int layout)
...
+    for (InlineSuggestion inlineSuggestion : inlineSuggestions)
+    {
+      try
+      {
+        inlineSuggestion.inflate(
+                this,
+                autofillSize,
+                getMainExecutor(),
+                inlineContentView -> {
+                  _autofillContainer.setVisibility(View.VISIBLE);
```
Julow, line 599: `setVisibility` should perhaps be called outside of the loop.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -520,4 +545,66 @@ private View inflate_view(int layout)
...
+
+    return builder.setMaxSuggestionCount(InlineSuggestionsRequest.SUGGESTION_COUNT_UNLIMITED)
+            .build();
+  }
+
+  @RequiresApi(api = Build.VERSION_CODES.R)
+  @Override
+  public boolean onInlineSuggestionsResponse(@NonNull InlineSuggestionsResponse response)
+  {
+    List<InlineSuggestion> inlineSuggestions = response.getInlineSuggestions();
```
Julow, line 581: The code that manages the autofill container views and manages the reponse should be moved to an other class to avoid making this class bigger.

res/layout/keyboard.xml
```diff
@@ -1,2 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
-<juloo.keyboard2.Keyboard2View xmlns:android="http://schemas.android.com/apk/res/android" android:hardwareAccelerated="false" android:background="?attr/colorKeyboard"/>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android" android:orientation="vertical" android:hardwareAccelerated="false" android:background="#00000000">
```
Julow, line 2: I'm working on a suggestion strip for a future word suggestion feature, in this branch: https://github.com/Julow/Unexpected-Keyboard/compare/master...candidates_view

Do you have ideas how to make these compatible ? For example, it has a fixed width instead of being scrollable and it has a fixed number of spots for suggestions (3).

Julow (commented):
Awesome! Do you have examples of apps that send autofills that I can test ?

sdrapha:
> Awesome! Do you have examples of apps that send autofills that I can test ?

Bitwarden is one.
Most likely, all other passwd managers should have similar functionality.

JFronny:
I used [this test page](https://fill.dev/form/login-simple) with KeePassDX to quickly validate changes in Chrome and Firefox.

res/layout/keyboard.xml
Julow: Does the autofill makes sense after the user has started typing on the keyboard ? Perhaps the autofill container could be hidden in favor of the suggestion strip after the first key was pressed ?

srcs/juloo.keyboard2/InlineAutofill.java
```diff
@@ -0,0 +1,90 @@
...
+@RequiresApi(api = Build.VERSION_CODES.R)
+public class InlineAutofill {
+  private HorizontalScrollView _autofillContainer;
+  private LinearLayout _autofillSuggestions;
+
+  public void onInflate(View keyboardViewParent)
+  {
+    _autofillContainer = keyboardViewParent.findViewById(R.id.autofill_container);
+    _autofillSuggestions = keyboardViewParent.findViewById(R.id.autofill_suggestions);
+    _autofillContainer.setVisibility(View.GONE);
```
Julow, line 38: It's already GONE when inflating.

srcs/juloo.keyboard2/InlineAutofill.java
```diff
@@ -0,0 +1,90 @@
...
+  {
+    List<InlineSuggestion> inlineSuggestions = response.getInlineSuggestions();
+
+    float height =
+            TypedValue.applyDimension(
+                    TypedValue.COMPLEX_UNIT_DIP, 40, context.getResources().getDisplayMetrics());
+    Size autofillSize = new Size(ViewGroup.LayoutParams.WRAP_CONTENT, ((int) height));
+
+    _autofillSuggestions.removeAllViews();
+    if (!inlineSuggestions.isEmpty()) _autofillContainer.setVisibility(View.VISIBLE);
```
Julow, line 71: I think the visibility should be set to GONE when the suggestions are empty. The view is otherwise visible as an empty thin white strip.

Julow (commented):
Thanks! I could test with KeePassDX

res/layout/keyboard.xml
JFronny: An implementation of this that piggybacks on the existing suggestions code has been implemented in [`5f9baf0` (this PR)](https://github.com/Julow/Unexpected-Keyboard/pull/1110/changes/5f9baf0bb33f9a12a9862541d5b9f80ead6be381).
It isn't perfect (when writing a new word, for as long as the word is too small for a suggestion, the autofill shows back up), but this is the cleanest I could get it without using the IDE (the package setup is broken).

JFronny:
Sorry for taking so long with this, I kind of forgot the PR even existed.
I have now updated it to merge your changes and attempted to implement your suggestion.

Issue #983, opened 2025-04-16
Question: Can custom compose key combination?


Houfalafel:
Can I make my own combinations for compose?
Is there like a code I can add to the layout xml?
I want to type these specifically:
`compose` + `n` + `g` + `<dead_tilde>` = n͠g 
`compose` +  `shift` + `n` + `g` + `<dead_tilde>` = N͠g
`compose` + `g` + `<dead_tilde>` = g̃

These are characters that were used historically in the Philippines. I want to make my own combinations because compose is cool. Also, adding it as a swipe thing could clutter my keyboard as I only use these like twice a month.

Julow:
These characters should be added to the compose key for everyones.
Could you do it ? There are some explanations here: https://github.com/Julow/Unexpected-Keyboard/blob/master/CONTRIBUTING.md#adding-symbols-to-shift-fn-compose-and-other-modifiers

I'll re-open because a custom layout is not the solution.

Julow:
I've added these compose sequences using the ASCII `~` symbol in d6dfb5c1. We can add many more sequences if needed.

It's not possible to use `dead_tilde` in sequences currently but I'd like to fix that. I'll keep this issue open until this is resolved.

equaa:
Would you accept a PR for an app-customizable system for custom compose sequences, similar to custom layouts, or do you prefer to keep any additions upstream?

Julow:
@equaa That would be fantastic !

PR #1118, opened 2025-10-19
Add compose sequence for ŋ and change sequences for n͠g and Ğ

Houfalafel:
ŋ is used more frequently than n͠g, so I think it should take n + g instead of the latter. For n͠g, it instead gets g + n + ~. This pull request also moves Ğ and adds the lowercase.


Julow (approved):
Thanks !

For the record, the previous compose sequence for n͠g was added in d6dfb5c1 after the discussion in https://github.com/Julow/Unexpected-Keyboard/issues/983

PR #1127, opened 2025-11-10
Fix insets being excluded from computed width

### LINKED ISSUE #1124: Landscape leaves empty space on the right since a08db256
Commit 4ed8594794691e5991c414337fd4ede7a9e857cc:

Image

Commit a08db256611a66ca14e085fbd62a8fe352847e89:

Image

Android 15 (OnePlus/OxygenOS); those screenshots (clearly) being with button navigation, but with gesture navigation it ends up with the empty space too (though without the buttons the keyboard is at least more centered)

dzaima:
Fixes #1124

### DIFF
--- a/srcs/juloo.keyboard2/Keyboard2View.java
+++ b/srcs/juloo.keyboard2/Keyboard2View.java
@@ -272,6 +272,7 @@ public class Keyboard2View extends View
     _marginLeft = Math.max(_config.horizontal_margin, _insets_left);
     _marginRight = Math.max(_config.horizontal_margin, _insets_right);
     _marginBottom = _config.margin_bottom + _insets_bottom;
+    width += _insets_left + _insets_right;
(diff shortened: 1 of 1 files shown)


Julow (approved):
Thanks a lot :)

Julow:
I had to revert this in https://github.com/Julow/Unexpected-Keyboard/pull/1265
I tested with Android 15, 16 and 17 and I found no related issue. I think your case is more complicated.

dzaima:
yeah, seems so.. presumably some bug with oxygenOS

Julow:
Do you still have the issue with the new release ? That's maybe a false hope because the only thing that happened since is the update to SDK version 36.

dzaima:
Yeah, issues on landscape still persist with the release 2.0.1 apk.

PR #1128, opened 2025-11-16
Ukrainian l10n update

andyca111100:
An update for Ukrainian translation


andyca111100:
I noticed that Weblate is missing some strings. Update them if possible.

PR #1129, opened 2025-11-16
Added keyboards

10Develops:
The almost finalized improvements. Only QZERTY keyboard is still.

Edit: Improvements finalized. Added QZERTY layout for Italian and Latin.


10Develops (commented):
It was fixed.

Issue #1166, opened 2026-01-31
buggy custom layout freezes app


axel-surfer:
With this ( **not** **usefull** ) additional modmap-config line:

_<ctrl a='left  b='left:ctrl,left_' /> 

then pressing 'ctrl' and 'left'

the keyboard freezes ~ 5s ... and restarts automatically the last layout again. 

Ok it is only a config-bug which could be fixed... and after restarting / loading the last layout I could work again :-) 

Maybe is it possible to prevent this or simular misconfiguration in the app? With a validation-function(?) to prevent circles/loops like that?

Julow:
Thanks for your issue! I think the best thing to do is to not take modmaps into account when evaluating macros: https://github.com/Julow/Unexpected-Keyboard/pull/1167

PR #1167, opened 2026-01-31
Avoid loops in modmaps

Julow:
Fix https://github.com/Julow/Unexpected-Keyboard/issues/1166

It was possible to define a loop that would crash the app with the following modmap:

    <ctrl a="left" b="left:ctrl,left"/>

With this change, modmap are not taken into account while evaluating macros. The modmap above modifies ctrl+left into ctrl+left.


axel-surfer:
😊👍

PR #1172, opened 2026-02-05
feat: Split keyboard layout

ribhar:
Adds a split keyboard layout option for landscape mode on wide screens.

Features:
- Split layout with adjustable gap.
- Dual space bars.

Screenshots:
Screenshot_20260204_215711_Unexpected Keyboard (Debug)

### DIFF
--- a/srcs/juloo.keyboard2/Keyboard2View.java
+++ b/srcs/juloo.keyboard2/Keyboard2View.java
@@ -247,15 +247,14 @@ public class Keyboard2View extends View
     float x = _marginLeft;
     if (row == null || tx < x)
       return null;
+    
     for (KeyboardData.Key key : row.keys)
     {
-      float xLeft = x + key.shift * _keyWidth;
-      float xRight = xLeft + key.width * _keyWidth;
-      if (tx < xLeft)
-        return null;
-      if (tx < xRight)
+      x += key.shift * _keyWidth;
+      float keyW = key.width * _keyWidth;
+      if (tx >= x && tx < x + keyW)
         return key;
-      x = xRight;
+      x += keyW;
     }
     return null;
   }
@@ -275,6 +274,7 @@ public class Keyboard2View extends View
     _marginRight = Math.max(_config.horizontal_margin, _insets_right);
     _marginBottom = _config.margin_bottom + _insets_bottom;
     width += _insets_left + _insets_right;
+
     _keyWidth = (width - _marginLeft - _marginRight) / _keyboard.keysWidth;
     _tc = new Theme.Computed(_theme, _config, _keyWidth, _keyboard);
     // Compute the size of labels based on the width or the height of keys. The
@@ -350,6 +350,7 @@ public class Keyboard2View extends View
       y += row.shift * _tc.row_height;
       float x = _marginLeft + _tc.margin_left;
       float keyH = row.height * _tc.row_height - _tc.vertical_margin;
+
       for (KeyboardData.Key k : row.keys)
       {
         x += k.shift * _keyWidth;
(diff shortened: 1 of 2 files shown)


srcs/juloo.keyboard2/Keyboard2View.java
```diff
@@ -247,10 +249,45 @@ private KeyboardData.Key getKeyAtPosition(float tx, float ty)
     float x = _marginLeft;
     if (row == null || tx < x)
       return null;
-    for (KeyboardData.Key key : row.keys)
+    int middleIndex = _isSplit ? row.keys.size() / 2 : -1;
```
Julow: I'm concerned with the added complexity into the drawing code. This could be made once by creating a modified version of `KeyboardData`, like it's done in `addNumPad`. What do you think ?

Julow (commented):
This is a feature I wished to add for a long time. Thanks !

srcs/juloo.keyboard2/Keyboard2View.java
ribhar: Thanks for the review! I agree that modifying `KeyboardData` directly is a much cleaner approach than handling the split logic in the drawing code. I'll refactor the implementation to transform the `KeyboardData` layout when split mode is enabled, similar to how `addNumPad` works.

akaoio:
Please make it split into 2 panels, not just one panel because it wastes the area in the center and it locks both left and right side vertically. Having 2 float panels, it is possible to have the left and right sides at 2 different positions and there is no center space that blocks user vision.
Thank you so much for your contributions to the world you are amazing people.

dzaima:
> Please make it split into 2 panels

Is there any keyboard that does this? Even Google's Gboard's split mode does one panel with a blocked center, so it might just not be possible.

(there might exist some options on making the center more transparent? Would need to also be a floating window for most effect, and still wouldn't allow clicking through)

akaoio:
> > Please make it split into 2 panels
> 
> Is there any keyboard that does this? Even Google's Gboard's split mode does one panel with a blocked center, so it might just not be possible.
> 
> (there might exist some options on making the center more transparent? Would need to also be a floating window for most effect, and still wouldn't allow clicking through)

Yeah actually it is just an idea, I don't have Java nor Android background. I believe you guys will do your best. Thank you.

srcs/juloo.keyboard2/Keyboard2View.java
ribhar: Hi @Julow, I've refactored the split keyboard feature based on your feedback about complexity in the drawing code. Here's what changed:

**Key improvements:**
✅ Centralized layout transformation: Split logic now lives in [LayoutModifier.split_layout()](vscode-file://vscode-app/c:/Users/ribha/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), similar to addNumPad()
✅ Cleaner view code: Removed all split-specific rendering and touch detection from Keyboard2View
✅ Simplified methods: onMeasure(), getKeyAtPosition(), and onDraw() no longer track gaps
✅ Preserved behavior: Current keyboard behavior is fully maintained—keys display and respond identically
Technical approach:
Instead of handling split mode in the drawing/touch code, the KeyboardData is now transformed at build time when both split_keyboard and [wide_screen](vscode-file://vscode-app/c:/Users/ribha/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) are enabled. This keeps all layout modifications (split, numpad, etc.) in one place.

**Testing:**
Tested on device—split layout displays correctly, keys respond properly, and normal keyboard mode is unaffected.

This addresses your concern about reducing drawing code complexity while keeping the feature fully functional. Ready for your review!

Julow:
I've implemented this by modifying the layout in https://github.com/Julow/Unexpected-Keyboard/pull/1334 Feedback welcome!

Issue #1185, opened 2026-02-23
Expanded key action XML (& more)


dzaima:
Would there be interest in an expanded XML way to do key actions? e.g. instead of
```xml
<key e="thingy:'custom format with custom escapes and logic'"></key>
```
you could so something like
```xml
<key>
  <e legend="thingy">
    <keyevent>123</keyevent>
    <action>ctrl,a</action>
    <type>hello world</type>
  </e>
</key>
```

Certainly not the most concise thing (still would want to keep the existing system for simple cases (and of course for backwards compat)), but would be nice for programmatically-generated layouts, and would fill in the holes of things currently unrepresentable. And would allow for a simple way to add extensions, e.g. a `<switch-layout>specific-layout-to-switch-to-by-name</switch-layout>`, or a `repeat="true"` attribute that one could add to backspace if they want only it to autorepeat and nothing else.

Half-relatedly, from hacking in a switch-layout action in the current system for myself - the `KeyValue` `_payload` system seems kinda odd to me. I think it'd be nicer to do it via subclasses of `KeyValue` (kinda replacing `enum Kind` by said subclasses, though it wouldn't necessarily need to be done that way). Perhaps feels a bit OOPy, but basically a tagged union just done inside-out.

Some other smaller(-ish) things:
- Running a `git submodule update --init` or similar seems to be necessary for `./gradlew assembleDebug` to work, but this isn't noted anywhere in `CONTRIBUTING.md`, nor does it happen automatically.
- The custom layout XML text editor view seems to take up 100% of a CPU core; I unfortunately have failed to get any setup for profiling (or debugging in general) working to debug this myself, but it's probably worth noting if you have some idea about it.
- Opening the clipboard manager, and closing it with the "ABC" button, makes the background transparent, after which closing and reopening the keyboard will result in nothing at all being shown; bisected to 98c1b8db, may investigate further later.

Julow:
Keys don't live in layouts, they can appear in `<modmap>` or in the app's settings. Do you need this for a feature that you are designing ?

We don't need to invent keys that are unrepresentable in the short syntax. Also, the syntax can be changed if needed.

[...]

PR #1192, opened 2026-02-25
Make layout text editor not redraw every frame

dzaima:
Technically shouldn't change anything that wasn't already broken I think (assuming `setPadding` works how I think it does), but who knows. Works in my testing at least.

Fixes part of #1185

### DIFF
--- a/srcs/juloo.keyboard2/CustomLayoutEditDialog.java
+++ b/srcs/juloo.keyboard2/CustomLayoutEditDialog.java
@@ -98,6 +98,9 @@ public class CustomLayoutEditDialog
 
     /** A mutable Rect object that is used during onDraw. */
     Rect _clip_bounds = new Rect();
+    
+    /** The currently-set left padding amount */
+    int _prev_padding = Integer.MIN_VALUE;
 
     @Override
     protected void onDraw(Canvas canvas)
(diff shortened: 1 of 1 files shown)


srcs/juloo.keyboard2/CustomLayoutEditDialog.java
```diff
@@ -98,19 +98,26 @@ public void set_on_text_change(OnChangeListener l)
...
     @Override
     protected void onDraw(Canvas canvas)
     {
       float digit_width = _ln_paint.measureText("0");
       int line_count = getLineCount();
       // Extra '+ 1' serves as padding.
-      setPadding((int)(((int)Math.log10(line_count) + 1 + 1) * digit_width), 0, 0, 0);
+      int padding = (int)(((int)Math.log10(line_count) + 1 + 1) * digit_width);
+      if (padding != _prev_padding) {
+        setPadding(padding, 0, 0, 0);
```
Julow: That should work but maybe this can be done in `onMeasure` ? I suppose it should be called when the content changes. 
With this, I think `digit_width` could also be computed less often.

dzaima: I'm not an android dev, I don't know what a `onMeasure` is, I just move existing things around :) From some quick searching, I don't think `onMeasure` really relates to any goal here?

But yeah the main (if not only? given that `_ln_paint.setTextSize(…)` is done just once, in the constructor; unless `Paint` references instead of copying some constructor arg paint parts idk) padding update reason I think should be when the number of lines changes. Which could presumably be done in `onTextChanged` I suppose? (`digit_width` could maybe just be moved to the constructor?)

Julow: You are right, we should probably call `setPadding` in `onTextChanged`. I was worried that maybe `getLineCount` would return the old value during `onTextChanged` but that's probably not the case.
`_ln_paint.setTextSize(…)` can definitely be done in the constructor.

dzaima: Looks like `onTextChanged` gets called with an updated `getLineCount()` on changes, but `getLineCount()` doesn't give the correct value on initial text insertion (even with doing it after the `input.setText(initial_text);`); and separating things leads to a cascade of requiring either duplicating or moving to a field the `line_count` and `digit_width` variables.

I think the current version should be fine, these would be very tiny changes anyway; at the very least just leaving the editor open doing nothing shouldn't cause the device to heat up and battery to drain noticeably fast :)

Julow (approved):
Thanks a lot !

Issue #1169, opened 2026-02-02
on ePaper Black theme, swiping the keys towards the corners makes the middle key white (which makes the middle letter invisible) instead of black  


Zareeff:
I don't know if this is intentional or a bug but it does't feel right.

![Screenshot_2026-02-02-17-25-05-418_com.miui.notes.jpg](https://github.com/user-attachments/assets/d7bd7fa5-a44a-4bf0-970c-cc8788a511cb)

Julow:
Hi! Thanks for your issue. This is fixed in https://github.com/Julow/Unexpected-Keyboard/pull/1173

dzaima:
This has the downside that, on themes where all colors are distinguishable, it no longer shows which corner is being activated. Importantly, if there are multiple modifier keys on a key (e.g. the ctrl/meta key), it's impossible to see which are active.

Perhaps it'd make sense to have separate label colors for "on pressed key, but not specifically active" vs "active"?

Julow:
That would be nice! Could you add this ?

PR #1193, opened 2026-02-25
Separate colors for pressed vs activated labels

dzaima:
Keeps the same color scheme as after #1173 for the ePaper Black theme, while preserving all other themes as they were before that PR.

As per #1169


Julow (approved):
Awesome! Thanks a lot!

PR #1199, opened 2026-03-07
Making dictionaries compatible with 16 KB devices.

xxcavemanxx:
Related to apk compatibility with 16 KB devices for dictionaries.

APK Unexpected-Keyboard-debug.apk is not compatible with 16 KB devices. Some libraries have LOAD segments not aligned at 16 KB boundaries: lib/arm64-v8a/libcdict_java.so Starting November 1st, 2025, all new apps and updates to existing apps submitted to Google Play and targeting Android 15+ devices must support 16 KB page sizes. For more information about compatibility with 16 KB devices, visit developer.android.com/16kb-page-size.


Julow (approved):
Awesome! Thanks!

PR #1219, opened 2026-04-01
Fix building on NixOS

exzombie:
This bumps the platform in `shell.nix` to match the target platform. It also exports an environment variable that makes gradle use `aapt2` from Nix. Without this, it tries to run the downloaded one, which fails on NixOS. These changes allowed me to run `./gradlew assembleDebug` after entering `nix-shell`.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -36,26 +36,16 @@ let
 
   ANDROID_SDK_ROOT = "${android.androidsdk}/libexec/android-sdk";
 
-  gradle = pkgs.gradle.override { java = jdk; };
-  # Without this option, aapt2 fails to run with a permissions error.
-  gradle_wrapped = pkgs.runCommandLocal "gradle-wrapped" {
-    nativeBuildInputs = with pkgs; [ makeBinaryWrapper ];
-  } ''
-    mkdir -p $out/bin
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -58,4 +58,5 @@ in pkgs.mkShell {
   ];
   JAVA_HOME = jdk.home;
   inherit ANDROID_SDK_ROOT;
+  GRADLE_OPTS = "-Dorg.gradle.project.android.aapt2FromMavenOverride=${ANDROID_SDK_ROOT}/build-tools/${build_tools_version}/aapt2";
```
Julow, line 50: This is also in the wrapper (`gradle_wrapped` above). I think you can remove the wrapper entirely.

Julow (approved):
Thanks!

shell.nix
exzombie: Dang, I completely missed that, thanks! I seldom build Android apps, so I just expected `./gradlew` to work, and didn't look further after I found the right option. I have now removed the wrapper as I think most people expect `./gradlew` to work.

Julow (approved):
Thanks a lot!

PR #1221, opened 2026-04-04
Add Finnish QWERTY keyboard layout XML

bqv:
Language support would be nice too.


Julow (approved):
Perfect, thanks!

I've added the language support.

Issue #1224, opened 2026-04-07
Dictionary offers en-US only


bqv:
This is just documentation of how I fixed the issue. 

My install only offered me english (united states) for a dictionary, despite me being the furthest thing from american.

It turns out that OxygenOS only offers the language part of a locale, and sets the country part to the local region, so my system_locales prop was set to e.g. "en_FI,fi_FI,sv_FI". 

Manually changing that en_FI to en_GB using Setedit (app), and then rebooting, fixed it. Now even the keyboard options correctly include "English (United Kingdom)", and the dictionary is available too.

I guess I'm an edge case. Shan't be using en_US though, and horrified to find I have been.

Julow:
You should also have a Finish dictionary, right ?

You can add `en_FI` as a language supported by the keyboard: https://github.com/Julow/Unexpected-Keyboard/blob/master/gen_method_xml.py#L13
This way, you can select the right dictionary: `loc("en_FI", ..., dictionary="en_GB")`. You can also add extra keys like €, ä (or `accent_trema`), etc..

`en_US` is the default when the full language tag is not recognized and starts with `en`. Perhaps it's a bad default but I think the fix is to add more languages to the list.

bqv:
I'm not sure everyone would agree with me, but I feel like en_GB would should be default for europe, majority of africa, and most of asia, as it's the version that is officially taught in those places... That said, lots of people seem to prefer american now. 

It's probably rather best left an option, maybe you could offer all regional variants rather than just the one reccomended one? Then even e.g. a frenchman that lives in canada, isn't forced to use quebecois (fr_CA) and can spellcheck to fr_FR instead

Julow:
> It's probably rather best left an option, maybe you could offer all regional variants rather than just the one reccomended one? Then even e.g. a frenchman that lives in canada, isn't forced to use quebecois (fr_CA) and can spellcheck to fr_FR instead

That would be a good UX but maybe few users need it. In the frenchman example, you'd add fr_FR in your phone settings. The options are useful when the keyboard doesn't have the language in the list.

I'd be happy to use en_GB as a default for `en_*` languages though

Julow:
What do you think of also adding `en_FI` to the supported locales ? With extra keys for example.

bqv:
It feels wrong to me, as English isn't technically an official language here, it's just highly spoken. I would definitely add sv_FI though, as Swedish actually is the other official language

Julow:
[...]

PR #1237, opened 2026-04-13
Change default locale for en_* and update IME layout

bqv:
As mentioned in #1224

### DIFF
--- a/gen_method_xml.py
+++ b/gen_method_xml.py
@@ -80,7 +80,7 @@ LOCALES = [
 ]
 
 # The locale that is at the beginning of the list
-DEFAULT_LOCALE = loc("en", "latin", "latn_qwerty_us", tag="en", dictionary="en_US")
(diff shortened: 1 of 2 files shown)


gen_method_xml.py
```diff
@@ -80,7 +80,7 @@ def loc(loc_name, script, default_layout, **kwargs):
 ]
 
 # The locale that is at the beginning of the list
-DEFAULT_LOCALE = loc("en", "latin", "latn_qwerty_us", tag="en", dictionary="en_US")
+DEFAULT_LOCALE = loc("en", "latin", "latn_qwerty_gb", tag="en", dictionary="en_GB")
```
Julow: I'm not sure the layout change. The UK layout is the same as the US one with some punctuation and other symbols moved around.
This change can only annoy users. Would it be so bad to use the UK dictionary and the US layout ?

bqv: oh, sure!

Julow (approved):
Awesome, thanks!

PR #1238, opened 2026-04-14
Add Japanese JIS soft layout

sandy2008:
## Summary
- add a new built-in `JIS (Japan)` soft layout as `latn_jis_jp`
- register `ja_JP` as a supported subtype defaulting to `latn_jis_jp`
- update generated layout and IME resource lists

## Why
- expose a Japanese locale/subtype and a compatibility-safe JIS soft layout without changing the existing `QWERTY (Japan)` layout id or behavior

## Validation
- `python3 gen_layouts.py`
- `python3 gen_method_xml.py > res/xml/method.xml`
- `python3 check_layout.py`


Julow (commented):
Thanks for your PR. This is just the QWERTY layout. I don't think it's worth adding a Japanese layout that just opens the QWERTY layout with no additional feature. This should be added at the same time as a full Japanese support.

PR #1146, opened 2025-12-21
Fix various linting issues

Julow:
Fix various lint issues reported by `gradle lint`.

I initially intended to check the lint-baseline file into the repository and to catch new lint issues using a github ci check. I decided not to do this because the lint-baseline file is very unstable (it's likely to change after every changes) and contains mostly false positives.

Issue #942, opened 2025-03-15
Use system styling in settings


ghost:
A few parts of Unexpected Keyboard use system styling so it could be expanded for use in the settings.

PR #1240, opened 2026-04-14
Restore Prefs Themeing

bqv:
I like dark mode, I believe #1146 may have killed it

### DIFF
--- a/res/values/styles.xml
+++ b/res/values/styles.xml
@@ -121,7 +121,7 @@
     <item name="android:layout_width">24dp</item>
     <item name="android:layout_height">24dp</item>
   </style>
-  <style name="appTheme" parent="@android:style/Theme.Material.DeviceDefault.DayNight"/>
+  <style name="appTheme" parent="@android:style/Theme.Material"/>
   <style name="settingsTheme" parent="appTheme">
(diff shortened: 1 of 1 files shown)


Julow:
This doesn't work for me. It's using the dark theme independently from the system theme

bqv:
I'm not sure how to fix this then. I'd love to use the pre-releases, but at the moment I'm stuck waiting for a full release, mainly because the debug builds don't properly upgrade on my phone - no signature (I really don't know why they refuse to install as test apps, on oxygenos), but also because of this style issue.

If you can fix this, I'd really appreciate it, and I'd have more incentive to test/pr

Julow:
The debug build installs as a separate application from the release. You should see it in the setting page for enabling the keyboard.
I'll make more frequent releases once the spell checking is released.

Julow:
Related issue: https://github.com/Julow/Unexpected-Keyboard/issues/942

bqv:
Once again, please see if you can fix this, it's a shame of a regression

Julow (commented):
This doesn't work for me, it just sets the dark theme independently from the system theme.

Julow:
Sorry I said the same thing before. I haven't found an easy solution after some time. I'll close this and try again in the future.

PR #1250, opened 2026-04-26
Add optional AZERTY FR layout with larger bottom joystick

dianedef:
## Summary
- add a new optional layout: `AZERTY (Français, Big joystick)`
- keep existing `AZERTY (Français)` unchanged
- define a custom bottom row in the new layout with a larger joystick key:
  - joystick key width: `2.2` (was `1.1` in default bottom row)
  - row rebalanced by reducing nearby key widths (`ctrl`, `fn`, `space`, `enter`)
- regenerate `res/values/layouts.xml` so the new layout appears in Settings

## Why
This provides an accessibility/ergonomics-oriented option for users who want larger directional controls, without changing defaults for everyone else.

## Scope
- No change to official default AZERTY behavior unless the new layout is explicitly selected.

## Testing
- Not tested locally (Android SDK not configured in this environment).
- XML parsing and layout listing generation were updated via `python3 gen_layouts.py`.

### DIFF
--- a/res/values/layouts.xml
+++ b/res/values/layouts.xml
@@ -45,6 +45,7 @@
     <item>latin_kbdtuf_tr</item>
     <item>latn_azerty_be</item>
     <item>latn_azerty_fr</item>
+    <item>latn_azerty_fr_big_joystick</item>
     <item>latn_bepo_fr</item>
     <item>latn_bone</item>
     <item>latn_neo2</item>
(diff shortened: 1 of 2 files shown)


Julow (commented):
Sorry but I don't want to add layout variants into the app at the moment. The reason is that the layout list is already hard to use and that layout variants increase maintenance cost by not being uniform with other layouts.

Can you submit it to https://github.com/Julow/Unexpected-Keyboard-layouts instead ?

PR #1251, opened 2026-04-26
Improve clipboard history pane truncation and retention

dianedef:
## Summary
- limit each "Recently copied text" entry preview to 5 lines in the clipboard pane
  - prevents very long clips from pushing action buttons off-screen
- improve clipboard history retention behavior
  - increase max in-memory history from 6 to 12 entries
  - keep at least 5 entries even when older entries reach expiration

## Why
Large clipboard entries can make the history pane hard to use because buttons become harder to reach.
Also, users can quickly lose recent items with the current retention behavior.

## Testing
- Not tested locally (Android SDK not configured in this environment).
- Changes are localized to UI XML and history service logic.

### DIFF
--- a/srcs/juloo.keyboard2/ClipboardHistoryService.java
+++ b/srcs/juloo.keyboard2/ClipboardHistoryService.java
@@ -47,10 +47,14 @@ public final class ClipboardHistoryService
       _paste_callback.paste_from_clipboard_pane(clip);
   }
 
+  /** Keep at least this many entries around even if old entries have reached
+      their configured expiration time. */
+  public static final int MIN_HISTORY_SIZE = 5;
+
   /** The maximum size limits the amount of user data stored in memory but also
       gives a sense to the user that the history is not persisted and can be
       forgotten as soon as the app stops. */
-  public static final int MAX_HISTORY_SIZE = 6;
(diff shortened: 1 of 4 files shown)


srcs/juloo.keyboard2/ClipboardHistoryService.java
```diff
@@ -158,6 +182,47 @@ int get_history_ttl_minutes() {
     return Config.globalConfig().clipboard_history_duration;
   }
 
+  synchronized void restore_history()
```
Julow, line 185: No. The clipboard history must not be stored. This is for security and simplicity reason.

srcs/juloo.keyboard2/ClipboardHistoryService.java
```diff
@@ -47,21 +50,32 @@ public static void paste(String clip)
       _paste_callback.paste_from_clipboard_pane(clip);
   }
 
+  /** Keep at least this many entries around even if old entries have reached
+      their configured expiration time. */
+  public static final int MIN_HISTORY_SIZE = 5;
```
Julow, line 55: The history expiration is a privacy feature, so keeping a minimum number of entry doesn't make sense.
I think you might want to add an option to control this with values such as "5 minutes", "30 minutes", "until the app closes".

PR #1273, opened 2026-05-04
Fix missing null gate

FlippingBinary:
This adds a null gate to protect against a null pointer exception.


Julow (approved):
This is fantastic! Thanks a lot!

Julow:
In case you need to check the debug build, it's here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/25330918807

PR #1279, opened 2026-05-05
Update README

alotbsol555:
Removed mention of network requests from the README, due to dictionary downloads.


Julow (approved):
Good catch, thanks! I should also update the store description.

Issue #1259, opened 2026-05-03
Crash when navigating around emoji


knoellle:
Steps to reproduce:
- Insert emoji into a text field, e.g. 👀
- Use either the arrow keys or the spacebar slide mechanism to move the cursor around the emoji character, specifically upon moving to the position right after the emoji.
- Crashes

Doesn't happen when the keyboard just restarted after a crash without editing the emoji in between. But after inserting a fresh emoji it crashes 100% of the time for me.

Successfully reproduced in 3 apps so far: Discord, Signal, GitHub.
Happens regardless of whether the emoji was inserted from Unexpected Keyboard's emoji picker or the app's built-in one.

Crash log:
```
type: crash
osVersion: google/lynx/lynx:16/BP4A.260205.001/2026042101:user/release-keys
flags: dev options enabled
package: juloo.keyboard2:51, targetSdk 36
process: juloo.keyboard2
processUptime: 1116080 + 445 ms
installer: dev.imranr.obtainium

java.lang.IndexOutOfBoundsException: start 21, end 20, length 21
	at java.lang.AbstractStringBuilder.checkRange(AbstractStringBuilder.java:1742)
	at java.lang.AbstractStringBuilder.insert(AbstractStringBuilder.java:1267)
	at java.lang.StringBuilder.insert(StringBuilder.java:384)
	at juloo.keyboard2.j.o(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:48)
	at juloo.keyboard2.j.k(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:20)
	at juloo.keyboard2.j.i(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:35)
	at juloo.keyboard2.j.j(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:43)
	at juloo.keyboard2.w.v(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:8)
	at juloo.keyboard2.Keyboard2.onUpdateSelection(r8-map-id-1bc1df4b31f7768a9e71517982efe853977df0ad65555ba2251eae08e3cf7523:11)
[...]

PR #1281, opened 2026-05-06
fix: Suggestions do not appear after a space

### LINKED ISSUE #1269: Autocorrect/suggestion bugs out with spaces
When typing in a text box/field, the first letters/word triggers the autocorrect to suggest some words, however after the first space, it stops working entirely until backspace is pressed. Please refer to the below video for a demonstration. Version 2.0.1.

https://github.com/user-attachments/assets/4fd9f5e6-c302-481a-8f02-5bcad232a94d

AndreiD049:
Fixes #1269 : because we increment the loop counter before the boundary check, it needs to be inclusive.
Example: entering a single space will not start a new word because `i < end` will not hold true.
Works fine with emojis as well (#1259)
Verified on Android 16


Julow (approved):
Thanks a lot for this! I'll release this along with other fix during the week.

PR #1315, opened 2026-05-29
Add support for language switching for autocorrect

firefox97:
Added keyboard symbol for switching language, switch_dictionary
Added ability to download multiple languages at once
Added personal dictionary to check first
Added capture of backspace undo of autocorrect, to pre-populate add to personal dictionary

This does have my Dvorak layouts, which are also on another pull request with comments, so those can be left out if merged...


res/layout/dictionary_picker_pane.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Julow, line 1: The dictionary picker is way too complex. What about a key (maybe on the right side of the suggestion bar) that cycles through dictionaries ?

res/layout/personal_dict_activity.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Julow, line 1: My idea about the personal dictionary (that is maybe bad) is to present it as a text file with one word per line. Like the popup for entering a custom dictionary.
That way, people can copy and paste entire lists of words or they can add many words quickly without clicking on buttons.
What do you think ?

srcs/juloo.keyboard2/dict/DictionaryListView.java
```diff
@@ -39,24 +37,16 @@ public DictionaryListView(Context ctx, AttributeSet attrs)
 
   void inflate_views(Context ctx)
   {
-    DeviceLocales locales = DeviceLocales.load(ctx);
     SupportedDictionaries ds = new SupportedDictionaries(ctx.getResources());
     DownloadBtnListener listener = this.new DownloadBtnListener();
     _dict_views = new ArrayList<DictView>();
-    for (DeviceLocales.Loc loc : locales.installed)
+    for (int i = 0; i < ds.length(); i++)
```
Julow, line 43: The list is way too long. At least, the dictionaries for installed languages should be at the top. Maybe even in a different list view.

Julow (commented):
This is way too big. Here are some comments but to make progress, you really need to split this into separate PRs that implement a single feature at a time.

Julow:
I made a lighter version for switching dictionary in https://github.com/Julow/Unexpected-Keyboard/pull/1365
I'll close this PR as it's too big to review and contain too many features at once. Please open small PRs with individual features instead.

Issue #1309, opened 2026-05-26
Reverse clipboard history order


BaccanoMob:
Current when you copy something it puts the content to the last of the list. It would be helpful if its put it in the begining instead.

Thia is helpful when the clipboard history is set to longer duration which causes to have lengthy list.

Julow:
Fixed by @lokeshlkr in https://github.com/Julow/Unexpected-Keyboard/pull/1328

BaccanoMob:
@Julow @lokeshlkr Great work but It partially fixes. I gave the debug build I try. I wanted the clipboard history itself in reverse order as well.

lokeshlkr:
@BaccanoMob order reversed in history panel as well.
@Julow raised a new PR for same.

Julow:
Thanks :) The new debug build is here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/27876132414

BaccanoMob:
Yea, gave it a try. Works perfect!
Waiting for the next release.

PR #1328, opened 2026-06-13
when pinning history item, keep it on top

lokeshlkr:
when pinning history item, keep it on top

solves #1309


Julow (approved):
Awesome, thanks!

Issue #1325, opened 2026-06-13
[FEATURE] Key value to hide keyboard


Sp3EdeR:
Hey @Julow and other maintainers!

I'm freshly in love with this keyboard. Just discovered it. I would like to request a key value option that can be used in a custom layout that can hide the keyboard. The action and esc keys kind of do this in some apps already, but sometimes esc exits the entire program, and sometimes action does something else too. It would be pretty handy if you could hide the keyboard without moving your fingers out from the centre area, such as by swiping down on space.

Julow:
This was implemented by @lokeshlkr last month in https://github.com/Julow/Unexpected-Keyboard/pull/1329

PR #1329, opened 2026-06-13
add hide_self key to hide keyboard

lokeshlkr:
adds a `Event` key `hide_self` that will hide the keyboard when possible.

addresses issue #1325


Julow (approved):
Awesome! I pushed to your branch to update the doc.
Let's merge!

PR #1330, opened 2026-06-14
add new item on top for history view as well

lokeshlkr:
add new item on top for history view as well

Address #1309


Julow (approved):
Many thanks :)

PR #1334, opened 2026-06-17
Split layout in landscape mode

Julow:
This inserts empty space in the center of layouts when in landscape orientation.
The keys are smaller and pushed to the edges of the screen. The keys that are on the middle of the screen are duplicated on both sides, as they might be reach by both thumbs, and to avoid unbalanced layouts.
This way, the fingers don't have to reach the center of the screen, which is inconvenient in landscape orientation.

Here's how it looks:
shot

### DIFF
--- a/srcs/juloo.keyboard2/LayoutLandscapeModifier.java
+++ b/srcs/juloo.keyboard2/LayoutLandscapeModifier.java
@@ -1,0 +1,49 @@
+package juloo.keyboard2;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class LayoutLandscapeModifier
+{
+  /** Width added to the layout, in the key width unit of the layout. */
+  public static final int ADDED_WIDTH = 5;
+
+  public static KeyboardData transform_to_landscape(KeyboardData kw)
+  {
+    ArrayList<KeyboardData.Row> new_rows = new ArrayList<KeyboardData.Row>();
+    for (KeyboardData.Row r : kw.rows)
+      new_rows.add(split_row(r));
+    return kw.with_rows(new_rows);
+  }
+
+  public static KeyboardData.Row transform_number_row(KeyboardData.Row r)
+  {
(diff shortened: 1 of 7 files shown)


Julow:
I also plan to implement this for the number layout, after I done more testing. Feedback welcome.

ribhar:
LGTM

matthiakl:
Thank you for implementing this.
With my keyboard layout (neo 2) I noticed that the space bar width is inconsistent. I think it should align either with the row above (as on the left) or with the furthest character row.
Maybe in this case the "p" is placed invisibly on the right? It is not duplicated at all, probably because in portrait mode it is exactly in the middle.
Screenshot_20260618-181811

Julow:
Thanks for your comment :) I fixed the bug. The intention is that they align with the s and k keys.
The p key is not duplicated on purpose. It's unambiguously closer to the left thumb so it's on the left.
I plan to allow more control for custom layouts in the future, I'll ping you.
The new build is here: https://github.com/Julow/Unexpected-Keyboard/actions/runs/27875028618?pr=1334

srcs/juloo.keyboard2/LayoutLandscapeModifier.java
```diff
@@ -0,0 +1,73 @@
+package juloo.keyboard2;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class LayoutLandscapeModifier
+{
+  /** Width added to the layout, in the key width unit of the layout. */
+  public static final int ADDED_WIDTH = 5;
```
matthiakl, line 9: Could you make this configurable then, too?
To allow adjustment for different screen and hand sizes.

Julow: Not yet! I'm making some experimentation to use the space in the middle and I don't want to make it configurable until I have a clearer idea.

Issue #636, opened 2024-05-19
How to export my current setting/layout?


AnabasisXu:
I would like to start from my current layout instead of starting from the default. 

Is there any way I can export my current layout?

Spike-from-NH:
I keep my custom layout in a text file, which can be copied to other phones in several ways. On a new Unexpected installation, use Settings→Layout to select Custom. Then Select All and paste in your personal text file.

If you don't have your preferences in a text file, create a file and copypaste the text from Unexpected into it.

The other issue you raised suggests you already have a text file. Please clarify if I misunderstand your question.

Julow:
It's not possible to find the definition of a layout within the app but you can find them all here: https://github.com/Julow/Unexpected-Keyboard/tree/master/srcs/layouts

For changing fn and alt, you will need to redefine the bottom row. This requires setting the `bottom_row="false"` attribute and then defining the bottom row, see this layout that does that: https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/layouts/latn_neo2.xml

Spike-from-NH:
That is a good answer, but again, @AnabasisXu seems already to have a copy.

@JapanYoshi, the `layouts` folder should be added to your documentation; users who want to write XML will look there for a file as their starting point.

AnabasisXu:
Thank you both! @Spike-from-NH As you said, I wish I could point to a file in the app as a config file and update my layout/setting from there, which would save a lot of copying/pasting effort.

JapanYoshi:
> That is a good answer, but again, @AnabasisXu seems already to have a copy.
> 
> @JapanYoshi, the `layouts` folder should be added to your documentation; users who want to write XML will look there for a file as their starting point.

Added a mention.

Spike-from-NH:
Thanks, @JapanYoshi!

@AnabasisXu, copypasting layouts into Unexpected→Settings (I have two) for a new phone is a minor annoyance but I know no better way.  One can still update a new installation in less than one minute.

58673:
Can you add a singular option to export everything to a file? I change several of the options to suit personal preferences and haven't been able to export between devices (and work profiles) by "pasting" the layout.

PR #1337, opened 2026-06-21
Export/Import settings

### LINKED ISSUE #856: Backup/Restore settings
An import and restore app settings function is extremely needed. The app has been getting more and more configurations, which takes more and more time to get it ready for use. Please consider this in the future. Thank you.

matthiakl:
Provide menu entries to export and import settings to a text file, including custom layout definitions.
Resolves #856 and #636
Code inspired by [grocy-android](https://github.com/patzly/grocy-android/blob/d0b881637ac71463354c106843b0164ae4b9e4ed/app/src/main/java/xyz/zedler/patrick/grocy/util/PrefsUtil.java)


res/values/strings.xml
```diff
@@ -165,4 +165,11 @@
     <string name="pref_physical_keyboard_behavior">When a physical keyboard is connected</string>
     <string name="pref_physical_keyboard_behavior_hide">Hide everything</string>
     <string name="pref_physical_keyboard_behavior_show">Show everything</string>
+    <string name="pref_category_backup">Backup</string>
+    <string name="pref_settings_import">Import settings</string>
+    <string name="pref_settings_export">Export settings</string>
+    <string name="export_success">Settings exported</string>
+    <string name="export_fail">Settings export failed</string>
+    <string name="import_success">Settings imported, app will restart</string>
+    <string name="import_fail">Settings import failed</string>
```
Julow: I think there's too many strings. Do we need all these toast message ? Importing and exporting should be instantaneous.
I think only the error case need a toast message.

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -1,29 +1,75 @@
...
+
+      return true;
+    });
+
+    final Preference exportDataPreference = findPreference("settings_export");
+    exportDataPreference.setOnPreferenceClickListener((final Preference p) -> {
+      Intent intent = new Intent(Intent.ACTION_CREATE_DOCUMENT);
+      intent.addCategory(Intent.CATEGORY_OPENABLE);
+      intent.setType("text/plain");
+      intent.putExtra(Intent.EXTRA_TITLE, "prefs.txt");
```
Julow: Let's name it `"unexpected-keyboard-prefs.txt"` ?

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -43,7 +108,83 @@ protected void onStop()
...
+
+      post_toast(R.string.import_success);
+      // Restart app
+      new Handler().postDelayed(
+              () -> {
+                Intent intent = new Intent(this, SettingsActivity.class);
+                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+                this.startActivity(intent);
+                this.finish();
+                Runtime.getRuntime().exit(0);
```
Julow: Isn't the settings updated automatically after `editor.apply()` ? I'd prefer if the activity was not restarted.

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -43,7 +108,83 @@ protected void onStop()
...
+    }
+  }
+  private void importFromFile(Uri uri) {
+    try (InputStream inputStream = this.getContentResolver().openInputStream(uri);
+         BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+
+      // Clear all existing preferences
+      SharedPreferences.Editor editor = sharedPreferences.edit();
+      editor.clear();
+      editor.apply();
```
Julow: I think it's not right to clear the prefs. For example, an old backup or a file modified by the user may not contain all the keys.

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -43,7 +108,83 @@ protected void onStop()
...
+
+      editor = sharedPreferences.edit();
+      String line;
+      while ((line = reader.readLine()) != null) {
+        String[] keyValue = line.split("=", 2);
```
Julow: You can change the format to avoid calling `split` twice, which is less efficient:
```
key;String;value
```
and split it like this:
```
String[] kvt = line.split(";", 3);
```

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -43,7 +108,83 @@ protected void onStop()
...
+      new Handler().postDelayed(
+              () -> {
+                Intent intent = new Intent(this, SettingsActivity.class);
+                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+                this.startActivity(intent);
+                this.finish();
+                Runtime.getRuntime().exit(0);
+              }, 2000
+      );
+    } catch (IOException e) {
```
Julow: `IOException` is not the only exception. There might be one from `Integer.parseInt`, etc..
When you wrap unsafe code (in the sense that it can throw due to user input) into code that is not affected by failures (such as the settings activity), you can just catch everything with `catch (Exception e)`.

srcs/juloo.keyboard2/SettingsActivity.java
```diff
@@ -43,7 +108,83 @@ protected void onStop()
...
+
+  private void exportToFile(Uri uri) {
+    try (OutputStream stream = this.getContentResolver().openOutputStream(uri);
+         OutputStreamWriter writer = new OutputStreamWriter(stream, StandardCharsets.UTF_8)) {
+      Map<String, ?> allPrefs = sharedPreferences.getAll();
+      for (String key : allPrefs.keySet()) {
+        Object value = allPrefs.get(key);
+        if (value == null) continue;
+        String valueType = value.getClass().getSimpleName();
+        writer.write(key + "=" + valueType + ";" + value + "\n");
```
Julow: The format breaks if `value` ever contains a newline character. For some reason this doesn't happen but nothing in this code prevents it.
Custom layouts can contain newlines but by chance, when converted to JSON (in [ListGroupPreference](https://github.com/Julow/Unexpected-Keyboard/blob/master/srcs/juloo.keyboard2/prefs/ListGroupPreference.java)), we use the [compact](https://developer.android.com/reference/kotlin/org/json/JSONArray#tostring) encoding. This was not made on purpose and there's chance that we break this in the future.

To avoid any format problems, I suggest switching to JSON. It might be worth looking at this constructor, which takes a Map as input: https://developer.android.com/reference/kotlin/org/json/JSONObject#JSONObject(kotlin.collections.MutableMap)

Julow (commented):
Sorry for the slow reply. Thank you very much for working on this very nice feature :)

res/values/strings.xml
matthiakl: Done. I also merged the failure into a single message.

srcs/juloo.keyboard2/SettingsActivity.java
matthiakl: Done.

srcs/juloo.keyboard2/SettingsActivity.java
matthiakl: No, it somehow is updated in memory, but the UI does not receive an update. I will try to find another way.

srcs/juloo.keyboard2/SettingsActivity.java
matthiakl: I replaced the format with a JSONArray.

srcs/juloo.keyboard2/SettingsActivity.java
matthiakl: After some searching I found a simple solution. Instead of restarting, only the activity is refreshed by recalling onCreate().

PR #1353, opened 2026-07-03
More informative "download failed" message

### LINKED ISSUE #1347: unable to download dictionary
1 out of 3 devices was able to download the dictionary after multiple attempts. Both devices unable to download is on GrapheneOS.

Julow:
Fix: https://github.com/Julow/Unexpected-Keyboard/issues/1347

Ask users to allow internet access, which will make sense to users of grapheneOS and other systems that restrict internet access to apps.


BritishTeapot:
This does not necessarily have to happen because the access is not allowed. IMO it would be far better to print the HTTP error that happened during download and a tip at the bottom like "check if internet access is allowed".

Julow:
@BritishTeapot There were issues from users of systems that disallow internet access to apps, so this solves that. There's no HTTP error in this case and I'm not sure it's useful.

PR #1368, opened 2026-07-24
Prototype: embed direct Doubao voice input

Molaison:
## What changed

- Embed the Doubao IME registration, settings-token, WebSocket, protobuf, and
  streaming ASR flow directly in Unexpected Keyboard.
- Capture 16 kHz mono audio and encode exact 20 ms Opus frames with Concentus,
  retaining the existing API 21 minimum.
- Apply interim results through `setComposingText` and commit final text through
  `commitText`.
- Keep normal voice activation as a start/stop toggle.
- Treat long press as push-to-talk: start at the hold threshold and stop on
  release or touch cancellation.
- Add explicit microphone permission UI, lifecycle cancellation, visible
  failures, and protocol regression tests.
- Use the separate application ID `com.molaison.unexpectedkeyboard.doubao` so
  this prototype can coexist with the upstream installation.

## Why

Unexpected Keyboard's existing voice action switches to an enabled IME subtype
whose mode is `voice`. Doubao and several Chinese IMEs expose only a normal
`keyboard` subtype, with speech implemented inside their own keyboard UI.
Therefore they cannot be discovered by the existing switcher and the chooser
reports "No voice typing app installed".

This prototype calls the speech service directly and writes recognition results
through the current `InputConnection`, without switching IMEs.

## Validation

- `./gradlew testDebugUnitTest assembleDebug`
- 29 unit tests passed, 0 failed.
- `git diff --check`
- Physical vivo V2426A / Android 16 validation:
  - real credential, WebSocket, microphone, Opus, and final text path completed;
  - committed Mandarin text: `端到端语音输入测试。`;
  - normal activation toggled start/stop;
  - long press stopped on release and `ACTION_CANCEL`;
  - the selected IME remained unchanged;
  - the system voice-IME chooser did not appear.

## Draft notes

This is intentionally a draft because it currently uses Doubao-specific
endpoints/client identity and changes the application ID for side-by-side
testing. Upstream may prefer a configurable provider boundary and retention of
the original application identity before considering a merge.

### DIFF
--- a/.github/workflows/make-apk.yml
+++ b/.github/workflows/make-apk.yml
@@ -1,10 +1,8 @@
-name: Make Apk CI
+name: Build Doubao APK
 
 on:
   workflow_dispatch:
   push:
-    branches:
-      - master
   pull_request:
 
 jobs:
@@ -41,10 +39,11 @@ jobs:
 
     - name: Run tests
       run: |
-        gradle test
+        gradle checkKeyboardLayouts testDebugUnitTest
         # Warn about outdated generated files.
-        if ! git diff --name-only; then
+        if ! git diff --quiet; then
           echo "Warning: Generated files are not uptodate. Run 'gradle test'."
+          git diff --name-only
         fi
 
     - name: Build debug APK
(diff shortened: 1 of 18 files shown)


Julow (changes requested):
Sorry but I cannot accept this patch. At least two reasons are:

- The application must remain light and this adds a lot of code and dependencies. The debug APK grows by 1MB.
- I don't want Internet access. The app currently connects to the Internet in one place but I plan to remove it and I don't want more.

I suggest that you maintain of a fork of the app with this feature, possibly named "Unexpected Keyboard with Doubao".

Molaison:
> Sorry but I cannot accept this patch. At least two reasons are:
> 
> * The application must remain light and this adds a lot of code and dependencies. The debug APK grows by 1MB.
> * I don't want Internet access. The app currently connects to the Internet in one place but I plan to remove it and I don't want more.
> 
> I suggest that you maintain of a fork of the app with this feature, possibly named "Unexpected Keyboard with Doubao".

Copy that! Thanks for your reply!

Julow:
Thanks for your understanding :)

PR #1376, opened 2026-08-04
Add Dracula theme

tgll:
Adds a Dracula-inspired dark theme based on the official Dracula palette.

- Adds the theme resources and preference entry
- Does not modify keyboard behavior or rendering
(Tested on a Pixel 8 Pro API 37 emulator)

image


Julow (approved):
Many thanks !

PR #1377, opened 2026-08-05
Add floating keyboard toggle button on the candidates strip

zFitness:
## What
Add a toggle_float key action to switch between floating and docked keyboard modes.
Add a toggle button next to the change-language button on the candidates strip.
Persist the floating keyboard state via the floating_keyboard preference.
## Changes
KeyValue.java: new toggle_float action
Keyboard2.java: wire the toggle action to the floating-mode switch
CandidatesView.java: add the toggle button and update its selected state
New drawables: btn_floating_toggle, ic_floating_toggle(_selected)
keyboard.xml / strings.xml: button placement and label

## Screenshot
screen4
* setting
screen5


scripts/fix-ndk-arm64-host.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
Julow: Can you explain a bit more ? I'm not sure this should live in the repo. Instead, you should distribute it to other aarch64 users separately.

srcs/juloo.keyboard2/KeyValue.java
```diff
@@ -663,6 +664,7 @@ public static KeyValue getSpecialKeyByName(String name)
       case "complete_emoji": return statefulKey(Stateful.Complete_emoji);
       case "hide_self": return eventKey("⊻", Event.HIDE_SELF, FLAG_SMALLER_FONT);
       case "change_dictionary": return eventKey(0xE01D, Event.CHANGE_DICTIONARY, 0);
+      case "toggle_float": return eventKey("Float", Event.TOGGLE_FLOATING, FLAG_SMALLER_FONT);
```
Julow: You added an icon for the button in the status bar, so the key could use it too.
Add the (original) svg to `srcs/special_font` and run `gradle buildKeyboardFont`. The next available name is `01D.svg`. I can run the build if the dependencies are a problem.

Julow: The key should perhaps be called `toggle_floating` to be consistent with the event name. (or the opposite)

.gitignore
```diff
@@ -11,3 +11,5 @@
 # Directory _build is not used anymore
 /_build
 /release
+/.opencode
```
Julow: Please add these to your personal gitignore. I don't want to support dev environments in this file because it creates constant maintenance.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -313,6 +364,107 @@ private void updateSoftInputWindowLayoutParams() {
...
+    lp.x = pos.x;
+    lp.y = pos.y;
+    window.setAttributes(lp);
+  }
+
+  /** Width of the floating keyboard window, in pixels. */
+  private int floatingWindowWidth()
+  {
+    android.util.DisplayMetrics dm = getResources().getDisplayMetrics();
+    return (int)Math.min(dm.widthPixels * Config.FLOATING_KEYBOARD_WIDTH_RATIO,
```
Julow, line 387: The same computation is done in Keyboard2View too. It would be nice to share this. Or perhaps compute it in one place and store the result in Config.

srcs/juloo.keyboard2/Keyboard2.java
```diff
@@ -56,6 +63,12 @@ public class Keyboard2 extends InputMethodService
 
   private FoldStateTracker _foldStateTracker;
 
+  /** Drag state of the floating keyboard handle. */
+  private float _dragStartX;
```
Julow, line 67: It would be great if all the code that works with the floating handle be moved in an other class. Maybe it could be moved into `FloatingHandleView` with a callback to call back into `Keyboard2`.
Perhaps the code that goes in and out of floating mode could also go in a new class ?

srcs/juloo.keyboard2/suggestions/CandidatesView.java
```diff
@@ -101,6 +104,8 @@ public void refresh_config(Config config)
     else if (_status_no_dict != null)
       _status_no_dict.setVisibility(View.GONE);
     should_show_dictionary_switch = config.should_show_dictionary_switch;
+    _floating_toggle_button.setSelected(config.floating_keyboard);
+    _floating_toggle_button.setVisibility(View.VISIBLE);
```
Julow, line 108: It's always visible. Did you plan to remove the button in some cases ?
Does floating mode makes sense on small screens ? Perhaps the button should be removed in this case ?

Julow (commented):
This is awesome!

PR #1381, opened 2026-08-07
Add additional virama sign to Sinhala layout

bergentroll:
XKB Sinhala (phonetic) layout has multiple virama (al-lakuna) signs, but Unexpected Keyboard layout based on XKB one currently has only one virama on layer 2 (Shift + අ). Because it is a frequent sign the PR adds another one on NE of rayanna (ර) key.

Also numeric codes in a script replaced with human-readable Unicode char names.


Julow (approved):
Thanks!

PR #1384, opened 2026-08-09
Add a personal dictionary with shortcuts

Gareth-Rouse:
Adds a user-maintained personal dictionary: words and abbreviations entered in settings appear in the suggestion bar, independent of the compiled dictionaries.

## Features

- New **Personal dictionary** settings screen (under *Suggestions*) using `ListGroupPreference`; entries are stored as JSON (plain strings, or `{word, shortcut}` objects when a shortcut is attached — old plain-string lists keep loading).
- Words are suggested on a case-insensitive prefix match; stored casing is preserved and follows the capitalization of the typed word, matching the behavior of compiled-dictionary suggestions.
- **Shortcuts**: typing a shortcut exactly expands to the full word/text (e.g. `aa` → an email address). Expansions rank first and are returned verbatim.
- Matching uses the same substitution table as the compiled dictionaries (`ComposeKey.transform_char(ComposeKeyData.substitutions, …)`), normalized once at load time, so typing `cafe` matches a stored `café` and vice-versa.
- Suggestion slots: shortcut expansions take the first slots, then the compiled dictionary, then personal word matches fill whatever slots remain — personal words can never hide dictionary results entirely, and an empty dictionary result never wastes a slot.
- The entry dialog rejects shortcuts that can never fire (shorter than the 2-character suggestion gate, or containing non-word characters per `CurrentlyTypedWord.is_word_char`).
- Personal suggestions work with no dictionary installed.

## Also in this PR

- `ListGroupPreference.StringSerializer.load_item` now throws `JSONException` instead of `ClassCastException` when an item isn't a string, so a future format change degrades to an empty list instead of crashing `Config.refresh` on downgrade/upgrade.
- Unit tests for `PersonalDictionary` (matching, ranking, capitalization, substitutions, dedup, truncation): 7 tests.

Tested on-device (debug build, installed as a separate app id).

### DIFF
--- a/res/layout/dialog_personal_dict_entry.xml
+++ b/res/layout/dialog_personal_dict_entry.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android" android:orientation="vertical" android:layout_width="fill_parent" android:layout_height="wrap_content">
+  <EditText android:id="@+id/personal_dict_word" android:hint="@string/pref_personal_dict_word_hint" android:inputType="text" android:importantForAutofill="no" android:layout_width="fill_parent" android:layout_height="wrap_content" android:layout_margin="10dp"/>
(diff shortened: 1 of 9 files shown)


Julow (commented):
Sorry, I have to postpone reviewing your PR a bit more as I'm preparing for a new release and don't have enough time.

Issue #1394, opened 2026-08-17
Crash


SainoNamkho:
Process to reproduce:
- type `a`
- move cursor left by one step
- type space
- backspace

Crash report
```
java.lang.StringIndexOutOfBoundsException: start 0, end -1, length 0
	at java.lang.AbstractStringBuilder.checkRangeSIOOBE(AbstractStringBuilder.java:1729)
	at java.lang.AbstractStringBuilder.delete(AbstractStringBuilder.java:868)
	at java.lang.StringBuilder.delete(StringBuilder.java:304)
	at juloo.keyboard2.m.c(r8-map-id-a9863cef79b6b3346410c1e7cbc9e52527bfb8a33b88656bf8833236dbe7be4f:27)
	at juloo.keyboard2.f0.l(r8-map-id-a9863cef79b6b3346410c1e7cbc9e52527bfb8a33b88656bf8833236dbe7be4f:100)
	at juloo.keyboard2.f0.e(r8-map-id-a9863cef79b6b3346410c1e7cbc9e52527bfb8a33b88656bf8833236dbe7be4f:31)
	at juloo.keyboard2.Keyboard2View.onTouch(r8-map-id-a9863cef79b6b3346410c1e7cbc9e52527bfb8a33b88656bf8833236dbe7be4f:141)
	at android.view.View.performOnTouchCallback(View.java:16676)
	at android.view.View.dispatchTouchEvent(View.java:16640)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3158)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3158)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3158)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3158)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3158)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
[...]

PR #1395, opened 2026-08-17
Minimal fix for #1394

SainoNamkho:
The logic is still incorrect, but remove the crash for the moment.

### DIFF
--- a/srcs/juloo.keyboard2/CurrentlyTypedWord.java
+++ b/srcs/juloo.keyboard2/CurrentlyTypedWord.java
@@ -128,9 +128,12 @@ public final class CurrentlyTypedWord
       return;
     int len = _w.length();
     int c = len + _w_cursor;
-    _w.delete(Math.max(c - remove_before, 0), Math.min(c + remove_after, len));
+    assert remove_before >= 0 && remove_after >= 0;
+    _w.delete(Math.max(c - remove_before, 0), Math.clamp(c + remove_after, 0, len));
     _cursor -= remove_before;
-    _w_cursor -= Math.min(remove_after, 0);
+    _w_cursor += remove_after;
+    if (_w.length() == 0)
+        _w_cursor = 0;
     callback();
   }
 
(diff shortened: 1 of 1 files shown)


SainoNamkho:
@Julow I finally untangled the logic and tried to fix it. After testing, I'm optimistically confident that the current version is correct.
However I left a TODO 😂, there's still work to be done for deleting selections and emojis,
like
https://github.com/Julow/Unexpected-Keyboard/blob/396541e832cf3013a898321078d1269486d2ee5f/srcs/juloo.keyboard2/KeyEventHandler.java#L178-L186

I feel it a little difficult to figure out the supposed behavior when the cursor is inside a grapheme cluster, so I can only leave the most difficult task back to you.

Wishes.

SainoNamkho:
Fix #1394

SainoNamkho:
Maybe this? I'm not sure
```java
public void remove_surrounding_text(int remove_before, int remove_after)
{
  if (!_enabled)
    return;
  if (_has_selection)
  {
    delayed_refresh();
    return;
  }
  ...
  if (remove_before == 1)
  {
    int codePointBefore = _w.codePointBefore(c);
    int typeBefore = Character.getType(codePointBefore);
    if (Character.charCount(codePointBefore) > 1  // cursor inside surrogate pair, possible?
        || typeBefore == Character.NON_SPACING_MARK 
        || typeBefore == Character.COMBINING_SPACING_MARK 
        || typeBefore == Character.ENCLOSING_MARK)
    {
      delayed_refresh();
      return;
    }

  // look forward for composition letters and VSs
  if (c < len)
  {
    int codePointAfter = _w.codePointAt(c);
    int type = Character.getType(codePointAfter);
    if (type == Character.NON_SPACING_MARK 
        || type == Character.COMBINING_SPACING_MARK 
        || type == Character.ENCLOSING_MARK
        || Character.isLowSurrogate(_w.charAt(c)))
    {
      delayed_refresh();
      return;
    }
  }
  ...
}
```

Julow:
Thanks but I think the logic could be made simpler. The code that makes `_w_cursor` go below 0 still exist too.
I merged https://github.com/Julow/Unexpected-Keyboard/pull/1396 to fix the immediate bug and I will look at it again later.

SainoNamkho:
> Thanks but I think the logic could be made simpler. The code that makes `_w_cursor` go below 0 still exist too. I merged #1396 to fix the immediate bug and I will look at it again later.

Glad to hear that. I personally don't use suggestions, so keeping it simple and fast is good  enough to me.

Not important, but https://github.com/Julow/Unexpected-Keyboard/blob/9ce5c41814164a881dccc6842ba4d920748ab78d/srcs/juloo.keyboard2/CurrentlyTypedWord.java#L133
is this supposed be `+=`?

PR #1398, opened 2026-08-19
Allow `symbol:slider`, make single sliders not macro in this syntax

SainoNamkho:
Currently keys like `cursor_up` does not work if configured with the long syntax.

### DIFF
--- a/srcs/juloo.keyboard2/KeyModifier.java
+++ b/srcs/juloo.keyboard2/KeyModifier.java
@@ -455,7 +455,7 @@ public final class KeyModifier
           case Cursor_left: name = "selection_cursor_left"; break;
           case Cursor_right: name = "selection_cursor_right"; break;
         }
-        break;
+        return (name == null) ? k : KeyValue.getSpecialKeyByName(name).withSymbol(k.getString());
       case Keyevent:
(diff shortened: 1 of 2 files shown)


Julow (commented):
I feel like this is a lot of code for an edge case. What do you think of leaving the bug in place with a mention in the doc ?

SainoNamkho:
> I feel like this is a lot of code for an edge case. What do you think of leaving the bug in place with a mention in the doc ?

Maybe I mixed up too much problems in a single patch because one cannot be fixed without the others.😂

I have a specific use case, what do you think of it?
I make every swipe in cardinal directions, which have no function yet, a cursor move. So I can swipe to move cursor from any position. And I hide the icons, because otherwise there will be arrows everywhere on the keyboard. Sounds like a well-founded need?

In favor of that, I need:
- cursor moves with customized symbols, which are macros now, work as sliders.
- the symbols persist while modified.
- in order to implement the previous item, calls to `withSymbol` might pass key font symbols. I choose to check if the first character is in PUA. This is inconsistent with other typed based overload solutions, but there isn't a good way to do this.

I'm pleased if there would be other ways for this need.😊

PR #1408, opened 2026-09-02
Fix to light theme issue

### LINKED ISSUE #942: Use system styling in settings

Insistar:
Reverted a file change in #1146 
Fix #942 
Should now work like before (use system theme instead of lighr theme only)


Julow (approved):
Thanks a lot !

PR #1414, opened 2026-09-05
Fix crash when the number row is added to a split layout

SickleSellion:
Enabling the number row (with or without symbols) while the layout is split crashes the keyboard, and then keeps crashing on every `onStartInputView` until the number row is disabled again.

```
java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 4
    at juloo.keyboard2.LayoutLandscapeModifier.add_middle_key(LayoutLandscapeModifier.java:86)
    at juloo.keyboard2.LayoutLandscapeModifier.split_at_index(LayoutLandscapeModifier.java:57)
    at juloo.keyboard2.LayoutLandscapeModifier.split_row(LayoutLandscapeModifier.java:45)
    at juloo.keyboard2.LayoutLandscapeModifier.transform_number_row(LayoutLandscapeModifier.java:23)
    at juloo.keyboard2.LayoutModifier.modify_layout(LayoutModifier.java:52)
    at juloo.keyboard2.Keyboard2.current_layout(Keyboard2.java:81)
    at juloo.keyboard2.Keyboard2.onSharedPreferenceChanged(Keyboard2.java:381)
```

**Cause.** `transform_number_row()` splits the number row with `row_index = -1` to mean "no middle-column key", but `add_middle_key()` only guards the upper bound and indexes `split_middle_column.keys` at -1. Introduced with the suggestion column in #1350, so it affects every layout that splits: wide screens and tablets with the default `split_layout=wide`, and phones in landscape.

**Reproduce.** Any device in a state where the layout is split (a phone in landscape is enough), Settings → Layout → Number row → with or without symbols.

**Fix.** Guard negative indexes in `add_middle_key()`.

**Test.** Adds `LayoutLandscapeModifierTest`, a plain JVM test of the split geometry with rows built programmatically (no Android resources needed): the number-row regression, middle-key insertion, the duplicate-key path, rows with fewer than two keys, and whole-layout splitting from the bottom. It fails before the fix with the exception above and passes after.

Seen on a Galaxy Z Fold (Android 17) where the layout is split on both screens, but the code path is device-independent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/srcs/juloo.keyboard2/LayoutLandscapeModifier.java
+++ b/srcs/juloo.keyboard2/LayoutLandscapeModifier.java
@@ -81,7 +81,9 @@ public final class LayoutLandscapeModifier
       int row_index)
   {
     List<KeyboardData.Key> mid_keys = LayoutModifier.split_middle_column.keys;
-    if (row_index >= mid_keys.size())
+    // [row_index] is negative for rows that must not receive a middle key,
+    // such as the number row. See [transform_number_row].
+    if (row_index < 0 || row_index >= mid_keys.size())
       return;
     KeyboardData.Key mid_key = mid_keys.get(row_index);
     KeyboardData.Key right_key = new_keys.get(i);
--- a/test/juloo.keyboard2/LayoutLandscapeModifierTest.java
+++ b/test/juloo.keyboard2/LayoutLandscapeModifierTest.java
@@ -1,0 +1,133 @@
+package juloo.keyboard2;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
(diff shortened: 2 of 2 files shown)


Julow (approved):
Awesome, thanks !

eandersons:
Because of the negative and significant impact of the current implementation of LLMs, there are people (including me) that prefer using and are trying to avoid as much as possible software that was not created with the help of slop-generators.
@julow, would you consider to apply no-“AI” policy to this project? Or, if you have different view and opinions, you could add AI policy (as a separate file or as a section in `README.md`) so it would be clear what is permitted and what is not.

Julow:
I personally don't use LLMs at all and I prefer not to read LLM generated code. But at least for now, I still want to review contributors code when it's not outright slop. It's a good idea to clarify things in the readme!

PR #1432, opened 2026-09-14
Override setExtractViewShown to always be false

granitrocky:
On newer android versions, simply overriding `onEvaluateFullscreenMode` isn't enough to disable the fullscreen editor. I had to also override `setExtractViewShown` to always be false. This corrects the behavior and now the view never goes full screen.


Julow (approved):
Thanks a lot!

